Issue #1, opened 2016-09-27
Internal transport discussion


jonparrott:
This library needs to satisfy two orthogonal goals related to HTTP:

1. It needs to be able to make HTTP requests internally (e.g., in [refresh()](https://github.com/jonparrott/goth/blob/master/google/auth/service_account.py#L260))
2. It needs to allow users to attach credentials to an HTTP object of their choosing.

If there were only one Python HTTP library this would all be trivial. If the library was called `unicorn`, we could just use `unicorn.request` directly everywhere internally for (1). For (2), we would just write a single function to attach the credentials to `unicorn`.

This discussion is exclusively about (1), but care should be made not to limit our options for (2).

# Use Case (1)

Use case (1) boils down to the need for a consistent way to make an http request and get response information. This use-case is only for library internals. 

## What's done today

`transport.request` takes any http object and then uses isinstance to figure out which transport should be used to call. Basically:

```python
# Calling code
credentials.refresh(http, ...)

# credentials.py
def refresh(http):
    ...
    r = transport.request(http, ...)
    ...

# transport.py
def request(http, method, uri, ...):
    transport_module = get_transport_for_http(http)
    return transport_module.request(http, method, uri, ...)

# urllib3_transport.py
def request(http, method, uri, ...):
    return http.request(method, uri)
```

The indirection has the benefit that we can accept any supported http object at the call sites of methods like `refresh()`. The drawback is the indirection and the cost of the lookup for every request.

### Use http.client

We could easily use the built-in `http.client` - but urlilb3 and requests have significant benefits, and properly doing ssl with http.client can be error prone. It also leads us to situation where the http client the user is using and the http client we're using can be different.

### Set credentials.http

Presently credentials are completely independent of http. Any methods that need to make requests must accept an ``http`` parameter. This proposes adding ``credentials.http`` that is set to a new **private** interface ``transport._HTTP``. None of the request-making methods need http as an argument anymore.

This has the distinct drawback that credentials can now only be associated with a single http object. It also introduces a new interface (and subclasses for every transport).

### Just pass in transport_impl.request

Instead of passing in an http object and letting the lookup happen, make the argument to `refresh()` et al be an interface that matches `transport.request` like this:

```python
def request(method, uri, ...)
```
Since the caller of refresh *knows* the transport implementation that's currently being used it can just pass in, for example, `functools.partial(urlilb3_transport.request, bare_http)`. The call stack would then look like this:

```python
# Calling code
[...]

PR #2, opened 2016-09-28
Add transport

jonparrott:
Proof-of-concept for #1.

(Not fully implemented, tests will not pass).

I validated this with:

```python
import logging

from google.auth import service_account
import google.auth.transport.urllib3
import urllib3

logging.basicConfig(level=logging.DEBUG)

http = urllib3.PoolManager(cert_reqs='CERT_REQUIRED')

request = google.auth.transport.urllib3.Request(http)

sa_credentials = service_account.Credentials.from_service_account_file(
    'service_account.json',
    scopes=['https://www.googleapis.com/auth/cloud-platform'])

sa_credentials.refresh(request)

print(sa_credentials.token)
```


jonparrott:
Also by happy accident I forgot to implement ``transport.urllib3.Response``, but the urllib3 Response class interface is a superset of ``transport.Response``, so everything still works.

jonparrott:
@dhermes I know I promised to unplug, but I had a hunch about the tests. It seems like a good indication that this allows me to reduce the usage of `mock.patch` in the tests. :)

google/auth/_default.py
```diff
@@ -112,7 +112,9 @@ def _get_gae_credentials():
 
 
 def _get_gce_credentials():
-    if _metadata.ping():
+    # TODO: Ping now requires a request argument. Figure out how to deal with
```
dhermes: I am big time :-1: on `TODO` in code that doesn't have a corresponding issue filed

google/auth/exceptions.py
```diff
@@ -20,6 +20,11 @@ class GoogleAuthError(Exception):
     pass
 
 
+class TransportError(Exception):
+    """Used to indicate an error occurred during an HTTP request."""
+    pass
```
dhermes: If you have a docstring you don't need `pass`

google/auth/transport/__init__.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2016 Google Inc. All rights reserved.
```
dhermes: (Sorry for just noticing, but the "All rights reserved." is no longer needed)

https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2390

google/auth/transport/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Returns:
+            Response: The HTTP response.
+
+        Raises:
+            google.auth.exceptions.TransportError: If any exception occurred.
+        """
+        # pylint: disable=redundant-returns-doc, missing-raises-doc
+        # (pylint doesn't play well with abstract docstrings.)
+        raise NotImplementedError('__call__ must be implemented.')
```
dhermes: 1. You should have a `pylint: enable` at the end of the block
1. Why is the `NotImplementedError` needed here, but not in the other abstract methods?

google/auth/transport/urllib3.py
```diff
@@ -0,0 +1,67 @@
...
+        http (urllib3.requests.RequestMethods): An instance of any urllib3
+            class that implements :cls:`~urllib3.requests.RequestMethods`,
+            usually :cls:`urllib3.PoolManager`.
+    """
+    def __init__(self, http):
+        self.http = http
+
+    def __call__(self, url, method='GET', body=None, headers=None,
+                 timeout=None, **kwargs):
+        """
```
dhermes: No one-line description?

google/auth/_default.py
jonparrott: I agree. I've temporarily suspended pylint's TODO checker, but will re-enable it as this transitions from proof-of-concept to actual library. TODOs are there because I'm not quite sure how it all fits together just yet.

google/auth/exceptions.py
jonparrott: that feels funny.

google/auth/transport/__init__.py
jonparrott: Thanks, I'll use your script to fix all of them.

google/auth/transport/__init__.py
jonparrott: Per https://docs.pylint.org/en/latest/user_guide/message-control.html, disable comments are scoped at the block level, so the comment at the end of the block shouldn't be needed.

For (2) - good point, will fix.

google/auth/transport/urllib3.py
jonparrott: Proof-of-concept, will fix.

google/auth/transport/__init__.py
dhermes: I should have mentioned that I realized the block scope terminates itself, but it's just good hygiene to make sure you close up and disables.

google/auth/transport/__init__.py
jonparrott: thanks.

PR #7, opened 2016-10-04
Add google.auth.jwt

jonparrott:


google/auth/jwt.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+    if not header:
+        header = {}
+
+    if key_id is None:
+        key_id = signer.key_id
+
+    header.update({'typ': 'JWT', 'alg': 'RS256'})
+
+    if key_id is not None and key_id is not False:
```
dhermes: I don't understand the `False` bit

google/auth/jwt.py
```diff
@@ -0,0 +1,233 @@
...
+        header['kid'] = key_id
+
+    segments = [
+        base64.urlsafe_b64encode(json.dumps(header).encode('utf-8')),
+        base64.urlsafe_b64encode(json.dumps(payload).encode('utf-8')),
```
dhermes: Don't you already have a helper for doing `.encode('utf-8')`?

google/auth/jwt.py
```diff
@@ -0,0 +1,233 @@
...
+
+    return b'.'.join(segments)
+
+
+def _decode_jwt_segment(encoded_section):
+    """Decodes a single JWT segment."""
+    section_bytes = base64.urlsafe_b64decode(encoded_section)
+    try:
+        return json.loads(section_bytes.decode('utf-8'))
+    except:
```
dhermes: Please use `ValueError`. `json.loads` doesn't need a broad except?

google/auth/jwt.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def _unverified_decode(token):
+    """Decodes a token and does no verification.
+
+    Args:
+        token (Union[str, bytes]): The encoded JWT.
+
+    Returns:
+        Tuple(str, str, str, str): header, payload, signed_setion, and
```
dhermes: s/`signed_setion`/`signed_section`

google/auth/jwt.py
```diff
@@ -0,0 +1,233 @@
...
+    if isinstance(certs, collections.Mapping):
+        key_id = header.get('kid')
+        if key_id:
+            if key_id not in certs:
+                raise ValueError(
+                    'Certificate for key id {} not found.'.format(key_id))
+            certs = [certs[key_id]]
+        # If there's no key id in the header, check against all of the certs.
+        else:
+            certs = certs.values()
```
dhermes: Both `certs = certs.values()` and `certs = [certs[key_id]]` are strange. Can you use a different name for the LHS variable? `cert_vals`?

tests/test_jwt.py
```diff
@@ -0,0 +1,175 @@
+# Copyright 2014 Google Inc.
```
dhermes: s/2014/2016

tests/test_jwt.py
```diff
@@ -0,0 +1,175 @@
...
+DATA_DIR = os.path.join(os.path.dirname(__file__), 'data')
+
+with open(os.path.join(DATA_DIR, 'privatekey.pem'), 'rb') as fh:
+    PRIVATE_KEY_BYTES = fh.read()
+
+with open(os.path.join(DATA_DIR, 'public_cert.pem'), 'rb') as fh:
+    PUBLIC_CERT_BYTES = fh.read()
+
+with open(os.path.join(DATA_DIR, 'other_cert.pem'), 'rb') as fh:
+    OTHER_CERT_BYTES = fh.read()
```
dhermes: I feel like I've seen these values before

tests/test_jwt.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@pytest.fixture
```
dhermes: Fancy. What does this do?

google/auth/_helpers.py
```diff
@@ -14,9 +14,24 @@
...
 """Helper functions for commonly used utilities."""
 
+
+import calendar
+import datetime
+
 import six
 
 
+def now():
```
jonparrott: Suggest a better name?

google/auth/_helpers.py
```diff
@@ -14,9 +14,24 @@
...
+    """Gets the current UTC datetime."""
```
jonparrott: Fixed.

google/auth/_helpers.py
```diff
@@ -14,9 +14,24 @@
...
+    return datetime.datetime.utcnow()
+
+
+def datetime_to_secs(value):
+    """Convert a datetime object to the number of seconds since the UNIX epoch.
+    """
```
jonparrott: Optional, the args and return type are obvious here, no?

(If you disagree, I'm happy to add them, but it'll be rather redundant).

google/auth/jwt.py
```diff
@@ -0,0 +1,233 @@
...
+import base64
+import collections
+import json
+
+from google.auth import crypt
+from google.auth import _helpers
+
+
+_DEFAULT_TOKEN_LIFETIME_SECS = 3600  # 1 hour in sections
+CLOCK_SKEW_SECS = 300  # 5 minutes in seconds
```
jonparrott: The user may want to change the clock skew amount for servers that can potentially have large skews.

The other is private as it'll be a default argument value.

google/auth/jwt.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def encode(signer, payload, header=None, key_id=None):
+    """Make a signed JWT.
+
+    Args:
+        signer (google.auth.crypt.Signer): The signer used to sign the JWT.
```
jonparrott: It'll implement the same interface as `Signer`, I assume?

google/auth/jwt.py
```diff
@@ -0,0 +1,233 @@
...
+        payload (Mapping): The JWT payload.
+        header (Mapping): Additional JWT header payload.
+        key_id (str): The key id to add to the JWT header. If the
+            signer has a key id it will be used as the default. If this is
+            specified, it'll override the signer's key id.
+
+    Returns:
+        bytes: The encoded JWT.
+    """
+    if not header:
```

google/auth/jwt.py
jonparrott: It leaves the key id all together. I've updated the docstring.

google/auth/jwt.py
jonparrott: Yes, but I prefer to use that when I don't know the type - e.g., with user input.

google/auth/jwt.py
jonparrott: done.

google/auth/jwt.py

google/auth/jwt.py
jonparrott: Done.

tests/test_jwt.py
jonparrott: Based on code from oauth2client, date stands.

tests/test_jwt.py
jonparrott: Yes, this is the one area of duplication. Again, up to you if you want me to move to a shared module.

tests/test_jwt.py
jonparrott: Pytest fixtures: http://doc.pytest.org/en/latest/fixture.html

The best feature of pytest.

google/auth/_helpers.py
dhermes: Sorry I meant `Returns` section of docstring

google/auth/_helpers.py
dhermes: Neither are obvious. I honestly can't say what either are without Googling

google/auth/jwt.py
dhermes: That feels wrong overloading `key_id` to perform two functions. Can you introduce a second keyword to disable it altogether? Also, why would someone want to disable it altogether? How common will this be?

google/auth/jwt.py
dhermes: Seems we disagree on the utility of `from_bytes` / `to_bytes`. For me it's a Python 2/3 agnostic way to say: "this is bytes" and clearly communicate that to other maintainers.

google/auth/_helpers.py
jonparrott: done.

google/auth/_helpers.py
jonparrott: Add docstrings, redundancy be damned.

google/auth/jwt.py
jonparrott: So would you rather make it private?

google/auth/jwt.py
jonparrott: So is there anything to change?

google/auth/jwt.py
jonparrott: This doesn't really seem unusual:

* `None` is the default, which just picks up the key id from the signer.
* `False` excludes the key id altogether.
* Any other value becomes the key id.

I'm not sure how common it'll be.

google/auth/jwt.py
jonparrott: We do. I prefer to use those functions to say "The caller can pass in a str or a bytes, I need it to be a certain type.". If the type is already know (like the return value of json.dumps), then it's unnecessary indirection.

google/auth/_helpers.py
jonparrott: done.

google/auth/jwt.py
dhermes: I think so. "You can always cut more, but you can never cut less"

google/auth/jwt.py
dhermes: Well do you have a plan to implement an abstract `Signer` base class?

google/auth/jwt.py
dhermes: > I'm not sure how common it'll be.

Then why have the feature at all?

google/auth/jwt.py
jonparrott: Not particularly. At this point I'm unsure if I'm going to implement an appengine Signer class, and if I do I'm not entirely convinced that have one other signer necessitates an abstract base class. Python is not Java and abstract interfaces are not required, but they can be useful to describe a contract to users of the library (hence why I think they're useful in the case of credentials). In any case, can we defer this conversation until we do actually have another signer to implement?

google/auth/jwt.py
jonparrott: Having a key id will be common, disabling the key id probably won't. It's more statement and one more test, and the behavior isn't terribly surprising, but if you feel strongly I'll remove it.

google/auth/jwt.py
dhermes: SGTM. FWIW the GAE signer should be trivial because of `app_identity.sign_blob`

google/auth/jwt.py
dhermes: I "feel strongly" that allowing the variable to be 3 types (null, str, bool) is not a good API

google/auth/jwt.py
jonparrott: I'm not sure what you're after here. I don't feel like there's anything unclear about this statement:

```base64.urlsafe_b64encode(json.dumps(header).encode('utf-8'))```

that this statement improves on:

```base64.urlsafe_b64encode(_helpers.to_bytes(json.dumps(header)))```

The first statement is clear to me: `json.dumps` returns a string, but `urlsafe_b64encode` requires bytes so it must be encoded. The types are known. `_helpers.to_bytes` is there to take an 'uknown' type and make it a known type.

If you really feel so strong I can change this to use `_helpers`, but I'd rather not.

google/auth/jwt.py
jonparrott: I will remove it then.

google/auth/jwt.py
jonparrott: Done.

google/auth/jwt.py
jonparrott: Done.

google/auth/jwt.py
dhermes: It's fine.

dhermes:
LGTM?

jonparrott:
Thanks for reviewing, @dhermes. :)

PR #8, opened 2016-10-04
Add google.auth.credentials - the public interfaces for credentials

jonparrott:
@nathanielmanistaatgoogle this is the section you indicated you wanted to give input on. PTAL.

### DIFF
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -123,10 +123,16 @@ class Scoped(object):
     OAuth 2.0-based credentials allow limiting access using scopes as described
     in `RFC6749 Section 3.3`_.
     If a credential class implements this interface then the credentials either
-    require or use scopes in their implementation.
+    use scopes in their implementation.
 
-    Credentials that require scopes to obtain access tokens must either be
-    constructed with scopes::
+    Some credentials require scopes in order to obtain a token. You can check
+    if scoping is necessary with :attr:`requires_scopes`::
+
+        if credentials.requires_scopes:
+            # Scoping is required.
+            credentials = credentials.create_scoped(['one', 'two'])
+
+    Credentials that require scopes must either be constructed with scopes::
 
         credentials = SomeScopedCredentials(scopes=['one', 'two'])
 
@@ -134,11 +140,8 @@ class Scoped(object):
 
         scoped_credentials = credentials.with_scopes(scopes=['one', 'two'])
 
-    Some credentials have scopes but do not allow or require scopes to be set.
-    You can check if scoping is necessary with :attr:`requires_scopes`::
-
-        if credentials.requires_scopes:
-            credentials = credentials.create_scoped(['one', 'two'])
+    Some credentials have scopes but do not allow or require scopes to be set,
+    these credentials can be used as-is.
 
     .. _RFC6749 Section 3.3: https://tools.ietf.org/html/rfc6749#section-3.3
     """
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -144,15 +144,12 @@ class ScopedCredentials(object):
     """
     def __init__(self):
         super(ScopedCredentials, self).__init__()
-        self.__scopes = None
+        self._scopes = None
 
     @property
-    def _scopes(self):
-        return self.__scopes
-
-    @_scopes.setter
-    def _scopes(self, value):
-        self.__scopes = _helpers.string_to_scopes(value)
+    def scopes(self):
+        """Sequence[str]: the credentials' current set of scopes."""
+        return self._scopes
 
     @abc.abstractproperty
     def requires_scopes(self):
@@ -165,8 +162,7 @@ class ScopedCredentials(object):
         """Create a copy of these credentials with the specified scopes.
 
         Args:
-            scopes (Union[str, Sequence]): The scope or list of scopes to
-                request.
+            scopes (Sequence[str]): The list of scopes to request.
 
         Raises:
             NotImplementedError: If the credentials' scopes can not be changed.
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -65,7 +65,7 @@ class Credentials(object):
         Note that credentials can be invalid but not expired becaue Credentials
         with :attr:`expiry` set to None is considered to never expire.
         """
-        now = _helpers.now()
+        now = _helpers.utcnow()
         if self.expiry is None or self.expiry > now:
             return False
         else:
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -124,7 +124,7 @@ class ScopedCredentials(object):
     """Interface for scoped credentials.
 
     OAuth 2.0-based credentials allow limiting access using scopes as described
-    in `RFC6749 <https://tools.ietf.org/html/rfc6749#section-3.3>`__.
+    in `RFC6749 Section 3.3`_.
     If a credential class implements this interface then the credentials either
     require or use scopes in their implementation.
 
@@ -143,6 +143,7 @@ class ScopedCredentials(object):
         if credentials.requires_scopes:
             credentials = credentials.create_scoped(['one', 'two'])
 
+    .. _RFC6749 Section 3.3: https://tools.ietf.org/html/rfc6749#section-3.3
     """
     def __init__(self):
         super(ScopedCredentials, self).__init__()
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -117,7 +117,7 @@ class Credentials(object):
 
 
 @six.add_metaclass(abc.ABCMeta)
-class ScopedCredentials(object):
+class Scoped(object):
     """Interface for scoped credentials.
 
     OAuth 2.0-based credentials allow limiting access using scopes as described
@@ -143,7 +143,7 @@ class ScopedCredentials(object):
     .. _RFC6749 Section 3.3: https://tools.ietf.org/html/rfc6749#section-3.3
     """
     def __init__(self):
-        super(ScopedCredentials, self).__init__()
+        super(Scoped, self).__init__()
         self._scopes = None
 
     @property
@@ -184,7 +184,7 @@ class ScopedCredentials(object):
 
 
 @six.add_metaclass(abc.ABCMeta)
-class SigningCredentials(object):
+class Signing(object):
     """Interface for credentials that can cryptographically sign messages."""
 
     @abc.abstractmethod
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -92,7 +92,7 @@ class Credentials(object):
             token (Optional[str]): If specified, overrides the current access
                 token.
         """
-        headers[b'authorization'] = 'Bearer {}'.format(
+        headers['authorization'] = 'Bearer {}'.format(
             _helpers.from_bytes(token or self.token))
 
     def before_request(self, request, method, url, headers):
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -50,6 +50,16 @@ class Credentials(object):
         If this is None, the token is assumed to never expire."""
 
     @property
+    def expired(self):
+        """Checks if the credentials are expired.
+
+        Note that credentials can be invalid but not expired becaue Credentials
+        with :attr:`expiry` set to None is considered to never expire.
+        """
+        now = _helpers.utcnow()
+        return self.expiry is not None and self.expiry <= now
+
+    @property
     def valid(self):
         """Checks the validity of the credentials.
 
@@ -58,25 +68,12 @@ class Credentials(object):
         """
         return self.token is not None and not self.expired
 
-    @property
-    def expired(self):
-        """Checks if the credentials are expired.
-
-        Note that credentials can be invalid but not expired becaue Credentials
-        with :attr:`expiry` set to None is considered to never expire.
-        """
-        now = _helpers.utcnow()
-        if self.expiry is None or self.expiry > now:
-            return False
-        else:
-            return True
-
     @abc.abstractmethod
     def refresh(self, request):
         """Refreshes the access token.
 
         Args:
-            request (google.auth.transport.Request): A callable used to make
+            request (google.auth.transport.Request): The object used to make
                 HTTP requests.
 
         Raises:
@@ -105,7 +102,7 @@ class Credentials(object):
         apply the token to the authentication header.
 
         Args:
-            request (google.auth.transport.Request): A callable used to make
+            request (google.auth.transport.Request): the object used to make
                 HTTP requests.
             method (str): The request's HTTP method.
             url (str): The request's URI.
@@ -181,7 +178,7 @@ class ScopedCredentials(object):
     def has_scopes(self, scopes):
         """Checks if the credentials have the given scopes.
 
-        .. warning: This method is not guarenteed to be accurate if the
+        .. warning: This method is not guaranteed to be accurate if the
             credentials are :attr:`~Credentials.invalid`.
 
         Returns:
@@ -202,7 +199,7 @@ class SigningCredentials(object):
             message (bytes): The message to sign.
 
         Returns:
-            bytes: The messages cryptographic signature.
+            bytes: The message's cryptographic signature.
         """
         # pylint: disable=missing-raises-doc,redundant-returns-doc
         # (pylint doesn't recognize that this is abstract)
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -1,0 +1,208 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Interfaces for credentials."""
+
+import abc
+
+import six
+
+from google.auth import _helpers
+
+
+@six.add_metaclass(abc.ABCMeta)
+class Credentials(object):
+    """Base class for all credentials.
+
+    All credentials have a :attr:`token` that is used for authentication and
+    may also optionally set an :attr:`expiry` to indicate when the token will
+    no longer be valid.
+
+    Most credentials will be :attr:`invalid` until :meth:`refresh` is called.
+    Credentials can do this automatically before the first HTTP request in
+    :meth:`before_request`.
+
+    Although the token and expiration will change as the credentials are
+    :meth:`refreshed <refresh>` and used, credentials should be considered
+    immutable. Various credentials will accept configuration such as private
+    keys, scopes, and other options. These options are not changeable after
+    construction. Some classes will provide mechanisms to copy the credentials
+    with modifications such as :meth:`ScopedCredentials.with_scopes`.
+    """
+    def __init__(self):
+        self.token = None
+        """str: The bearer token that can be used in HTTP headers to make
+        authenticated requests."""
+        self.expiry = None
+        """Optional[datetime]: When the token expires and is no longer valid.
+        If this is None, the token is assumed to never expire."""
+
+    @property
+    def valid(self):
+        """Checks the validity of the credentials.
+
+        This is True if the credentials have a :attr:`token` and the token
+        is not :attr:`expired`.
+        """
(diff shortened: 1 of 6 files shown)


jonparrott:
I am unsure if I should leave the `scopes` property as is (completely private) or make it a public read-only property.

dhermes:
I vote for public read-only. I'm holding off a review here until you tell me to look @jonparrott

jonparrott:
Sgtm, gonna let Nathaniel chime in first.

jonparrott:
@nathanielmanistaatgoogle ping, this is now blocking.

dhermes:
@jonparrott rebase?

jonparrott:
@dhermes doing it now.

jonparrott:
Rebased.

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @property
+    def expired(self):
+        """Checks if the credentials are expired.
+
+        Note that credentials can be invalid but not expired becaue Credentials
+        with :attr:`expiry` set to None is considered to never expire.
+        """
+        now = _helpers.utcnow()
+        if self.expiry is None or self.expiry > now:
```
nathanielmanistaatgoogle: Why shouldn't these four lines be `return self.expiry is not None and self.expiry <= now`?

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def apply(self, headers, token=None):
+        """Apply the token to the authentication header.
+
+        Args:
+            headers (Mapping): The HTTP request headers.
+            token (Optional[str]): If specified, overrides the current access
+                token.
+        """
+        headers[b'authorization'] = 'Bearer {}'.format(
```
nathanielmanistaatgoogle: How certain are we that a native string is appropriate for `'Bearer {}'`?

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+        If this is None, the token is assumed to never expire."""
+
+    @property
+    def valid(self):
+        """Checks the validity of the credentials.
+
+        This is True if the credentials have a :attr:`token` and the token
+        is not :attr:`expired`.
+        """
+        return self.token is not None and not self.expired
```
nathanielmanistaatgoogle: Since `self.expired` is used here, move this below where `expired` is defined?

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+            NotImplementedError: If the credentials' scopes can not be changed.
+                This can be avoided by checking :attr:`requires_scopes` before
+                calling this method.
+        """
+        raise NotImplementedError('This class does not require scoping.')
+
+    def has_scopes(self, scopes):
+        """Checks if the credentials have the given scopes.
+
+        .. warning: This method is not guarenteed to be accurate if the
```
nathanielmanistaatgoogle: Typo "guarenteed".

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @abc.abstractmethod
+    def sign_bytes(self, message):
+        """Signs the given message.
+
+        Args:
+            message (bytes): The message to sign.
+
+        Returns:
+            bytes: The messages cryptographic signature.
```
nathanielmanistaatgoogle: Missing apostrophe? "messages" is meant to be possessive, not plural, right?

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+        """True if these credentials require scopes to obtain an access token.
+        """
+        return False
+
+    @abc.abstractmethod
+    def with_scopes(self, scopes):
+        """Create a copy of these credentials with the specified scopes.
+
+        Args:
+            scopes (Union[str, Sequence]): The scope or list of scopes to
```
nathanielmanistaatgoogle: How strong is the need for ad hoc polymorphism here? How burdensome would it be to require applications to know whether or not they have a single string or a sequence of strings?

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+        # pylint: disable=unused-argument
+        # (Subclasses may use these arguments to ascertain information about
+        # the http request.)
+        if not self.valid:
+            self.refresh(request)
+        self.apply(headers)
+
+
+@six.add_metaclass(abc.ABCMeta)
+class ScopedCredentials(object):
```
nathanielmanistaatgoogle: I'm having a hard time understanding this class and at least the largest up-front part of that is that I'm not yet seeing how `self.__scopes` can ever be set to something other than `None`. Is there a `self._scopes =` line I'm missing somewhere?

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+            return False
+        else:
+            return True
+
+    @abc.abstractmethod
+    def refresh(self, request):
+        """Refreshes the access token.
+
+        Args:
+            request (google.auth.transport.Request): A callable used to make
```
nathanielmanistaatgoogle: If you've already specified the type of this parameter to be `google.auth.transport.Request`, is it redundant or less specific to also describe the parameter as a `callable`?

google/auth/credentials.py
```diff
@@ -0,0 +1,209 @@
...
+    Although the token and expiration will change as the credentials are
+    :meth:`refreshed <refresh>` and used, credentials should be considered
+    immutable. Various credentials will accept configuration such as private
+    keys, scopes, and other options. These options are not changeable after
+    construction. Some classes will provide mechanisms to copy the credentials
+    with modifications such as :meth:`ScopedCredentials.with_scopes`.
+    """
+    def __init__(self):
+        self.token = None
+        """str: The bearer token that can be used in HTTP headers to make
```
nathanielmanistaatgoogle: Is there a URL at which I can view how the doc strings in this code are rendered? Or is that not possible for unmerged code in a pull request?

jonparrott: There's a render of the prototype's docs [here](http://goth-experimental.readthedocs.io/reference/google.auth.credentials.html), but I can stage a new render if you want. I'm currently getting a sphinx extension through OSPO that properly handles the typing parameters.

google/auth/credentials.py
jonparrott: Done.

google/auth/credentials.py
jonparrott: No reason, done.

google/auth/credentials.py
jonparrott: Yes, changed to 'object'.

google/auth/credentials.py
jonparrott: They should be, and it'll be verified with the transport compliance tests.

google/auth/credentials.py
jonparrott: Scopes are set by a concrete class in a way that is specific to the concrete class, some examples:

1. [`service_account.Credentials`](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/prototype/google/oauth2/service_account.py#L194) has scopes that can be changed with `with_scopes`.
2. [`oauth2.credentials.Credentials`](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/prototype/google/oauth2/credentials.py#L57) has scopes that can only be set at creation time.

google/auth/credentials.py
jonparrott: If you feel strongly, I can definitely make this require `Sequence[str]` always.

google/auth/credentials.py
jonparrott: Fixed.

google/auth/credentials.py
jonparrott: Yep. Fixed.

google/auth/credentials.py
nathanielmanistaatgoogle: That helps a lot and I look forward to the typing parameters; thanks.

google/auth/credentials.py
jonparrott: Hah, done. I also took @dhermes' recommendation of making `scopes` a public read-only property.

google/auth/credentials.py
nathanielmanistaatgoogle: Ahh. Aha! This might be the best place to attempt [one of the concrete recommendations for which you asked](https://github.com/google/oauth2client/issues/597#issuecomment-252756551).

What's confusing me and worrying me about this class right now is that it is a public API element that appears intended to serve both internal and external audiences - its documentation clearly speaks to developer-users of the library, but it also has private and double-private code elements intended for use by other parts of the library. The single-underscore code elements threw me because (1) I only ever use single-underscore to mean private-to-a-code-element (the class, in this case) and the single-underscore code elements they aren't enough on their own to add up to useful behavior within just the class, and (2) I'm aware that others use single-underscore to mean visible-to-subclasses-like-Java's-`protected` but the class isn't documented sufficiently for subclassing. Double-underscore I see rarely enough and it always seems to mean "really really private".

Reading the rendered documentation really helps this class feel more like a completely-abstract interface. **But:** this is open source, and the documentation includes links to the source (except for `requires_scopes`; that's weird), so I wonder: are developer-users of the library going to inherit from this class, call inherited concrete behavior, and get upset when that inherited concrete behavior changes? That's where I think there's accidental leakage - the class is in the library's public API, its methods are class's public API and therefore in the library API, and yet its methods have concrete implementations intended only for library-internal use and those concrete implementations are totally accessible through the API.

My specific recommendation would be to express this class as
```
@six.add_metaclass(abc.ABCMeta)
class ScopedCredentials(object):
    """Interface for scoped credentials.

    OAuth 2.0-based credentials allow limiting access using scopes as described
    in `RFC6749 Section 3.3`_.

    If a credential class implements this interface then the credentials either
    require or use scopes in their implementation.

    Some credentials have scopes but do not allow or require scopes to be set.

    You can check if scoping is necessary with :attr:`requires_scopes`::
        if credentials.requires_scopes:
            credentials = credentials.create_scoped(['one', 'two'])

    .. _RFC6749 Section 3.3: https://tools.ietf.org/html/rfc6749#section-3.3
    """

    @abc.abstractmethod
    def requires_scopes(self):
        """Return True if these credentials require scopes to obtain an access token.
        """
        raise NotImplementedError()

    @abc.abstractmethod
    def with_scopes(self, scopes):
        """Create a copy of these credentials with the specified scopes.

        Args:
            scopes (Union[str, Sequence]): The scope or list of scopes to
                request.

        Returns:
          A `ScopedCredentials` equal to this object but for the additional scopes.
        """
        raise NotImplementedError()

    @abc.abstractmethod
    def has_scopes(self, scopes):
        """Checks if the credentials have the given scopes.

        .. warning: This method is not guaranteed to be accurate if the
            credentials are :attr:`~Credentials.invalid`.

        Returns:
            bool: True if the credentials have the given scopes.
        """
        raise NotImplementedError()
```
.

I have a feeling that implementations of this interface only make sense if they are also implementations of `Credentials`. If that's the case, and `Credentials` also has its undocumented-but-accidentally-publicly-accessible concrete elements removed, then should `ScopedCredentials` subclass `Credentials`? It's fine for one interface to extend another, right?

Responding to "There's no part of the credentials that I expect to be implemented by the consuming application - implementations are all first party but the interfaces are for the benefit of the users": Over a long period of time I have had little success presenting a class in a public API, documenting that it is for me to subclass and for users to merely use, and having users respect that boundary and not subclass the class. That's part of what's led me to the place of no-classes-with-concrete-elements-in-my-APIs: if a class is a pure interface, users can subclass it to their heart's content without depending on any of my concrete behavior or getting bothered when my internals change. You don't support users subclassing `ScopedCredentials` in its current form, but your currently-drafted documentation doesn't expressly say "only classes in the library are allowed to subclass this class". Given that after my first reading of its code my reaction was "it looks like I'm supposed to subclass this but I don't see enough in the documentation about how to do so", how many others will react the same way?

Responding to "Everything in credentials is abstract and is intended to tell the developer what can be done with a specific credential instance": as currently drafted, users are free to subclass `ScopedCredentials`, to call methods like `requires_scopes`, and to build on that concrete behavior that is not supported for them.

A secondary concern, from a future maintainer's perspective: with all else held equal (particularly the number of code elements in the public API and their size), smaller total code elements and more of them are better than fewer, larger total code elements. You don't seem to have to privately partially implement your interface in the same code element (`class ScopedCredentials`) in which you publicly define your interface, so that alone is an argument against doing so.

If you're in the mood for levity, the idea that a code element (particularly a class) should be either _entirely in_ or _entirely out_ of the API of the component in which it is placed (rather than having a mix of some subelements included in the API and some subelements intended to be used only within the component) is what I've taken to calling "The Hokey Pokey API Design Principle". If you're not in the mood for levity, I hope it's still a good principle.

jonparrott: > re developer-users of the library going to inherit from this class, call inherited concrete behavior, and get upset when that inherited concrete behavior changes? 

That's not intended, as mentioned before, developer-users of this library are intended to be able to ask if a instance implements this interface and thus can then make assumptions about how to use the instance.

>  should ScopedCredentials subclass Credentials?

No, because if a concrete class extends from both `SigningCredentials` and `ScopedCredentials` we'll have the diamond inheritance problem.

> Given that after my first reading of its code my reaction was "it looks like I'm supposed to subclass this but I don't see enough in the documentation about how to do so", how many others will react the same way?

Not sure, but I can do a better job of explaining this in the modules docstring, and indeed I'll go into great detail about this in the docstring for application default credentials.

Overall, you seem mostly fine with this as long as I get rid of any concrete elements. I'm mostly okay with that, but how do you propose I deal with stuff like `has_scopes` which will be identical for all first-party implementations and could be easy to screw up? How do you propose similarly for `Credentials` properties `expired`, and `valid`? I'm not in the game of repeating code all over the place for the sake of purity.

nathanielmanistaatgoogle: I'm not following the diamond inheritance problem worry. If all three interfaces contain no concrete elements (not even private ones), there's no diamond inheritance problem, right? It's not like what would be the two subinterfaces have conflicting method names.

Yes, I am mostly fine with this as long as the concrete elements are dropped from the interfaces.

For common behavior, implement it once in one (private) place and delegate to it from the n places where it is exposed. Have it directly tested by an access-to-implementation test and have it indirectly tested by the API-surface-level tests that test each place where it is exposed. Yes, these extra connections are [forwarding methods](https://en.wikipedia.org/wiki/Composition_over_inheritance#Drawbacks) that in isolation would be considered redundant cruft but in the context of _allowing you to separate the way your system is presented from the way it is implemented_ are a price well worth paying. Absolutely don't repeat the executable statements that form the actual useful logic of the library; that would be bloat as well as inviting bugs and inconsistency to creep in over time.

It may be the case that you choose for the private implementation of the library classes very similar to those originally drafted for the library's public API. Should that come to pass, I won't raise much of a fuss since they'll be all private implementation.

jonparrott: > I'm not following the diamond inheritance problem worry. If all three interfaces contain no concrete elements (not even private ones), there's no diamond inheritance problem, right? It's not like what would be the two subinterfaces have conflicting method names.

Yes but the MRO would still be ... weird. I'd like to avoid it if possible.

> For common behavior, implement it once in one (private) place and delegate to it from the n places where it is exposed.

I could be on board with this but IMO it's separation without much purpose. The thing I really dislike about it how how the documentation will present the base classes:

With the current implementation:

screen shot 2016-10-11 at 4 58 04 pm

With inheriting from a private concrete class:

screen shot 2016-10-11 at 4 59 49 pm

Any recommendations here? Failing that, is there any way you could be comfortable with keeping some of the "concreteness" around here? One thing I can do is mark *everything* as `abstractmethod` or `abstractproperty` and that forces all subclasses to either implement or explicitly call super.

jonparrott: @elibixby I'm more concerned about the logic/constraints in the `Credentials` class, notably:

1. If expired is None, the credentials do not expire.
2. Credentials are only valid if they have a token and are not expired.

elibixby: Yes. I agree that it's unnecessary for all the concreteness should be removed from the parent classes. The way you have it they are neatly separated by functionality, and the common concrete elements seem reasonable in scope to me.

nathanielmanistaatgoogle: Regarding the MRO: that hadn't occurred to me; it's been so long since I wrote anything that has one concrete method implementation override another concrete method implementation. I suppose I forgot that people did that. It makes code shorter, but doesn't it make it harder to reason about?

Regarding separation without much purpose: yes, it's separation without much purpose _now_. The purpose is to remain able to respond to the needs of the future by having made minimal promises in the past. That's why I'm advocating for all this extra consideration only on the API surface of the library; it is more code and more abstraction and there would be no point internally where everything remains free for all time to be changed in every way.

Regarding how classes are presented: yes, it is an abstraction violation to have a public class inherit from a private class. Why have concrete classes in the API at all? If you already have an interface that describes the full behavior that the object passed from the library to the application will support, why not just have a function that returns an instance of that interface? Something like
```py
class Credentials:
  # some public abstract method specifications

class ScopedCredentials(Credentials):
  # some more public abstract method specifications

def fixed_scoped_credentials(token, scopes):
  """<summary string>

  Args:
    <parameter descriptions>

  Returns:
    A ScopedCredentials that <more information>.
  """
  import _private_implementation_module
  return _private_implementation_module.FixedScopedCredentials(token, scopes)
```
? Including a public class in an API is a big commitment - much bigger than including an interface describing what the API user may do with a type and one or more functions available for constructing objects that implement the type. Does that big commitment need to be made? Or is it just so tantalizingly straightforward and direct?

You may be worried about the situation of having presented several interfaces and needing to pass out of the library objects that implement several of the interfaces - it would be truthful and complete to say "just describe the value as an instance of all the interfaces it implements" but in practice I've met friction doing this. An alternative is to define extra interfaces that define no methods of their own but by inheriting from two or more others serve to specify a union of them - in theory this leads to exponential empty interface writing but in practice it leads to one or two. Will `google-auth-library-python` emit any object that is a `ScopedCredentials` but not also a `Credentials`? If not, should `ScopedCredentials` inherit from `Credentials`? Will the library emit any object that is a `SigningCredentials` but not also a `Credentials`? If not, should `SigningCredentials` be a subinterface of `Credentials`? Will the library emit any value that is both a `SigningCredentials` and a `ScopedCredentials`? If so, should a `ComboCredentials(ScopedCredentials, SigningCredentials)` be defined?

Sorry that I'm dragging this out and probably not advocating well enough. Years of experience has led me to believe that these questions of interface design are crucially important, and I'm always amazed by how few other people's experiences lead them to the same perspective. Maybe I've made more mistakes along the way than I've realized...

jonparrott: Ah! I think I finally I see what you're after, you're essentially wanting to hide the concrete classes altogether and only expose the public interfaces, so we'd end up with something like this:

```python

class PublicInterface(object):
    pass

# may or may not be needed, but if it is, this is where it'll be.
class _PrivateCommonConcreteStuff(PublicInterface):
   pass

class _PrivateConcrete(_PrivateCommonConcreteStuff):
   pass

def constructor_for_thing():
    return _PrivateConcrete()

```


This sort of implementation hiding seems way over the top and unpythonic to me. Granted, I'm not (nor have ever claimed to be) an expert in object-oriented design, but it seems really unintuitive as a user of the library. It's also somewhat unintuitive as the maintainer of the library (I mean, it took you no less than 3 very long comments to express this to me, and I'm still unsure if I'm 100% clear on it).

I'm happy to meet in person to discuss, or if you feel really confident that there's no way I'll talk you out of this then I'll relent and implement it as above. I'd also like to hear @dhermes's thoughts on it.

Some specific responses:

> Including a public class in an API is a big commitment - much bigger than including an interface describing what the API user may do with a type and one or more functions available for constructing objects that implement the type. 

This seems like a viewpoint coming from a Java-centric world. Not saying it's wrong or that it doesn't have advantages (it does, and I see that now), but it does feel unfamiliar in Python.

>  An alternative is to define extra interfaces that define no methods of their own but by inheriting from two or more others serve to specify a union of them

I get what you're saying here, but in terms of actual usage the consumers of this library aren't going to ask 'Is this a ComboCredential'? They're going to say 'Can this credential sign bytes' - so the combined interfaces would just be for the benefit of the docstrings. Perhaps it'll be better if I renamed `ScopedCredentials` and `SigningCredentials` to `HasScopes` and `CanSign` (I was thinking of doing so anyway).

dhermes: > I'd also like to hear @dhermes's thoughts on it.

I am **very** against non-public objects that users are guaranteed to have instances of / see in stacktraces.

jonparrott: Discussed with @dhermes offline and settled on `Scoped` and `Signing` as the names for the mixin classes.

jonparrott:
@nathanielmanistaatgoogle after discussing with @dhermes offline, I'm gonna go ahead and merge this so we can unblock reviewing some concrete classes.

We can still continue to discuss here and I'll send follow-up PRs to make changes. This is all fluid until 1.0. Unless there is a very significant change I don't anticipate the impact on concrete classes to be large.

PR #10, opened 2016-10-06
Add http.client transport

jonparrott:


google/auth/transport/_http_client.py
```diff
@@ -0,0 +1,97 @@
...
+        # http.client doesn't allow None as the headers argument.
+        if headers is None:
+            headers = {}
+
+        # http.client needs the host and path parts specified separately.
+        parts = urllib.parse.urlsplit(url)
+        path = urllib.parse.urlunsplit(
+            ('', '', parts.path, parts.query, parts.fragment))
+
+        connection = http_client.HTTPConnection(parts.netloc)
```
dhermes: What if `parts.scheme == 'https'`?

tests/transport/compliance.py
```diff
@@ -0,0 +1,89 @@
...
+NXDOMAIN = 'test.invalid'
+
+
+class RequestResponseTests(object):
+
+    @pytest.fixture
+    def server(self):
+        """Provides a test HTTP server that is automatically created before
+        a test and destroyed at the end. The server is serving a test
+        application that can be used to verify requests."""
```
dhermes: Regarding my dogma. This could easily be expressed as follows

```python
        """Provides a test HTTP server.

        The test server is automatically created before
        a test and destroyed at the end. The server is serving a test
        application that can be used to verify requests.
        """
```

tests/transport/compliance.py
```diff
@@ -0,0 +1,89 @@
...
+        # pylint: disable=unused-variable
+        # (pylint thinks the flask routes are unusued.)
+        app = flask.Flask(__name__)
+        app.debug = True
```
dhermes: Wouldn't `app.debug` make it not unused? Also when do you `enable=`?

tests/transport/compliance.py
```diff
@@ -0,0 +1,89 @@
...
+
+        @app.route('/basic')
+        def index():
+            return 'Basic Content', http_client.OK, {
+                'X-Test-Header': flask.request.headers.get(
+                    'x-test-header', 'value')}
```
dhermes: Hard to parse visually. Create the dict before the return?

google/auth/transport/_http_client.py
```diff
@@ -0,0 +1,97 @@
...
+
+from six.moves import http_client
+from six.moves import urllib
+
+from google.auth import exceptions
+from google.auth import transport
+
+
+class Response(transport.Response):
+    """http.client transport request adapter."""
```
jonparrott: Done.

google/auth/transport/_http_client.py
```diff
@@ -0,0 +1,97 @@
...
+    def __init__(self, response):
+        self._status = response.status
+        self._headers = {
+            key.lower(): value for key, value in response.getheaders()}
+        self._data = response.read()
+
+    @property
```
jonparrott: Unfortunately abc doesn't detect this as properly implementing the interface.

(also, I'm all for these being read-only).

google/auth/transport/_http_client.py
```diff
@@ -0,0 +1,97 @@
...
+    def data(self):
+        return self._data
+
+
+class Request(transport.Request):
+    """http.client transport request adapter."""
+    def __init__(self):
+        pass
```
jonparrott: It isn't. Fixed.

google/auth/transport/_http_client.py
```diff
@@ -0,0 +1,97 @@
...
+            kwargs: Additional arguments passed throught to the underlying
+                :meth:`~http.client.HTTPConnection.request` method.
+
+        Returns:
+            Response: The HTTP response.
+
+        Raises:
+            google.auth.exceptions.TransportError: If any exception occurred.
+        """
+        # socket._GLOBAL_DEFAULT_TIMEOUT is the default in http.client.
```
jonparrott: Is that in anyway cleaner than `socket._GLOBAL_DEFAULT_TIMEOUT`?

google/auth/transport/_http_client.py
jonparrott: Not supported by this. Do you want me to make it explicitly raise if the scheme is https?

tests/transport/compliance.py
jonparrott: Done, thanks for suggesting an alternative.

tests/transport/compliance.py
jonparrott: It's not `app` that's being marked as unused, it's `index` and `server_error`.

I don't `enable=`. Pylint has block scoping on options and I see no reason to clutter up things with enables at the end of the block.

tests/transport/compliance.py
jonparrott: Done.

google/auth/transport/_http_client.py
dhermes: I wasn't trying to indicate that you'd use that snippet, just to indicate that the default is not totally inaccessible. In Python 3.3, `inspect` added some nice functionality for this:

```python
>>> import inspect
>>> def f(a, b=1, c=2):
...     return (a, b, c)
...
>>> inspect.signature(f).parameters['b'].default
1
```

jonparrott: I see, I'd like to leave as-is.

tests/transport/compliance.py
jonparrott: > By not closing, future edits to this code may change your original intent. Explicitly declaring your intent is better than implicitly letting the close of the block indicate it. Feel like I've heard that before?

I'm not sure I understand what you mean here? Can you demonstrate a case where manually, explicitly doing what block scoping does automatically will prevent future issues? If so, I can most likely be convinced.

> You can avoid the disable if you just define the functions as top-level

I could, but it seems unidiomatic from the perspective of those familiar with how flask works. It also puts functions that are *very much* implementation details of this fixture at the top-level. If you feel very strongly about it, I'll move them.

dhermes: > Can you demonstrate

```diff
diff --git a/foo.py b/foo.py
index a3b3035..aeaddea 100644
--- a/foo.py
+++ b/foo.py
@@ -1,9 +1,12 @@
 def my_func():
     app = app_factory_bean_factory()
 
+    this_is_extra_but_pylint_will_yell = 3.14
+
     # pylint: disable=unused-variable
     @app.sprinkles
     def cupcakes():
         return 'sugar'
 
+    haha_oops_debug = 42
     return app
```

----

I don't feel strongly in the sense of not contradicting `flask`, but I dislike closures (lots of indents are hard to read), so that's my bias. I'm surprised that Pylint doesn't have builtin support for common flask situations.

jonparrott: Cool, I agree in this case, so I'll fence off these guys.

pylintrc
```diff
@@ -115,6 +115,7 @@ disable =
     wrong-import-position,
     no-name-in-module,
     locally-disabled,
+    locally-enabled,
```
dhermes: @jonparrott What does this do?

tests/transport/compliance.py
dhermes: Thanks

google/auth/transport/_http_client.py
dhermes: Yes please

google/auth/transport/_http_client.py
dhermes: Ack

dhermes:
Still remaining:

- Me understanding what `locally-enabled` does
- Raising for non-`http` schemes in the `http_client` impl.
- Using `remove=` instead of the `None` means delete behavior in the query params replace (or another alternative based on our discussion)

pylintrc
jonparrott: pylint warns when you using `#pylint: enable` just as when you do `#pylint: disable`, so this is here to quiet that warning.

dhermes: Got it. So this is like me disabling the `I` message. Jinx.

jonparrott:
> Me understanding what locally-enabled does

I think I cleared that up.

> Raising for non-http schemes in the http_client impl.

Done.

> Using remove= instead of the None means delete behavior in the query params replace (or another alternative based on our discussion)

Does the new examples in the docstring help here? Do you feel strongly that remove should be a separate argument?

google/auth/transport/_http_client.py
jonparrott: Done.

dhermes:
> Do you feel strongly that remove should be a separate argument?

Yes. Just because Python let's us allow any-old type in doesn't mean it's a good API for users.

PR #14, opened 2016-10-10
Add urllib3 transport

jonparrott:


google/auth/transport/urllib3.py
```diff
@@ -0,0 +1,91 @@
...
+                urllib3 :meth:`urlopen` method.
+
+        Returns:
+            Response: The HTTP response.
+
+        Raises:
+            google.auth.exceptions.TransportError: If any exception occurred.
+        """
+        # urllib3 uses a sentinel default value for timeout, so only set it if
+        # specified.
```
dhermes: Lulz. I feel like I've reviewed this snippet like 5 times.

jonparrott: At least twice now.

PR #22, opened 2016-10-14
Add compute engine credentials

jonparrott:


google/auth/compute_engine/__init__.py
```diff
@@ -11,3 +11,12 @@
...
 # limitations under the License.
+
+"""Google Compute Engine authentication."""
+
+from google.auth.compute_engine.credentials import Credentials
+
+
+__all__ = [
+    'Credentials'
+]
```
dhermes: tuple?

google/auth/compute_engine/__init__.py
```diff
@@ -11,3 +11,12 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+"""Google Compute Engine authentication."""
+
+from google.auth.compute_engine.credentials import Credentials
+
+
+__all__ = [
+    'Credentials'
```
dhermes: Trailing comma

google/auth/compute_engine/__init__.py
dhermes: Also why define `__all__` when the only thing would be Credentials anyway?

google/auth/compute_engine/credentials.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        info = _metadata.get_service_account_info(
+            request,
+            service_account=self._service_account_email)
+
+        self._service_account_email = info['email']
+        # pylint: disable=attribute-defined-outside-init
+        # (pylint doesn't recognize that this is defined in ScopedCredentials)
+        self._scopes = _helpers.string_to_scopes(info['scopes'])
+        # pylint: enable=attribute-defined-outside-init
```
dhermes: Instead of doing this Pylint disable, just define `self._scopes = None` in the constructor or `_scopes = None` on the class?

tests/compute_engine/test_credentials.py
```diff
@@ -0,0 +1,106 @@
...
+import datetime
+
+import mock
+import pytest
+
+from google.auth import exceptions
+from google.auth.compute_engine import credentials
+
+
+class TestCredentials:
```
dhermes: New-style

tests/compute_engine/test_credentials.py
```diff
@@ -0,0 +1,106 @@
...
+    @pytest.fixture(autouse=True)
+    def credentials(self):
+        self.credentials = credentials.Credentials()
```
dhermes: Method overwrites itself?

tests/compute_engine/test_credentials.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def test_default_state(self):
+        assert not self.credentials.valid
+        # Expiration hasn't been set yet
+        assert not self.credentials.expired
+        # Scopes aren't needed
+        assert not self.credentials.requires_scopes
+
+    def test_create_scoped_fails(self):
+        pass
```
dhermes: Huh?

google/auth/compute_engine/__init__.py
jonparrott: Are you suggesting to just move credentials into here?

google/auth/compute_engine/__init__.py
jonparrott: Done.

google/auth/compute_engine/__init__.py
jonparrott: re: tuple - flake8 doesn't properly recognize it.

google/auth/compute_engine/credentials.py
jonparrott: `self._scopes = None` is set by the `Scoped` mixin. Actually pylint doesn't seem to be complaining about this anymore, so i've just removed that stuff.

tests/compute_engine/test_credentials.py
jonparrott: Indeed it does. It's a fixture that sets a property of the same name. I can rename the fixture if you want, but there's never a need for methods to refer to the fixture directly.

tests/compute_engine/test_credentials.py
jonparrott: I have no explanation. This is covered by the test below, so it's been removed.

google/auth/compute_engine/__init__.py
dhermes: > Are you suggesting to just move credentials into here?

No. I'm suggesting `_all__` be removed since `Credentials` is already the only thing in the namespace.

tests/compute_engine/test_credentials.py
```diff
@@ -0,0 +1,103 @@
...
+import pytest
+
+from google.auth import exceptions
+from google.auth.compute_engine import credentials
+
+
+class TestCredentials(object):
+    @pytest.fixture(autouse=True)
+    def credentials(self):
+        self.credentials = credentials.Credentials()
```
dhermes: Please rename `def credentials` to something else (as in #25)

dhermes:
LGTM (after resolving the `def credentials` rename and whatever you decide on `__all__`)

google/auth/compute_engine/__init__.py
jonparrott: `__all__` is still required because imported members aren't considered part of the public interface.

tests/compute_engine/test_credentials.py
jonparrott: Done.

PR #23, opened 2016-10-14
Add copy_docstring helper

jonparrott:


google/auth/_helpers.py
```diff
@@ -21,6 +21,20 @@
 from six.moves import urllib
 
 
+def copy_docstring(source_class):
+    """Decorator that copies the a method's docstring from another class."""
```
dhermes: No args or returns sections

google/auth/_helpers.py
```diff
@@ -21,6 +21,20 @@
...
+    def decorator(method):
+        """Decorator implementation."""
```
dhermes: No args or returns section.

Also, I recommend using a `__call__`-able class to decorate rather than a closure (to avoid dynamically created functions). Not required that you do this.

tests/test__helpers.py
```diff
@@ -20,6 +20,38 @@
 from google.auth import _helpers
 
 
+class SourceClass(object):
+    def func(self):  # pragma: NO COVER
+        """example docstring"""
+        pass
```
dhermes: Don't need pass here.

tests/test__helpers.py
```diff
@@ -20,6 +20,38 @@
 from google.auth import _helpers
 
 
+class SourceClass(object):
+    def func(self):  # pragma: NO COVER
```
dhermes: You could always just call it just for fun to avoid the `NO COVER`

tests/test__helpers.py
```diff
@@ -20,6 +20,38 @@
...
+        """example docstring"""
+        pass
+
+
+def test_copy_docstring_success():
+    def func():  # pragma: NO COVER
```
dhermes: You could always just call it just for fun to avoid the `NO COVER`

google/auth/_helpers.py
jonparrott: Fixed.

tests/test__helpers.py
jonparrott: that feels more weird than the no cover.

PR #25, opened 2016-10-14
Add service account credentials

jonparrott:


google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+        return cls.from_service_account_info(info, **kwargs)
+
+    @property
+    def requires_scopes(self):
+        """Checks if the credentials requires scopes.
+
+        Returns:
+            bool: True if there are no scopes set otherwise False.
+        """
+        return True if not self._scopes else False
```
dhermes: `self._scopes` will be `None` or empty?

dhermes: Or both?

google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Args:
+            filename (str): The path to the service account json file.
+            kwargs: Additional arguments to pass to the constructor.
+
+        Returns:
+            google.auth.service_account.Credentials: The constructed
+                credentials.
+        """
+        with open(filename, 'r') as json_file:
```
dhermes: I feel shaky about `r` vs. `rb`

google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Returns:
+            bool: True if there are no scopes set otherwise False.
+        """
+        return True if not self._scopes else False
+
+    @_helpers.copy_docstring(credentials.Scoped)
+    def with_scopes(self, scopes):
+        return Credentials(
+            self._signer,
```
dhermes: You are OK having two objects own the same signer?

google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+        expiry = now + lifetime
+
+        payload = {
+            'iat': _helpers.datetime_to_secs(now),
+            'exp': _helpers.datetime_to_secs(expiry),
+            # The issuer must be the service account email.
+            'iss': self._service_account_email,
+            # The audience must be the auth token endpoint's URI
+            'aud': self._token_uri,
+            'scope': _helpers.scopes_to_string(self._scopes or [])
```
dhermes: Why not just let `scopes_to_string` handle the `or []`. Also, in those situations use `or ()`.

google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        # The subject can be a user email for domain-wide delegation.
+        if self._subject:
+            payload.setdefault('sub', self._subject)
```
dhermes: Why use `setedefault`? Did you mean to do it after the additional claims update?

google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+        payload.update(self._additional_claims)
+
+        token = jwt.encode(self._signer, payload)
+
+        return token
+
+    @_helpers.copy_docstring(credentials.Credentials)
+    def refresh(self, request):
+        assertion = self._make_authorization_grant_assertion()
+        access_token, expiry, _ = _client.jwt_grant(
```
dhermes: What's the 3rd return value?

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+with open(os.path.join(DATA_DIR, 'other_cert.pem'), 'rb') as fh:
+    OTHER_CERT_BYTES = fh.read()
+
+SERVICE_ACCOUNT_JSON_FILE = os.path.join(DATA_DIR, 'service_account.json')
+
+with open(SERVICE_ACCOUNT_JSON_FILE, 'r') as fh:
+    SERVICE_ACCOUNT_INFO = json.load(fh)
+
+
+@pytest.fixture
```
dhermes: Isn't there a way to tell `pytest.fixture` to cache the first return value and never call again. In this case it makes sense because you'll always get the same contents.

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+def signer():
+    return crypt.Signer.from_string(PRIVATE_KEY_BYTES, '1')
+
+
+class TestCredentials:
```
dhermes: New-style please.

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+    service_account_email = 'service-account@example.com'
+    token_uri = 'https://example.com/oauth2/token'
+
+    @pytest.fixture(autouse=True)
+    def credentials(self, signer):
+        self.credentials = service_account.Credentials(
```
dhermes: This method overwrites itself?

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@pytest.fixture
+def signer():
+    return crypt.Signer.from_string(PRIVATE_KEY_BYTES, '1')
+
+
+class TestCredentials:
+    service_account_email = 'service-account@example.com'
+    token_uri = 'https://example.com/oauth2/token'
```
dhermes: UPPERCASE the class attrs?

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+        info = dict(SERVICE_ACCOUNT_INFO)
+
+        credentials = service_account.Credentials.from_service_account_info(
+            info, scopes=['email', 'profile'], subject='subject',
+            additional_claims={'meta': 'data'})
+
+        assert credentials._signer.key_id == info['private_key_id']
+        assert credentials._service_account_email == info['client_email']
+        assert credentials._token_uri == info['token_uri']
+        assert credentials._scopes == ['email', 'profile']
```
dhermes: so much hardcoding, why not assign `scopes = ['email', 'profile']` above and then create your creds with `scopes=scopes` and assert `credentials._scopes == scopes`. But like also don't hardcode `subject` and anything else I missed (addl. claims)

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+        assert credentials._subject == 'subject'
+        assert credentials._additional_claims['meta'] == 'data'
+
+    def test_from_service_account_bad_key(self):
+        info = dict(SERVICE_ACCOUNT_INFO)
```
dhermes: This call to `dict()` is in lieu of `.copy()`?

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+        info['private_key'] = 'garbage'
+
+        with pytest.raises(ValueError) as excinfo:
+            service_account.Credentials.from_service_account_info(info)
+
+        assert excinfo.match(r'No key could be detected')
```
dhermes: Good thing you used a raw string, that was a close one

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def test_from_service_account_bad_format(self):
+        info = {}
+
+        with pytest.raises(KeyError):
+            service_account.Credentials.from_service_account_info(info)
```
dhermes: `(info)` --> `({})`? brevity is the soul of spit

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+        assert not self.credentials.valid
+        # Expiration hasn't been set yet
+        assert not self.credentials.expired
+        # Scopes haven't been specified yet
+        assert self.credentials.requires_scopes
+
+    def test_sign_bytes(self):
+        to_sign = b'123'
+        signature = self.credentials.sign_bytes(to_sign)
+        crypt.verify_signature(to_sign, signature, PUBLIC_CERT_BYTES)
```
dhermes: This doesn't assert anything. Does `verify_signature` have a return value?

tests/oauth2/test_service_account.py
```diff
@@ -0,0 +1,199 @@
...
+        jwt_grant_mock.return_value = (
+            'token', _helpers.utcnow() + datetime.timedelta(seconds=500), None)
+        request_mock = mock.Mock()
+
+        # Refresh credentials
+        self.credentials.refresh(request_mock)
+
+        # Check jwt grant call.
+        assert jwt_grant_mock.called
+        request, token_uri, assertion = jwt_grant_mock.call_args[0]
```
dhermes: Why use `call_args` here? Try to avoid it at all costs. (I'm not even sure what the 0 index means from `call_args`.)

google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+            additional_claims (Mapping[str, str]): Any additional claims for
+                the JWT assertion used in the authorization grant.
+        """
+        super(Credentials, self).__init__()
+        self._scopes = scopes
+        self._signer = signer
+        self._service_account_email = service_account_email
+        self._subject = subject
+        self._token_uri = token_uri
+        self._additional_claims = additional_claims or {}
```
jonparrott: Fair enough, fixed.

google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+        return cls(
+            signer, service_account_email=email, token_uri=token_uri, **kwargs)
+
+    @classmethod
+    def from_service_account_file(cls, filename, **kwargs):
+        """Creates a Credentials instance from a service account json file.
+
+        Args:
+            filename (str): The path to the service account json file.
+            kwargs: Additional arguments to pass to the constructor.
```
jonparrott: Pylint doesn't require kwargs to have a type, mostly because there's no way for a user to pass a different type as kwargs.

google/oauth2/service_account.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Returns:
+            google.auth.service_account.Credentials: The constructed
+                credentials.
+        """
+        email = info['client_email']
+        key_id = info['private_key_id']
+        private_key = info['private_key']
+        token_uri = info['token_uri']
```
jonparrott: Done.

google/oauth2/service_account.py
jonparrott: None or empty container are both considered not having scopes.

google/oauth2/service_account.py
jonparrott: Yes, python-rsa is thread-safe.

google/oauth2/service_account.py
jonparrott: because we removed `scopes_to_strings` ability to handle anything other than a sequence in the last review. :P 

Will pass an empty tuple.

google/oauth2/service_account.py
jonparrott: Yes, fixed.

google/oauth2/service_account.py
jonparrott: additional data returned by the OAuth 2.0 token endpoint, it's unused here.

tests/oauth2/test_service_account.py
jonparrott: yes, done.

tests/oauth2/test_service_account.py
jonparrott: Done.

tests/oauth2/test_service_account.py
jonparrott: done

tests/oauth2/test_service_account.py
jonparrott: Yes, as mentioned in the other review - test methods never need to call the fixture directly, but if it makes you happier I can rename it.

tests/oauth2/test_service_account.py
jonparrott: > don't hardcode

Did you mean do? Because I did. Let me know if you meant otherwise.

tests/oauth2/test_service_account.py
jonparrott: Yes, switched to copy().

tests/oauth2/test_service_account.py
jonparrott: lol it's good to be consistent and always use raw strings for regexes.

tests/oauth2/test_service_account.py
jonparrott: Fixed.

tests/oauth2/test_service_account.py
jonparrott: Yes, fixed.

tests/oauth2/test_service_account.py
jonparrott: Because there's no way to use `jwt.decode` to verify the assertion with `assert_called_with`.

google/oauth2/service_account.py
jonparrott: Switched to `io.open` with a specific encoding. json.decode requires str, not bytes.

google/oauth2/service_account.py
dhermes: Thread-safety not my concern here. Is the `signer` mutable? If yes, then having two objects share it is problematic, since one owner could edit it and cause the other owner to be impacted by accident.

tests/oauth2/test_service_account.py
dhermes: Gotcha

dhermes:
LGTM after `def credentials` rename?

PR #45, opened 2016-10-21
Add service_account.Credentials.to_jwt_credentials

### LINKED ISSUE #35: Implement service_account.Credentials.to_jwt_credentials()
(Context: #29)

jonparrott:
Closes #35.

### DIFF
--- a/google/oauth2/service_account.py
+++ b/google/oauth2/service_account.py
@@ -204,6 +204,33 @@ class Credentials(credentials.Signing,
             filename, require=['client_email', 'token_uri'])
         return cls._from_signer_and_info(signer, info, **kwargs)
 
+    def to_jwt_credentials(self):
+        """Creates a :cls:`google.auth.jwt.Credentials` instance from this
+        instance.
+
+        The new instance will use the same private key as this instance and
+        will use this instance's service account email as the issuer and
+        subject.
+
+        This is the same as calling
+        :meth:`jwt.Credentials.from_service_account_file` with the same
+        file used to create these credentials::
+
+            svc_creds = service_account.Credentials.from_service_account_file(
+                'service_account.json')
+            jwt_from_svc = svc_credentials.to_jwt_credentials()
+            # is the same as:
+            jwt_creds = jwt.Credentials.from_service_account_file(
+                'service_account.json')
+
+        Returns:
+            google.auth.jwt.Credentials: A new Credentials instance.
+        """
+        return jwt.Credentials(
+            self._signer,
+            issuer=self._service_account_email,
+            subject=self._service_account_email)
+
     @property
     def requires_scopes(self):
         """Checks if the credentials requires scopes.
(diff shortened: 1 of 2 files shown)


google/oauth2/service_account.py
```diff
@@ -204,6 +204,33 @@ def from_service_account_file(cls, filename, **kwargs):
             filename, require=['client_email', 'token_uri'])
         return cls._from_signer_and_info(signer, info, **kwargs)
 
+    def to_jwt_credentials(self):
+        """Creates a :cls:`google.auth.jwt.Credentials` instance from this
```
dhermes: cls a synonym for class? Why don't I **ALWAYS** use `cls`!

google/oauth2/service_account.py
```diff
@@ -204,6 +204,33 @@ def from_service_account_file(cls, filename, **kwargs):
...
+            jwt_from_svc = svc_credentials.to_jwt_credentials()
+            # is the same as:
+            jwt_creds = jwt.Credentials.from_service_account_file(
+                'service_account.json')
+
+        Returns:
+            google.auth.jwt.Credentials: A new Credentials instance.
+        """
+        return jwt.Credentials(
+            self._signer,
```
dhermes: Is there a worry here of one creds instance mutating the signer on purpose, which well make the other creds instance unhappy? (i.e. shouldn't you copy the signer?)

google/oauth2/service_account.py
```diff
@@ -204,6 +204,33 @@ def from_service_account_file(cls, filename, **kwargs):
             filename, require=['client_email', 'token_uri'])
         return cls._from_signer_and_info(signer, info, **kwargs)
 
+    def to_jwt_credentials(self):
```
dhermes: Add an optional `audience`? Or the class is already good at handling that?

tests/oauth2/test_service_account.py
```diff
@@ -58,15 +58,14 @@ def credentials_fixture(self, signer):
             signer, self.SERVICE_ACCOUNT_EMAIL, self.TOKEN_URI)
 
     def test_from_service_account_info(self):
-        with open(SERVICE_ACCOUNT_JSON_FILE, 'r') as fh:
-            info = json.load(fh)
```
dhermes: Pete and re-pete were on a boat?

You seem to like `info = SERVICE_ACCOUNT_INFO.copy()` everywhere else (would've also been a smaller diff, "diff-driven development")

google/oauth2/service_account.py
jonparrott: Signer is immutable.

tests/oauth2/test_service_account.py
jonparrott: copy() is used when I modify the info to induce particular behavior.

google/oauth2/service_account.py
jonparrott: `with_claims` covers it.

PR #46, opened 2016-10-21
Add app engine credentials

### LINKED ISSUE #36: Implement App Engine credentials
(Towards #3)

jonparrott:
Closes #36

### DIFF
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -28,7 +28,6 @@ from google.auth import _helpers
 from google.auth import credentials
 from google.auth import exceptions
 
-
 try:
     from google.appengine.api import app_identity
 except ImportError:
@@ -39,7 +38,7 @@ class Credentials(credentials.Scoped, credentials.Signing,
                   credentials.Credentials):
     """App Engine standard environment credentials.
 
-    These credentials use the App Engine App Idenity API to obtain access
+    These credentials use the App Engine App Identity API to obtain access
     tokens.
     """
 
@@ -50,7 +49,7 @@ class Credentials(credentials.Scoped, credentials.Signing,
                 API.
             service_account_id (str): The service account ID passed into
                 :func:`google.appengine.api.app_identity.get_access_token`.
-                This is not required as the default application service account
+                If not specified, the default application service account
                 ID will be used.
         """
         super(Credentials, self).__init__()
@@ -68,7 +67,7 @@ class Credentials(credentials.Scoped, credentials.Signing,
                 not available.
         """
         # pylint: disable=unused-argument
-        if not app_identity:
+        if app_identity is None:
             raise exceptions.RefreshError(
                 'The App Engine APIs are not available.')
 
@@ -85,7 +84,7 @@ class Credentials(credentials.Scoped, credentials.Signing,
         Returns:
             bool: True if there are no scopes set otherwise False.
         """
-        return True if not self._scopes else False
+        return not self._scopes
 
     @_helpers.copy_docstring(credentials.Scoped)
     def with_scopes(self, scopes):
@@ -104,7 +103,7 @@ class Credentials(credentials.Scoped, credentials.Signing,
         Raises:
             EnvironmentError: If the App Engine APIs are unavailable.
         """
-        if not app_identity:
+        if app_identity is None:
             raise EnvironmentError('The App Engine APIs are not available.')
 
         return app_identity.sign_blob(message)
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -26,7 +26,6 @@ import datetime
 
 from google.auth import _helpers
 from google.auth import credentials
-from google.auth import exceptions
 
 try:
     from google.appengine.api import app_identity
@@ -51,26 +50,21 @@ class Credentials(credentials.Scoped, credentials.Signing,
                 :func:`google.appengine.api.app_identity.get_access_token`.
                 If not specified, the default application service account
                 ID will be used.
+
+        Raises:
+            EnvironmentError: If the App Engine APIs are unavailable.
         """
+        if app_identity is None:
+            raise EnvironmentError(
+                'The App Engine APIs are not available.')
+
         super(Credentials, self).__init__()
         self._scopes = scopes
         self._service_account_id = service_account_id
 
+    @_helpers.copy_docstring(credentials.Credentials)
     def refresh(self, request):
-        """Refresh the access token and scopes.
-
-        Args:
-            request (google.auth.transport.Request): Unused.
-
-        Raises:
-            google.auth.exceptions.RefreshError: If the App Engine APIs are
-                not available.
-        """
         # pylint: disable=unused-argument
-        if app_identity is None:
-            raise exceptions.RefreshError(
-                'The App Engine APIs are not available.')
-
         token, ttl = app_identity.get_access_token(
             self._scopes, self._service_account_id)
         expiry = _helpers.utcnow() + datetime.timedelta(seconds=ttl)
@@ -91,19 +85,6 @@ class Credentials(credentials.Scoped, credentials.Signing,
         return Credentials(
             scopes=scopes, service_account_id=self._service_account_id)
 
+    @_helpers.copy_docstring(credentials.Signing)
     def sign_bytes(self, message):
-        """Signs the given message.
-
-        Args:
-            message (bytes): The message to sign.
-
-        Returns:
-            bytes: The message's cryptographic signature.
-
-        Raises:
-            EnvironmentError: If the App Engine APIs are unavailable.
-        """
-        if app_identity is None:
-            raise EnvironmentError('The App Engine APIs are not available.')
-
         return app_identity.sign_blob(message)
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -1,0 +1,110 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Google App Engine standard environment credentials.
+
+This module provides authentication for application running on App Engine in
+the standard environment using the `App Identity API`_.
+
+
+.. _App Identity API:
+    https://cloud.google.com/appengine/docs/python/appidentity/
+"""
+
+import datetime
+
+from google.auth import _helpers
+from google.auth import credentials
+from google.auth import exceptions
+
+
+try:
+    from google.appengine.api import app_identity
+except ImportError:
+    app_identity = None
+
+
+class Credentials(credentials.Scoped, credentials.Signing,
+                  credentials.Credentials):
+    """App Engine standard environment credentials.
+
+    These credentials use the App Engine App Idenity API to obtain access
+    tokens.
+    """
+
+    def __init__(self, scopes=None, service_account_id=None):
+        """
+        Args:
+            scopes (Sequence[str]): Scopes to request from the App Identity
+                API.
+            service_account_id (str): The service account ID passed into
+                :func:`google.appengine.api.app_identity.get_access_token`.
+                This is not required as the default application service account
+                ID will be used.
+        """
+        super(Credentials, self).__init__()
+        self._scopes = scopes
+        self._service_account_id = service_account_id
+
+    def refresh(self, request):
+        """Refresh the access token and scopes.
+
+        Args:
+            request (google.auth.transport.Request): Unused.
+
+        Raises:
+            google.auth.exceptions.RefreshError: If the App Engine APIs are
+                not available.
+        """
+        # pylint: disable=unused-argument
+        if not app_identity:
+            raise exceptions.RefreshError(
+                'The App Engine APIs are not available.')
+
+        token, ttl = app_identity.get_access_token(
+            self._scopes, self._service_account_id)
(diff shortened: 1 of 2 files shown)


google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+    https://cloud.google.com/appengine/docs/python/appidentity/
+"""
+
+import datetime
+
+from google.auth import _helpers
+from google.auth import credentials
+from google.auth import exceptions
+
+
```
nathanielmanistaatgoogle: No need for two blank lines here and it kind of breaks the imports up too much?

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+        Args:
+            request (google.auth.transport.Request): Unused.
+
+        Raises:
+            google.auth.exceptions.RefreshError: If the App Engine APIs are
+                not available.
+        """
+        # pylint: disable=unused-argument
+        if not app_identity:
```
nathanielmanistaatgoogle: Make an explicit `is None` comparison?

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+        self.token, self.expiry = token, expiry
+
+    @property
+    def requires_scopes(self):
+        """Checks if the credentials requires scopes.
+
+        Returns:
+            bool: True if there are no scopes set otherwise False.
+        """
+        return True if not self._scopes else False
```
nathanielmanistaatgoogle: This looks like a wordy way of saying `return not self._scopes`.

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+        Args:
+            message (bytes): The message to sign.
+
+        Returns:
+            bytes: The message's cryptographic signature.
+
+        Raises:
+            EnvironmentError: If the App Engine APIs are unavailable.
+        """
+        if not app_identity:
```
nathanielmanistaatgoogle: Again, favor explicit comparison to `None` since that better reflects your authorial intent.

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+    from google.appengine.api import app_identity
+except ImportError:
+    app_identity = None
+
+
+class Credentials(credentials.Scoped, credentials.Signing,
+                  credentials.Credentials):
+    """App Engine standard environment credentials.
+
+    These credentials use the App Engine App Idenity API to obtain access
```
dhermes: s/Idenity/Identity

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+
+from google.auth import _helpers
+from google.auth import credentials
+from google.auth import exceptions
+
+
+try:
+    from google.appengine.api import app_identity
+except ImportError:
+    app_identity = None
```
dhermes: Don't you want an `ImportError` here?

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def __init__(self, scopes=None, service_account_id=None):
+        """
+        Args:
+            scopes (Sequence[str]): Scopes to request from the App Identity
+                API.
+            service_account_id (str): The service account ID passed into
+                :func:`google.appengine.api.app_identity.get_access_token`.
+                This is not required as the default application service account
+                ID will be used.
```
dhermes: This is unclear. Is it "not required" or "completely unused"?

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+        """Refresh the access token and scopes.
+
+        Args:
+            request (google.auth.transport.Request): Unused.
+
+        Raises:
+            google.auth.exceptions.RefreshError: If the App Engine APIs are
+                not available.
+        """
+        # pylint: disable=unused-argument
```
dhermes: Are you opposed to having `dummy-variables-rgx=unused_$|_$|dummy` in your `pylintrc` (then you could call it `unused_request` and not have the `disable`? Or do you value having the positional arg also honor the name (in cases where callers use it as a keyword arg)?

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+
+        token, ttl = app_identity.get_access_token(
+            self._scopes, self._service_account_id)
+        expiry = _helpers.utcnow() + datetime.timedelta(seconds=ttl)
+
+        self.token, self.expiry = token, expiry
+
+    @property
+    def requires_scopes(self):
+        """Checks if the credentials requires scopes.
```
dhermes: This isn't what this method seems to do. It checks if the credentials **HAS** scopes (i.e. I don't undersatnd why you're saying that makes them considered to be "required")

google/auth/app_engine.py
```diff
@@ -0,0 +1,110 @@
...
+
+        Returns:
+            bool: True if there are no scopes set otherwise False.
+        """
+        return True if not self._scopes else False
+
+    @_helpers.copy_docstring(credentials.Scoped)
+    def with_scopes(self, scopes):
+        return Credentials(
+            scopes=scopes, service_account_id=self._service_account_id)
```
dhermes: This overwrites scopes instead of appending. Worth documenting or is it in the parent?

tests/test_app_engine.py
```diff
@@ -0,0 +1,118 @@
...
+@pytest.fixture
+def app_identity_mock(monkeypatch):
+    """Mocks the google.appengine.api.app_identity module."""
+    api_mock = mock.Mock()
+    app_identity_mock = api_mock.app_identity
+
+    monkeypatch.setitem(
+        sys.modules, 'google.appengine', mock.Mock())
+    monkeypatch.setitem(
+        sys.modules, 'google.appengine.api', api_mock)
```
dhermes: Ha, bad! Don't you need the mock objects to reference each other? Like

```python
import google.appengine as m
print(m.api)
```

tests/test_app_engine.py
```diff
@@ -0,0 +1,118 @@
...
+    from google.appengine.api import app_identity
+    assert app_identity == app_identity_mock
+
+    yield app_identity_mock
+
+
+@pytest.fixture
+def app_engine(app_identity_mock):
+    from google.auth import app_engine
+    reload_module(app_engine)
```
dhermes: Why the reload?

tests/test_app_engine.py
```diff
@@ -0,0 +1,118 @@
...
+    yield app_engine
```
dhermes: Why the yield?

tests/test_app_engine.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@pytest.fixture
+def app_engine_no_apis():
+    from google.auth import app_engine
+    reload_module(app_engine)
```
dhermes: Wouldn't it be easier to just mock `app_engine.app_identity`?

tests/test_app_engine.py
```diff
@@ -0,0 +1,118 @@
...
+    yield app_engine
+
+
+class TestCredentials(object):
+    def test_default_state(self, app_engine):
```
dhermes: Default meaning the import failed or just default meaning the constructor defaults?

tests/test_app_engine.py
```diff
@@ -0,0 +1,118 @@
...
+        assert scoped_credentials.has_scopes(['email'])
+        assert not scoped_credentials.requires_scopes
+
+    @mock.patch(
+        'google.auth._helpers.utcnow',
+        return_value=datetime.datetime.min)
+    def test_refresh(self, now_mock, app_engine, app_identity_mock):
+        token = 'token'
+        ttl = 100
+        app_identity_mock.get_access_token.return_value = (token, ttl)
```
dhermes: Fun fact: RHS parens are optional (in addition to LHS being "optional" when unpacking)

```python
>>> val = 55, 'shfifty-five'
>>> val
(55, 'shfifty-five')
```

google/auth/app_engine.py
jonparrott: FIxed.

google/auth/app_engine.py
jonparrott: Nah, it does so on `refresh` and `sign_blob`. Do you feel like we should raise here?

google/auth/app_engine.py
jonparrott: Fixed, thanks.

google/auth/app_engine.py
jonparrott: Clarified.

google/auth/app_engine.py
jonparrott: > Or do you value having the positional arg also honor the name (in cases where callers use it as a keyword arg)?

This.

google/auth/app_engine.py
jonparrott: Done.

google/auth/app_engine.py
jonparrott: So, this is to satisfy what used to be covered by `create_scoped_required`, see [here](https://github.com/google/google-api-python-client/blob/master/googleapiclient/discovery.py#L339).

Essentially, some credentials *require* scopes in order to be refreshed. Others can have scopes but don't need to have them to be refreshed. Basically, consumers of this library would do:

```python
if isinstance(credentials, google.auth.credentials.Scoped) and credentials.requires_scopes:
   credentials = credentials.with_scopes(scopes_for_this_api)
```

google/auth/app_engine.py
jonparrott: I've used this elsewhere, but sure, I'll switch it.

google/auth/app_engine.py
jonparrott: The superclass documents this as "Create a copy of these credentials with the specified scopes."

If you feel like it's not clear enough, care to suggest alternate wording?

google/auth/app_engine.py
jonparrott: Done.

tests/test_app_engine.py
jonparrott: Oh my god. Yes. I feel so dumb.

b0wp9

tests/test_app_engine.py
jonparrott: No, but no longer needed because I just totally changed all this.

tests/test_app_engine.py
jonparrott: It was to make it re-import `google.appengine.api`, but it's been removed.

tests/test_app_engine.py
jonparrott: It's how pytest fixtures work. Everything before the yield happens before the test, everything after the yield is run after the text completes.

tests/test_app_engine.py
jonparrott: constructor defaults.

tests/test_app_engine.py
jonparrott: Neat!

google/auth/app_engine.py
dhermes: Kind of. Then the ADC code could catch the `ImportError` but any direct import would fail immediately. I don't really care either way (i.e. "kind of")

google/auth/app_engine.py
dhermes: No, I understand why the `requires_scopes` feature exists, I don't understand why for GAE creds the value would correspond to this implementation

google/auth/app_engine.py
dhermes: Yeah it's clear enough

google/auth/app_engine.py
```diff
@@ -0,0 +1,109 @@
...
+                :func:`google.appengine.api.app_identity.get_access_token`.
+                If not specified, the default application service account
+                ID will be used.
+        """
+        super(Credentials, self).__init__()
+        self._scopes = scopes
+        self._service_account_id = service_account_id
+
+    def refresh(self, request):
+        """Refresh the access token and scopes.
```
nathanielmanistaatgoogle: Does this function need a doc string (as opposed to `@_helpers.copy_docstring`?)? Does this method behave in any worth specifying that's not already specified by the overridden method?

google/auth/app_engine.py
```diff
@@ -0,0 +1,109 @@
...
+
+        Args:
+            request (google.auth.transport.Request): Unused.
```
nathanielmanistaatgoogle: This seems like an odd thing to document (and thus promise going forward until the next API break-point).

google/auth/app_engine.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        return not self._scopes
+
+    @_helpers.copy_docstring(credentials.Scoped)
+    def with_scopes(self, scopes):
+        return Credentials(
+            scopes=scopes, service_account_id=self._service_account_id)
+
+    def sign_bytes(self, message):
+        """Signs the given message.
```
nathanielmanistaatgoogle: Why this doc string rather than `_helpers.copy_docstring`?

google/auth/app_engine.py
jonparrott: @dhermes same as Service Account credentials: GAE credentials require scopes to be able to obtain an access token. Since credentials can be created without scopes (ADC), this will initial be `True` indicating that callers should call `with_scopes`. Once scopes are set, it's no longer necessary to call `with_scopes` in order to obtain an access token.

google/auth/app_engine.py
jonparrott: As discussed offline, I'm going to make __init__() raise, so that ADC can easily ascertain if the credentials are valid and so that tests don't need to be needlessly complicated.

google/auth/app_engine.py
jonparrott: Not any more, removed.

google/auth/app_engine.py
jonparrott: Previously because it raised environmenterrror, but that's been fixed and now it just uses the parent class' docstring.

google/auth/app_engine.py
dhermes: I get it now. The previous code was throwing me off but `not self._scopes` is clear to me now. (I was reading it before as `return bool(self._scopes)`. Less is more, eh?

jonparrott:
@dhermes @nathanielmanistaatgoogle PTAL, should be good to go.

dhermes:
LGTM

PR #48, opened 2016-10-24
Add App Engine credentials to google.auth.default

### LINKED ISSUE #47: Add App Engine credentials to Application Default Credentials

jonparrott:
Resolves #47.

Additionally moves GCE's project id detection into `_metadata` to keep things consistent.

### DIFF
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -33,6 +33,21 @@ except ImportError:
     app_identity = None
 
 
+def get_project_id():
+    """Gets the project ID for the current App Engine application.
+
+    Returns:
+        str: The project ID
+
+    Raises:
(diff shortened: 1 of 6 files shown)


google/auth/app_engine.py
```diff
@@ -33,6 +33,21 @@
     app_identity = None
 
 
+def get_project_id():
+    """Gets the project ID for the current App Engine application.
+
+    Returns:
+        str: The project ID
```
dhermes: Can it ever return None? Like if in "flex" and the metadata server is being ornery?

jonparrott: Something will have to be seriously FUBARed for that to happen.

PR #51, opened 2016-10-25
Add system tests for service account credentials

### LINKED ISSUE #42: Create system tests for service account-based credentials

jonparrott:
Resolves #42


jonparrott:
Hrm. No idea why the certificate verification is failing. I'm gonna poke around and see what's up.

system_tests/conftest.py
```diff
@@ -0,0 +1,73 @@
...
+        response = request(url=url, method='GET')
+
+        return json.loads(response.data.decode('utf-8'))
+
+    yield _token_info
+
+
+def verify_environment():
+    """Checks to make sure that requisite data files are available."""
+    if not os.path.exists(DATA_DIR):
```
dhermes: `isdir`?

system_tests/test_service_account.py
```diff
@@ -0,0 +1,43 @@
...
+    credentials.refresh(request)
+
+    assert credentials.token
+
+    info = token_info(credentials.token)
+
+    assert info['email'] == credentials._service_account_email
+    assert info['scope'] == (
+        'https://www.googleapis.com/auth/userinfo.email '
+        'https://www.googleapis.com/auth/userinfo.profile')
```
dhermes: You don't want any constants here?

jonparrott: Not really.

jonparrott:
Holy old OpenSSL version, batman!

```
OpenSSL 1.0.1 14 Mar 2012
```

jonparrott:
@dhermes I'm not sure what to do here other than disable HTTPS for the system tests. :(

system_tests/conftest.py
jonparrott: Done

dhermes:
@jonparrott We've never had issues with HTTPS on Travis.

jonparrott:
@dhermes using httplib2? *mild shock*

httplib2 doesn't do SNI verification at all. Urllib3 is choking because google uses SNI extensively and since the openssl version is too low to include SNI support, it prefers to raise an SSL error instead of just continuing the request.

jonparrott:
Some research:

With our Python runtime (debian8) using both `certifi.where()` and OS certs work with SNI.

With Travis (Ubuntu 12.04 LTS) using `certifi.old_where()` works but generates an `InsecurePlatformWarning`. Using OS certs works but generates an `SNIMissingWarning`. Using OS certs with PyOpenSSL works without warnings.

dhermes:
Ha, should've guessed it.

dhermes:
LGTM (in light of #52)

PR #57, opened 2016-10-26
Read cloud sdk active_config file to determine the right config file to read

### LINKED ISSUE #53: Cloud SDK credentials should use the 'active' config not the 'default' config
Context: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2588

jonparrott:
Fixes #53

### DIFF
--- a/google/auth/_cloud_sdk.py
+++ b/google/auth/_cloud_sdk.py
@@ -14,6 +14,7 @@
 
 """Helpers for reading the Google Cloud SDK's configuration."""
 
+import io
 import os
 
 import six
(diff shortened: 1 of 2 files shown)


google/auth/_cloud_sdk.py
```diff
@@ -83,14 +80,49 @@ def get_application_default_credentials_path():
...
+    Returns:
+        str: The active configuration name.
+    """
+    active_config_filename = os.path.join(config_path, 'active_config')
+
+    if not os.path.isfile(active_config_filename):
+        return 'default'
+
+    with io.open(active_config_filename, 'r', encoding='utf-8') as file_obj:
+        active_config_name = file_obj.read().strip()
```
dhermes: Is this preferred to just using `subprocess` to ask the `gcloud` CLI? (Does it have a "porcelain" API like `git`?)

tests/test__cloud_sdk.py
```diff
@@ -41,15 +42,20 @@
...
     with CONFIG_PATH_PATCH as mock_get_config_dir:
         mock_get_config_dir.return_value = str(config_dir)
-        yield config_file
+        yield config_dir
+
+
+@pytest.fixture
+def config_file(config_dir):
+    config_file = py.path.local(_cloud_sdk._get_config_file(
+        str(config_dir), 'default'))
```
dhermes: Why `str(config_dir)`?

google/auth/_cloud_sdk.py
jonparrott: No porcelain, and the CLI changes it's face too often to be dependable. However, since it has to maintain backwards-compatibility with older versions, the config files are relatively stable. Finally, if any of this code fails the worst-case scenario is that we can't detect the project ID.

PR #81, opened 2016-11-10
Swap target, request for request, target in secure_authorized_channel's argument order.

jonparrott:


jonparrott:
@dhermes because the previous argument order just didn't flow logically to me, it was previously `auth object, grpc object, auth object, grpc object`, now it's `auth object, auth object, grpc object, grpc object`.

PR #87, opened 2016-11-16
Fix ID token verification

jonparrott:
* Fix call to `transport.Request`.
* Make jwt support base64 strings lacking padding characters.

### DIFF
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -14,6 +14,7 @@
 
 """Helper functions for commonly used utilities."""
 
+import base64
 import calendar
 import datetime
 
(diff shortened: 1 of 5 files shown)


google/auth/_helpers.py
```diff
@@ -194,3 +195,19 @@ def string_to_scopes(scopes):
...
+    Google infrastructure tends to omit the base64 padding characters.
+
+    Args:
+        value (Union[str, bytes]): The encoded value.
+
+    Returns:
+        bytes: The decoded value
+    """
+    b64string = to_bytes(value)
+    padded = b64string + b'=' * (4 - len(b64string) % 4)
```
dhermes: You want `(-len(b64string)) % 4`

```python
>>> for i in range(12):
...     print((i, (-i) % 4))
... 
(0, 0)
(1, 3)
(2, 2)
(3, 1)
(4, 0)
(5, 3)
(6, 2)
(7, 1)
(8, 0)
(9, 3)
(10, 2)
(11, 1)
```

jonparrott:
@dhermes when replacing some hand-rolled code that verified firebase auth tokens for an app engine sample.

google/auth/_helpers.py
jonparrott: This code was taking from (working) code in oauth2client, but met me change it and see what happens. I'll also add some test cases just to be sure.

dhermes: I get it, but it was unnecessarily adding padding in the 0 mod 4 case.

jonparrott: Done.

PR #89, opened 2016-12-02
Add Signing.singer_email

### LINKED ISSUE #85: Consider making `service_account_email` part of the `Signing` interface
google-cloud-python needs both a signer and a service account email to identify the signer to make a [signed url](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/115748814bf6a59d99c13e070dfdcfbdec5e865d/core/google/cloud/credentials.py#L119).

/cc @dhermes

jonparrott:
Resolves #85

### DIFF
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -110,3 +110,8 @@ class Credentials(credentials.Scoped, credentials.Signing,
     @_helpers.copy_docstring(credentials.Signing)
     def sign_bytes(self, message):
         return app_identity.sign_blob(message)
+
+    @property
+    @_helpers.copy_docstring(credentials.Signing)
+    def signer_email(self):
+        return self.service_account_email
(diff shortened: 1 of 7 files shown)


google/auth/app_engine.py
```diff
@@ -110,3 +110,8 @@ def with_scopes(self, scopes):
     @_helpers.copy_docstring(credentials.Signing)
     def sign_bytes(self, message):
         return app_identity.sign_blob(message)
+
+    @property
+    @_helpers.copy_docstring(credentials.Signing)
+    def signer_email(self):
+        return self.service_account_email
```
dhermes: Pylint won't import the code and check `my_obj.__doc__`, does static analysis pick up the fact that `__doc__` gets set by your decorator?

jonparrott: It appears so, because pylint doesn't complain about a missing docstring.

PR #95, opened 2016-12-13
Add crypt.Signer.from_service_account_file

### LINKED ISSUE #94: Make creating a Signer from a JSON keyfile easier.

jonparrott:
Resolves #94

### DIFF
--- a/google/auth/crypt.py
+++ b/google/auth/crypt.py
@@ -227,3 +227,18 @@ class Signer(object):
             raise ValueError('No key could be detected.')
 
         return cls(private_key, key_id=key_id)
+
+    @classmethod
+    def from_service_account_file(cls, filename):
+        """Creates a Signer instance from a service account .json file
+        in Google format.
+
+        Args:
+            filename (str): The path to the service account .json file.
+
+        Returns:
+            Signer: The constructed signer.
+        """
+        from google.auth import _service_account_info
+        _, signer = _service_account_info.from_filename(filename)
+        return signer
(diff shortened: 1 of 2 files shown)


google/auth/crypt.py
```diff
@@ -227,3 +227,18 @@ def from_string(cls, key, key_id=None):
...
+        """Creates a Signer instance from a service account .json file
+        in Google format.
+
+        Args:
+            filename (str): The path to the service account .json file.
+
+        Returns:
+            Signer: The constructed signer.
+        """
+        from google.auth import _service_account_info
```
dhermes: :-1: on the non-global import. Help me understand?

jonparrott: Without it, there's a circular import.

dhermes: Why?

jonparrott: because `_service_account_info` imports `crypt` to be able to create a `Signer` instance when loading the file.

dhermes: Can you move some of that behavior from `_service_account_info` into `crypt`?

jonparrott: I'd rather not.

dhermes: Why not?

jonparrott: Because right now `_service_account_info` does a really good job of containing all of the logic needed to parse and validate service account files. I'd need to either move or duplicate some of that logic into crypt (breaking single-responsibility). [This logic in particular](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/_service_account_info.py#L45) which constitues the majority of the logic in that module.

dhermes: OK SGTM. LGTM to merge

PR #97, opened 2016-12-13
Add app_engine.Signer

### LINKED ISSUE #49: Add GAE Signer

jonparrott:
Resolves #49

### DIFF
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -42,7 +42,8 @@ class Signer(object):
     def __init__(self):
         self.key_id = None
 
-    def sign(self, message):
+    @staticmethod
+    def sign(message):
         """Signs a message.
 
         Args:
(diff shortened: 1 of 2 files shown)


google/auth/app_engine.py
```diff
@@ -33,6 +33,31 @@
...
+    def sign(self, message):
+        """Signs a message.
+
+        Args:
+            message (Union[str, bytes]): The message to be signed.
+
+        Returns:
+            bytes: The signature of the message.
+        """
+        # pylint: disable=no-self-use
```
dhermes: :-1: on `no-self-use`, just make it a `@staticmethod`

google/auth/app_engine.py
```diff
@@ -33,6 +33,31 @@
...
+        # Locally disabled because this needs to confirm the to Signer
```
dhermes: "confirm the to" --> "conform to the"?

google/auth/app_engine.py
jonparrott: I'm on board with that, but doesn't it break the interface?

dhermes: I very much doubt it, but I'm open to learning.

google/auth/app_engine.py
jonparrott: Removed.

google/auth/app_engine.py
jonparrott: Switched to staticmethod

PR #99, opened 2016-12-14
Fix cyclic import and add Signer.from_service_account_info

jonparrott:


google/auth/crypt.py
```diff
@@ -229,6 +231,29 @@ def from_string(cls, key, key_id=None):
...
+            info (Mapping[str, str]): The service account info in Google
+                format.
+
+        Returns:
+            Signer: The constructed signer.
+
+        Raises:
+            ValueError: If the info is not in the expected format.
+        """
+        if 'private_key' not in info:
```
dhermes: Make `'private_key'` a module-level non-public constant? If for no other reason than being about to refer to it in tests.

google/auth/crypt.py
```diff
@@ -239,6 +264,7 @@ def from_service_account_file(cls, filename):
         Returns:
             Signer: The constructed signer.
         """
-        from google.auth import _service_account_info
-        _, signer = _service_account_info.from_filename(filename)
-        return signer
+        with io.open(filename, 'r', encoding='utf-8') as json_file:
```
dhermes: Why do we care about UTF-8? In Python 3 `json.load` eats bytes or unicode?

tox.ini
```diff
@@ -70,7 +70,7 @@ deps =
 commands = make -C docs html
 
 [testenv:lint]
-basepython = python3.5
+basepython = python3.4
```
dhermes: Unrelated change! (It's fine to leave in.)

google/auth/crypt.py
jonparrott: > Why do we care about UTF-8?

Because json files should be in utf-8 as per the spec, and we don't want the system default encoding.
 
> In Python 3 json.load eats bytes or unicode?

I don't know about that, I've always passed in a file in text mode.

PR #105, opened 2017-02-08
Add warning about oauth2.flow

jonparrott:


jonparrott:
In all likelyhood, I'll remove this warning before 1.0, but it's there as a CYA.

dhermes:
LGTM

PR #108, opened 2017-02-15
Add IAM signer

jonparrott:
/cc @matthewg

### DIFF
--- a/google/auth/iam.py
+++ b/google/auth/iam.py
@@ -22,8 +22,14 @@ API`_'s auth-related functionality.
 import base64
 import json
 
+from six.moves import http_client
+
 from google.auth import _helpers
 from google.auth import exceptions
+
+_IAM_API_ROOT_URI = 'https://iam.googleapis.com/v1'
+_SIGN_BLOB_URI = (
+    _IAM_API_ROOT_URI + '/projects/-/serviceAccounts/{}:signBlob?alt=json')
 
 
 class Signer(object):
@@ -57,9 +63,10 @@ class Signer(object):
 
                 - https://www.googleapis.com/auth/iam
                 - https://www.googleapis.com/auth/cloud-platform
-            service_account_email (str): The service account email to  use to
-                sign bytes. Often, this can be the same as the service account
-                email in the given credentials.
+            service_account_email (str): The service account email identifying
+                which service account to use to sign bytes. Often, this can
+                be the same as the service account email in the given
+                credentials.
         """
         self._request = request
         self._credentials = credentials
@@ -70,25 +77,22 @@ class Signer(object):
         message = _helpers.to_bytes(message)
 
         method = 'POST'
-        url = (
-            'https://iam.googleapis.com/v1/projects/-/serviceAccounts/{}'
-            ':signBlob?alt=json').format(self._service_account_email)
+        url = _SIGN_BLOB_URI.format(self._service_account_email)
         headers = {}
         body = json.dumps({
-            'bytesToSign': base64.b64encode(message).decode('utf-8')
+            'bytesToSign': base64.b64encode(message).decode('utf-8'),
         })
 
         self._credentials.before_request(self._request, method, url, headers)
         response = self._request(
             url=url, method=method, body=body, headers=headers)
 
-        if response.status != 200:
+        if response.status != http_client.OK:
             raise exceptions.TransportError(
                 'Error calling the IAM signBytes API: {}'.format(
                     response.data))
 
-        data = json.loads(response.data.decode('utf-8'))
-        return data
+        return json.loads(response.data.decode('utf-8'))
 
     @property
     def key_id(self):
--- a/google/auth/iam.py
+++ b/google/auth/iam.py
@@ -1,0 +1,113 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tools for using the Google `Cloud Identity and Access Management (IAM)
+API`_'s auth-related functionality.
+
+.. _Cloud Identity and Access Management (IAM) API:
+    https://cloud.google.com/iam/docs/
+"""
+
+import base64
+import json
+
+from google.auth import _helpers
+from google.auth import exceptions
+
+
+class Signer(object):
+    """Signs messages using the IAM `signBlob API`_.
+
+    This is useful when you need to sign bytes but do not have access to the
+    credential's private key file.
+
+    .. warning::
+        The IAM API signs bytes using Google-managed keys. Because of this
+        it's possible that the key used to sign bytes will change. In some
+        cases this change can occur between successive calls to :attr:`key_id`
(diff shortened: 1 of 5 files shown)


jonparrott:
FYI to @elibixby as well.

google/auth/iam.py
```diff
@@ -0,0 +1,113 @@
...
+    """Signs messages using the IAM `signBlob API`_.
+
+    This is useful when you need to sign bytes but do not have access to the
+    credential's private key file.
+
+    .. warning::
+        The IAM API signs bytes using Google-managed keys. Because of this
+        it's possible that the key used to sign bytes will change. In some
+        cases this change can occur between successive calls to :attr:`key_id`
+        and :meth:`sign`. This could result in a signature that was signed
```
dhermes: Does `sign` also mention this?

google/auth/iam.py
```diff
@@ -0,0 +1,113 @@
...
+        Args:
+            request (google.auth.transport.Request): The object used to make
+                HTTP requests.
+            credentials (google.auth.credentials.Credentials): The credentials
+                that will be used to authenticate the request to the IAM API.
+                The credentials must have of one the following scopes:
+
+                - https://www.googleapis.com/auth/iam
+                - https://www.googleapis.com/auth/cloud-platform
+            service_account_email (str): The service account email to  use to
```
dhermes: "to  use" --> "to use"

google/auth/iam.py
```diff
@@ -0,0 +1,113 @@
...
+                sign bytes. Often, this can be the same as the service account
+                email in the given credentials.
```
dhermes: Doesn't the "Often" bit map to making this an optional arg?

google/auth/iam.py
```diff
@@ -0,0 +1,113 @@
...
+        self._service_account_email = service_account_email
+
+    def _make_signing_request(self, message):
+        """Makes a request to the API signBlob API."""
+        message = _helpers.to_bytes(message)
+
+        method = 'POST'
+        url = (
+            'https://iam.googleapis.com/v1/projects/-/serviceAccounts/{}'
+            ':signBlob?alt=json').format(self._service_account_email)
```
dhermes: Can you make this template a non-public constant? (Either in module or on class)

google/auth/iam.py
```diff
@@ -0,0 +1,113 @@
...
+        headers = {}
+        body = json.dumps({
+            'bytesToSign': base64.b64encode(message).decode('utf-8')
```
dhermes: Trailing comma?

google/auth/iam.py
```diff
@@ -0,0 +1,113 @@
...
+        })
+
+        self._credentials.before_request(self._request, method, url, headers)
+        response = self._request(
+            url=url, method=method, body=body, headers=headers)
+
+        if response.status != 200:
```
dhermes: Don't hardcode 200?

google/auth/iam.py
```diff
@@ -0,0 +1,113 @@
...
+            raise exceptions.TransportError(
+                'Error calling the IAM signBytes API: {}'.format(
+                    response.data))
+
+        data = json.loads(response.data.decode('utf-8'))
+        return data
```
dhermes: Could just `return json.loads(...)`?

google/auth/iam.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @property
+    def key_id(self):
+        """Optional[str]: The key ID used to identify this private key.
+
+        .. note::
+            This makes an API request to the IAM API.
+        """
+        response = self._make_signing_request('')
```
dhermes: This is a crappy hack.

1. Is there no API endpoint to get this ID?
1. Is it not worth caching this value?

tests/test_iam.py
```diff
@@ -0,0 +1,98 @@
...
+    response.status = status
+
+    if data is not None:
+        response.data = json.dumps(data).encode('utf-8')
+
+    return mock.Mock(return_value=response, spec=transport.Request)
+
+
+def make_credentials():
+    class CredentialsImpl(google.auth.credentials.Credentials):
```
dhermes: Why not define this at module level?

tests/test_iam.py
```diff
@@ -0,0 +1,98 @@
...
+            # Force refresh
+            self.expiry = datetime.datetime.min
+
+        def refresh(self, request):
+            pass
+
+    return CredentialsImpl()
+
+
+class TestSigner(object):
```
dhermes: Don't you dislike these `TestFoo` classes? None of your methods use `self` so it is unneeded?

tests/test_iam.py
```diff
@@ -0,0 +1,98 @@
...
+
+        signer = iam.Signer(
+            request, credentials, mock.sentinel.service_account_email)
+
+        returned_signature = signer.sign('123')
+
+        assert returned_signature == signature
+
+    def test_sign_bytes_failure(self):
+        request = make_request(401)
```
dhermes: Ditto on hardcoding 401 here (and the two 200's above)

google/auth/iam.py
jonparrott: It doesn't, but I'm fine with this warning just being in the class docstring.

google/auth/iam.py
jonparrott: Fixed.

google/auth/iam.py
jonparrott: No, the Credentials can be non-service-account credentials.

google/auth/iam.py
jonparrott: Done.

google/auth/iam.py
jonparrott: Done.

google/auth/iam.py
jonparrott: Done.

google/auth/iam.py
jonparrott: Done.

google/auth/iam.py
jonparrott: 1) No, and I'm pretty pissed about it.
2) No, because Google can rotate the key and our cached ID could be out of date.

tests/test_iam.py
jonparrott: No reason, just following what I've done elsewhere and hiding these "mock" classes.

tests/test_iam.py
jonparrott: I do, but I prefer it in this case because of the following two conditions:

1) It tests a class.
2) The iam module may eventually contain other classes or top-level functions.

tests/test_iam.py
jonparrott: Done.

google/auth/iam.py
dhermes: 1. Sorry
1. See my comment on #109 about invalidating the cache.

google/auth/iam.py
```diff
@@ -0,0 +1,117 @@
...
+        headers = {}
+        body = json.dumps({
+            'bytesToSign': base64.b64encode(message).decode('utf-8'),
+        })
+
+        self._credentials.before_request(self._request, method, url, headers)
+        response = self._request(
+            url=url, method=method, body=body, headers=headers)
+
+        if response.status != http_client.OK:
```
lukesneeringer: Are there ever cases where this particular API would send back successes that are not `200`? (Say, `201` or `202`?)

jonparrott:
Nope, just 200.

lukesneeringer:
Is there any risk of that changing? Cards on table: I usually do `status < 400` for this kind of check, since it seems more future-proof.

If you are pretty sure that everything returned from the IAM API will always be 200 for a success case, though, then this is good to go.

jonparrott:
> If you are pretty sure that everything returned from the IAM API will always be 200 for a success case, though, then this is good to go.

I feel like any change in that behavior should necessitate a change in our code. I also don't think they'll mess with it.

PR #109, opened 2017-02-15
Fix issue where GAE Signer erroneously returns a tuple from sign()

jonparrott:
/cc @matthewg

### DIFF
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -34,13 +34,30 @@ except ImportError:
 
 
 class Signer(object):
-    """Signs messages using the App Engine app identity service.
+    """Signs messages using the App Engine App Identity service.
 
     This can be used in place of :class:`google.auth.crypt.Signer` when
     running in the App Engine standard environment.
+
+    .. warning::
+        The App Identity service signs bytes using Google-managed keys.
+        Because of this it's possible that the key used to sign bytes will
+        change. In some cases this change can occur between successive calls
+        to :attr:`key_id` and :meth:`sign`. This could result in a signature
+        that was signed with a different key than the one indicated by
+        :attr:`key_id`. It's recommended that if you use this in your code
+        that you account for this behavior by building in retry logic.
     """
-    def __init__(self):
-        self.key_id = None
+
+    @property
+    def key_id(self):
+        """Optional[str]: The key ID used to identify this private key.
+
+        .. note::
+           This makes a request to the App Identity service.
+        """
+        key_id, _ = app_identity.sign_blob(b'')
+        return key_id
 
     @staticmethod
     def sign(message):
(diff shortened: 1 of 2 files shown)


google/auth/app_engine.py
```diff
@@ -34,13 +34,30 @@
...
-        self.key_id = None
+
+    @property
+    def key_id(self):
+        """Optional[str]: The key ID used to identify this private key.
+
+        .. note::
+           This makes a request to the App Identity service.
+        """
+        key_id, _ = app_identity.sign_blob(b'')
```
dhermes: 1. The RPC isn't particularly expensive?
1. Worth caching this return value?

tests/test_app_engine.py
```diff
@@ -41,6 +41,28 @@ def test_get_project_id_missing_apis():
     assert excinfo.match(r'App Engine APIs are not available')
 
 
+class TestSigner(object):
+    def test_key_id(self, app_identity_mock):
```
dhermes: `no-self-use`

jonparrott: Disabled for tests. This test file contains tests for two separate classes, so it leads that the test cases should be organized in classes.

google/auth/app_engine.py
jonparrott: 1) Not on GAE, as it only goes as far as the virtualization host.
2) No, same reason as IAM. Google can rotate the key.

dhermes: I realize the key can be rotated, but can't you invalidate the cached value when you sign?

jonparrott: The problem is that the order of operations is important here, think about this order of events:

1. `jwt.encode` is called.
2. `jwt.encode` uses `signer.key_id` to figure out the `kid` header. `signer` caches `key_id`.
3. `jwt.encode` uses `signer.sign` to sign the header and payload.
4. Time passes, key gets rotated.
5. `jwt.encode` is called again.
6. `jwt.encode` uses `signer.key_id` to figure out the `kid` header. `signer` *uses the cached `key_id`*, which is now wrong.
7. `jwt.encode` uses `signer.sign` to sign the header and payload.
8. A client tries to verify the jwt and fails because the kid doesn't match.

dhermes: Why does `jwt.encode` have to do it in that order?

jonparrott: Because the key id goes in the header, which is joined with the payload which is then signed. The key ID is needed to create the stuff we're signing.

tests/test_app_engine.py
dhermes: Yeah that **really** sucks

google/auth/app_engine.py
dhermes: Yeah that **really** sucks

jonparrott: totes

PR #115, opened 2017-02-17
Create abstract Verifier and Signer, remove key_id hack from App Engine and IAM signers

### LINKED ISSUE #112: Consider allowing Signer.key_id to be None
#108 and #109 added a "hack" to allow the key id to be determined when signing with IAM and App Engine. The reasoning for this is that the key id is needed to populate the `kid` claim in the JWT header.

It turns out, the `kid` field is optional according to the JWT spec. If it's absent, it's undefined what to do with it, but it seems that most clients will try all available certs (which is what we do as well).

@dhermes @lukesneeringer what do you think about allowing key_id to be None for `iam.Signer` and `app_engine.Signer`?

jonparrott:
Resolves #112

### DIFF
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -26,6 +26,7 @@ import datetime
 
 from google.auth import _helpers
 from google.auth import credentials
+from google.auth import crypt
 
 try:
     from google.appengine.api import app_identity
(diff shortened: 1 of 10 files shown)


google/auth/app_engine.py
```diff
@@ -26,49 +26,33 @@
...
     running in the App Engine standard environment.
-
-    .. warning::
-        The App Identity service signs bytes using Google-managed keys.
-        Because of this it's possible that the key used to sign bytes will
-        change. In some cases this change can occur between successive calls
-        to :attr:`key_id` and :meth:`sign`. This could result in a signature
-        that was signed with a different key than the one indicated by
-        :attr:`key_id`. It's recommended that if you use this in your code
-        that you account for this behavior by building in retry logic.
```
dhermes: Why did this die?

google/auth/crypt.py
```diff
@@ -193,6 +201,33 @@ def sign(self, message):
         Returns:
             bytes: The signature of the message.
         """
+        # pylint: disable=missing-raises-doc,redundant-returns-doc
+        # (pylint doesn't recognize that this is abstract)
+        raise NotImplementedError('Sign must be implemented')
+
+
+class RSASigner(Signer):
+    """Signs messages with a RSA private key.
```
dhermes: You said "an RSA" above. Which is it: "a" or "an"?

google/auth/crypt.py
```diff
@@ -193,6 +201,33 @@ def sign(self, message):
...
+        private_key (rsa.key.PrivateKey): The private key to sign with.
+        key_id (str): Optional key ID used to identify this private key. This
+            can be useful to associate the private key with its associated
+            public key or certificate.
+    """
+
+    def __init__(self, private_key, key_id=None):
+        self._key = private_key
+        self._key_id = key_id
+        """"""
```
dhermes: Sextuple quotes! WAT?

google/auth/app_engine.py
jonparrott: Because we no longer even try to get the key ID.

google/auth/crypt.py
jonparrott: `an`. Fixed.

google/auth/crypt.py
jonparrott: Removed, remnants of a docstring.

dhermes:
LGTM?

jonparrott:
> LGTM?

6-anchorman-quotes-burgundy

PR #116, opened 2017-02-17
Fix gRPC to call credentials.before_request

### LINKED ISSUE #113: Determine how to handle before_request for gRPC
I discovered that I did something totally wrong when I implemented our grpc transport.

All of our HTTP transports call `credentials.before_request` to get the authorization header populated ([here](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/8e1dc4f0345bbcfbe0f43ce40a5f9152b56db652/google/auth/transport/requests.py#L167)). However, the gRPC transport instead calls `credentials.refresh` directly and constructs the header itself ([here](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/8e1dc4f0345bbcfbe0f43ce40a5f9152b56db652/google/auth/transport/grpc.py#L50)).

I think the gRPC transport needs to do the same as the others, but that's where it gets tricky. `before_request` takes 4 arguments `request, method, url, request_headers`. gRPC provides our auth plugin with a context that looks like `AuthMetadataContext(service_url='https://pubsub.googleapis.com/google.pubsub.v1.Publisher', method_name='ListTopics')`. Does it make sense to call `before_request(http_request, context.method_name, context.service_url, headers)` even though `context.method_name` is not an HTTP method? It it okay to overload the term `method` for different transports?

@dhermes @lukesneeringer thoughts?

jonparrott:
Resolves #113

### DIFF
--- a/system_tests/test_grpc.py
+++ b/system_tests/test_grpc.py
@@ -15,23 +15,40 @@
 import google.auth
 import google.auth.credentials
 import google.auth.transport.grpc
-from google.cloud.gapic.pubsub.v1 import publisher_api
+from google.cloud.gapic.pubsub.v1 import publisher_client
 
 
-def test_grpc_request(http_request):
+def test_grpc_request_with_regular_credentials(http_request):
     credentials, project_id = google.auth.default()
     credentials = google.auth.credentials.with_scopes_if_required(
         credentials, ['https://www.googleapis.com/auth/pubsub'])
 
-    target = '{}:{}'.format(
-        publisher_api.PublisherApi.SERVICE_ADDRESS,
-        publisher_api.PublisherApi.DEFAULT_SERVICE_PORT)
-
     channel = google.auth.transport.grpc.secure_authorized_channel(
-        credentials, http_request, target)
+        credentials,
+        http_request,
+        publisher_client.PublisherClient.SERVICE_ADDRESS)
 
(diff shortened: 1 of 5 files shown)


system_tests/test_grpc.py
```diff
@@ -15,23 +15,40 @@
...
+        publisher_client.PublisherClient.SERVICE_ADDRESS)
+
+    # Create a pub/sub client.
+    client = publisher_client.PublisherClient(channel=channel)
+
+    # list the topics and drain the iterator to test that an authorized API
+    # call works.
+    list_topics_iter = client.list_topics(
+        project='projects/{}'.format(project_id))
+    list(list_topics_iter)
```
dhermes: And then....?

jonparrott:
> It is a little odd to me that neither system test actually asserts anything.

They do assert something, they assert the absence of an exception. The return value of the pubsub test could be an empty list. However, if the auth code fails (doesn't inject the header, injects the wrong header, etc.) then an exception will be raised.

system_tests/test_grpc.py
jonparrott: https://www.youtube.com/watch?v=CkdyU_eUm1U

But more seriously, we just drain the iterator. There's no need to check this value as it might be an empty list. The real test is that we can even execute this request without exceptions.

PR #117, opened 2017-02-22
Remove one-time token behavior of JWT Credentials

### LINKED ISSUE #114: Determine how to deal with multiple audiences for jwt.Credentials
JWT bearer credentials require an audience that matches the service URL to work.

Right now, JWT credentials don't require an audience when constructing, but instead can [generate a one-time token](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/8e1dc4f0345bbcfbe0f43ce40a5f9152b56db652/google/auth/jwt.py#L499) during `before_request`. This works, but unfortunately this is pretty awful performance-wise.

My best as to why we did this is because within one client we may actually be talking to two different audiences. Consider pubusb - there is the publisher audience (`https://pubsub.googleapis.com/google.pubsub.v1.Publisher`) and the subscriber audience (`https://pubsub.googleapis.com/google.pubsub.v1.Subscriber`).

I see two options for making this better:

1. Store a cache of tokens for every given audience. This has lots of problems.
2. Make our clients pass in `credentials.with_claims(audience='...')` (example [here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/c79fc926092a4953b1c9938ebc8ae74e4fc629b9/pubsub/google/cloud/pubsub/client.py#L102)) instead of just credentials. This means that we'll essentially be using a copy of the credentials for each audience (pubsub will have two), so they will likely refresh at different times (which I don't consider a problem).

@dhermes @lukesneeringer thoughts?

jonparrott:
Resolves #114

### DIFF
--- a/google/auth/jwt.py
+++ b/google/auth/jwt.py
@@ -333,9 +333,7 @@ class Credentials(credentials.Signing,
         """
         kwargs.setdefault('subject', info['client_email'])
         kwargs.setdefault('issuer', info['client_email'])
-        return cls(
-            signer,
-            **kwargs)
+        return cls(signer, **kwargs)
 
     @classmethod
     def from_service_account_info(cls, info, **kwargs):
--- a/google/auth/jwt.py
+++ b/google/auth/jwt.py
@@ -45,8 +45,6 @@ import collections
 import datetime
 import json
 
-from six.moves import urllib
-
 from google.auth import _helpers
 from google.auth import _service_account_info
 from google.auth import credentials
@@ -246,11 +244,7 @@ class Credentials(credentials.Signing,
     """Credentials that use a JWT as the bearer token.
 
     These credentials require an "audience" claim. This claim identifies the
-    intended recipient of the bearer token. You can set the audience when
-    you construct these credentials, however, these credentials can also set
-    the audience claim automatically if not specified. In this case, whenever
-    a request is made the credentials will automatically generate a one-time
-    JWT with the request URI as the audience.
+    intended recipient of the bearer token.
 
     The constructor arguments determine the claims for the JWT that is
     sent with requests. Usually, you'll construct these credentials with
@@ -260,13 +254,15 @@ class Credentials(credentials.Signing,
     JSON file::
 
         credentials = jwt.Credentials.from_service_account_file(
-            'service-account.json')
+            'service-account.json',
+            audience='https://speech.googleapis.com')
 
     If you already have the service account file loaded and parsed::
 
         service_account_info = json.load(open('service_account.json'))
         credentials = jwt.Credentials.from_service_account_info(
-            service_account_info)
+            service_account_info,
+            audience='https://speech.googleapis.com')
 
     Both helper methods pass on arguments to the constructor, so you can
     specify the JWT claims::
@@ -280,7 +276,10 @@ class Credentials(credentials.Signing,
     :class:`~google.auth.crypt.Signer` instance::
 
         credentials = jwt.Credentials(
-            signer, issuer='your-issuer', subject='your-subject')
+            signer,
+            issuer='your-issuer',
+            subject='your-subject',
+            audience=''https://speech.googleapis.com'')
 
     The claims are considered immutable. If you want to modify the claims,
     you can easily create another instance using :meth:`with_claims`::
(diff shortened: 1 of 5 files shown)


google/auth/jwt.py
```diff
@@ -260,13 +254,15 @@ class Credentials(credentials.Signing,
     JSON file::
 
         credentials = jwt.Credentials.from_service_account_file(
-            'service-account.json')
+            'service-account.json',
+            audience='https://speech.googleapis.com')
```
dhermes: Protocol (`https`) needed?

google/auth/jwt.py
```diff
@@ -289,7 +288,7 @@ class Credentials(credentials.Signing,
             audience='https://vision.googleapis.com')
     """
 
-    def __init__(self, signer, issuer=None, subject=None, audience=None,
+    def __init__(self, signer, issuer, subject, audience,
```
dhermes: Ouch

google/auth/jwt.py
```diff
@@ -334,7 +332,10 @@ def _from_signer_and_info(cls, signer, info, **kwargs):
             ValueError: If the info is not in the expected format.
         """
         kwargs.setdefault('subject', info['client_email'])
-        return cls(signer, issuer=info['client_email'], **kwargs)
+        kwargs.setdefault('issuer', info['client_email'])
+        return cls(
+            signer,
+            **kwargs)
```
dhermes: Overkill to put this on 3 lines instead of one?

google/auth/jwt.py
```diff
@@ -381,9 +382,8 @@ def with_claims(self, issuer=None, subject=None, audience=None,
                 claim will be used.
             subject (str): The `sub` claim. If unspecified the current subject
                 claim will be used.
-            audience (str): the `aud` claim. If not specified, a new
-                JWT will be generated for every request and will use
-                the request URI as the audience.
+            audience (str): the `aud` claim. If unspecified the current
```
dhermes: Sphinx wants double backticks? (Ditto on `sub` above)

google/auth/jwt.py
```diff
@@ -452,15 +433,8 @@ def refresh(self, request):
         # (pylint doesn't correctly recognize overridden methods.)
         self.token, self.expiry = self._make_jwt()
 
+    @_helpers.copy_docstring(credentials.Signing)
```
dhermes: Orthogonal change! How dare thee!

google/auth/jwt.py
```diff
@@ -472,32 +446,3 @@ def signer_email(self):
     @_helpers.copy_docstring(credentials.Signing)
     def signer(self):
         return self._signer
-
-    def before_request(self, request, method, url, headers):
```
dhermes: Why can this be wholesale deleted?

google/oauth2/service_account.py
```diff
@@ -223,13 +223,18 @@ def to_jwt_credentials(self):
             jwt_creds = jwt.Credentials.from_service_account_file(
                 'service_account.json')
 
+        Args:
+            audience (str): the `aud` claim. The intended audience for the
```
dhermes: Backtick backtick

google/auth/jwt.py
jonparrott: Yep.

google/auth/jwt.py
jonparrott: Ouch why?

google/auth/jwt.py

google/auth/jwt.py
jonparrott: Actually all of these are `<em>` and I'm okay with that (over `<code>`) but if you feel strongly I can change it.

google/auth/jwt.py
jonparrott: I'm a simple man, I see duplication, I remove it.

google/auth/jwt.py
jonparrott: Because once the one-time token logic is removed, it's identical to the superclass's `before_request`:

```
if not self.valid():
    self.refresh(request)
self.apply(headers)
```

google/auth/jwt.py
dhermes: Now they are all required, makes the constructor less usable

google/auth/jwt.py
dhermes: I do feel .... about it. Not sure what word goes there though.

dhermes:
We good?

jonparrott:
@dhermes if you're good & travis is good then I'm good.

google/auth/jwt.py
jonparrott: Ah I see. I think it was a mistake to make them optional before.

google/auth/jwt.py
jonparrott: I prefer to reserve code for well.. code. The claim names are part of the JWT spec, so I prefer to just em them.

dhermes:
Seems you gotta fix CI but LGTM otherwise

PR #125, opened 2017-03-01
Add helpful error messages when importing optional dependencies

### LINKED ISSUE #101: Consider helpful error messages when importing optional dependencies

jonparrott:
Resolves #101

### DIFF
--- a/google/auth/transport/grpc.py
+++ b/google/auth/transport/grpc.py
@@ -20,8 +20,8 @@ try:
     import grpc
 except ImportError:  # pragma: NO COVER
     raise ImportError(
-        'gRPC is not installed, please install grpcio to use the gRPC '
-        'transport.')
+        'gRPC is not installed, please install the grpcio package to use the '
+        'gRPC transport.')
 import six
 
 
--- a/google/auth/transport/grpc.py
+++ b/google/auth/transport/grpc.py
@@ -16,7 +16,12 @@
 
 from __future__ import absolute_import
 
-import grpc
+try:
+    import grpc
+except ImportError:  # pragma: NO COVER
+    raise ImportError(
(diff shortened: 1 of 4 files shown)


google/auth/transport/grpc.py
```diff
@@ -16,7 +16,12 @@
 
 from __future__ import absolute_import
 
-import grpc
+try:
+    import grpc
+except ImportError:  # pragma: NO COVER
+    raise ImportError(
+        'gRPC is not installed, please install grpcio to use the gRPC '
```
dhermes: Worth mentioning that `grpcio` is a PyPI package?

jonparrott:
> How exhaustive were you in searching for places where this fails?

Pretty exhaustive

> Can you add a tox environment without the deps instead of doing the pragma: NO COVER? Then you can integrate that into the coverage report?

This'll make the coverage environment impossible to run standalone, unless I'm misunderstanding the request?

google/auth/transport/grpc.py
jonparrott: Done (for all imports)

dhermes:
> This'll make the coverage environment impossible to run standalone, unless I'm misunderstanding the request?

That's correct, but the contents of this PR indicate that true coverage could not be done in a single environment.

jonparrott:
> That's correct, but the contents of this PR indicate that true coverage could not be done in a single environment.

@dhermes that was already the case, we have several import guards.

dhermes:
We can do better than sprinkling the `NO COVER`-s, so why not do better?

jonparrott:
@dhermes per our side conversion, created #127 to discuss this separately.

dhermes:
SGTM LGTM

PR #128, opened 2017-03-02
Add InstalledAppFlow

### LINKED ISSUE #122: Add interactive installed app flow
Create a helper method for running an interactive installed app flow, similar to:

```python
def credentials_flow_interactive(client_secrets_file, scopes):
    """Initiate an interactive OAuth2InstalledApp flow.

    - Display a URL for the user to visit.
    - URL displays a code for the user to copy.
    - Wait on standard input for the user to enter the provided code.
    - Exchange OAuth2 tokens.

    Args:
      client_secrets_file: The path to the client secrets JSON file.
      scopes: The list of scopes to request during the flow.
    """
    flow = google.oauth2.flow.Flow.from_client_secrets_file(
        client_secrets_file,
        scopes=scopes,
        redirect_uri='urn:ietf:wg:oauth:2.0:oob')
    auth_url, _ = flow.authorization_url(prompt='consent')
    print('Please go to this URL: %s' % auth_url)
    code = input('Enter the authorization code: ')
    flow.fetch_token(code=code)
    return flow.credentials
```

jonparrott:
Resolves #122

### DIFF
--- a/additional_packages/google_auth_oauthlib/google_auth_oauthlib/flow.py
+++ b/additional_packages/google_auth_oauthlib/google_auth_oauthlib/flow.py
@@ -58,10 +58,9 @@ import webbrowser
 import wsgiref.simple_server
 import wsgiref.util
 
-from six.moves import input
-
 import google.auth.transport.requests
 import google.oauth2.credentials
+from six.moves import input
 
 import google_auth_oauthlib.helpers
 
@@ -281,7 +280,7 @@ class InstalledAppFlow(Flow):
 
     Example::
 
-        from google.oauth2.flow import InstalledAppFlow
+        from google_auth_oauthlib.flow import InstalledAppFlow
 
         flow = InstalledAppFlow.from_client_secrets_file(
             'client_secrets.json',
--- a/additional_packages/google_auth_oauthlib/google_auth_oauthlib/flow.py
+++ b/additional_packages/google_auth_oauthlib/google_auth_oauthlib/flow.py
@@ -50,11 +50,21 @@ authorization flow::
 """
 
 import json
+import logging
+import webbrowser
+
+from six.moves import BaseHTTPServer
+from six.moves import http_client
+from six.moves import input
+from six.moves import urllib
 
 import google.auth.transport.requests
 import google.oauth2.credentials
 
 import google_auth_oauthlib.helpers
+
+
+_LOGGER = logging.getLogger(__name__)
 
 
 class Flow(object):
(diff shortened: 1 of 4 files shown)


jonparrott:
This is DO NOT MERGE because I want a sanity check before I go any further.

1. What do you think about this being in the library in general?
2. What do you think about the way the implementation is structured?

Specifically for (2), I had some thoughts that it might make more sense to separate the strategies into separate classes:

```python
flow = google.oauth2.flow.InstalledAppFlow.from_client_secrets_file(...)

strategy = flow.ServerStrategy(port=9000)
# or
strategy = flow.ConsoleStrategy()

flow.run(strategy)
# or
flow.strategy = strategy
flow.run()
```

But I'm unsure if that's over-engineering, even if it feels cleaner.

lukesneeringer:
Can you provide some sample code for what it looks like with a single, combined flow? It would make it easier to reason about without doing a deep dive.

jonparrott:
@lukesneeringer not sure what you're asking for.

lukesneeringer:
> @lukesneeringer not sure what you're asking for.

I was referring to (2) specifically.

Right now it seems to be:

```python
flow = InstalledAppFlow(...)
flow.run(strategy=SERVER, port=9000)
```

I think that is fine. I do not think strategy objects are necessary. And it does seem reasonable that some arguments (like `port`) are ignored if they are not applicable to the strategy in use.

jonparrott:
@lukesneeringer that's nearly correct, it would be:

```python
flow = InstalledAppFlow(...)
flow.web_port = 9000
flow.run(strategy=SERVER)
```

I can keep that same interface but separate the internals into two separate strategy classes. Curious as to what @dhermes thinks as well.

lukesneeringer:
That makes a little less sense to be, but it certainly seems good enough. And it is probably better than making the user create a strategy object.

jonparrott:
>That makes a little less sense to me

It makes more sense to pass into `run` then we can do that. There's plenty of malleability here.

Another option is to have two `run`s: `run_console` and `run_server`.

lukesneeringer:
Oooh, I think `run_console` and `run_server` taking the appropriate arguments is a good approach.

jonparrott:
> Oooh, I think run_console and run_server taking the appropriate arguments is a good approach.

Cool, that's what I'll go with. I'll remove `run` and just have `run_server` and `run_console`.

jonparrott:
@dhermes @lukesneeringer this is ready for review. :)

jonparrott:
@proppy it would be super cool if you took a look as well.

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+        in their browser. Once the authorization is complete the authorization
+        server will give the user a code. The user then most copy & paste this
+        code into the application. The code is then exchanged for a token.
+
+        Args:
+            authorization_prompt_message (str): The message to display to tell
+                the user to navigate to the authorization URL.
+            authorization_code_message (str): The message to display when
+                prompting the user for the authorization code.
+            kwargs: Additional key-word arguments passed through to
```
dhermes: Is keyword hyphenated?

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+                :meth:`authorization_url`.
+
+        Returns:
+            google.oauth2.credentials.Credentials: The OAuth 2.0 credentials
+                for the user.
+        """
+        kwargs.setdefault('prompt', 'consent')
+
+        self.redirect_uri = self._OOB_REDIRECT_URI
```
dhermes: Why set this here and not somewhere like the constructor?

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+
+        auth_url, _ = self.authorization_url(**kwargs)
+
+        print(authorization_prompt_message.format(url=auth_url))
```
dhermes: You actually **want** `print`, right?

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+        The server strategy instructs the user to open the authorization URL in
+        their browser and will attempt to automatically open the URL for them.
+        It will start a local web server to listen for the authorization
+        response. Once authorization is complete the authorization server will
+        redirect the user's browser to the local web server. The web server
+        will get the authorization code from the response and shutdown. The
+        code is then exchanged for a token.
+
+        Args:
+            host (str): The web host for the local redirect server.
```
dhermes: 1. Also here: `(Optional[str])`
1. Worth noting that you assuming it is `http` not `https`

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+
+        authorization_response = urllib.parse.urljoin(
+            # Note: using https here because oauthlib is very picky that
+            # OAuth 2.0 should only occur over https.
+            'https://localhost', local_server.last_request_path)
+        self.fetch_token(authorization_response=authorization_response)
+
+        return self.credentials
+
+    class _LocalRedirectServer(BaseHTTPServer.HTTPServer):
```
dhermes: I am BIGLY :-1: on classes defined anywhere other than top-level. Why is this defined here rather than at module scope?

setup.py
```diff
@@ -39,7 +39,7 @@
     description='Google Authentication Library',
     long_description=long_description,
     url='https://github.com/GoogleCloudPlatform/google-auth-library-python',
-    packages=find_packages(exclude=('tests', 'system_tests')),
+    packages=find_packages(exclude=('tests*', 'system_tests*')),
```
dhermes: Isn't that other PR merged already?

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+                for the user.
+        """
+        self.redirect_uri = 'http://{}:{}/'.format(host, port)
+
+        auth_url, _ = self.authorization_url(**kwargs)
+
+        local_server = self._LocalRedirectServer(
+            success_message, (host, port), self._LocalRedirectRequestHandler)
+
+        webbrowser.open(auth_url, new=1, autoraise=True)
```
proppy: can we make that optional?

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+        return self.credentials
+
+    class _LocalRedirectServer(BaseHTTPServer.HTTPServer):
+        """A server to handle OAuth 2.0 redirects back to localhost."""
+        def __init__(self, success_message, *args, **kwargs):
+            BaseHTTPServer.HTTPServer.__init__(self, *args, **kwargs)
+            self.success_message = success_message
+            self.last_request_path = None
+
+    class _LocalRedirectRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
```
proppy: curious if you took a look at [wsgiref](https://docs.python.org/2/library/wsgiref.html#wsgiref.simple_server.make_server) instead? It doesn't require you creating a class etc.

proppy:
do we want a "smart" `run` that fallback to run_console if opening the browser failed?

google/oauth2/flow.py
```diff
@@ -54,12 +54,22 @@
 """
 
 import json
+import logging
+import webbrowser
```
jonparrott: If by "really doing this" you mean:

1. Using `webbrowser`, then yes, I think it's important if we want to be on parity with what oauth2client offered.
2. Having this code period - If you feel like this code shouldn't be here, by all means I'm happy to have that conversation. We have a lot of options, so if you feel it doesn't belong in `google-auth` proper let's talk.

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+
+    This flow has two strategies: The console strategy provided by
+    :meth:`run_console` and the local server strategy provided by
+    :meth:`run_local_server`.
+
+    Example::
+
+        import google.oauth2.flow
+
+        flow = google.oauth2.flow.InstalledAppFlow.from_client_secrets_file(
```
jonparrott: Yep but I hate importing classes. I'll do it anyway for the example.

google/oauth2/flow.py
```diff
@@ -256,3 +266,175 @@ class using this flow's :attr:`credentials`.
...
+            scopes=['profile', 'email'])
+
+        flow.run_local_server()
+
+        session = flow.authorized_session()
+
+        profile_info = session.get(
+            'https://www.googleapis.com/userinfo/v2/me').json()
+
+        print(profile_info)
```
jonparrott: 1. Will do
2. You don't have to, it handles one request (blocking) then shuts down.

google/oauth2/flow.py
jonparrott: I have no idea why I forgot about wsgiref. Thank you, I've updated this to use that instead.

google/oauth2/flow.py
jonparrott: Fixed, also switch this to wsgiref.

jonparrott:
> do we want a "smart" run that fallback to run_console if opening the browser failed?

Webbrowser doesn't give us any indication if the `open` call was successful or not.

google/oauth2/flow.py
```diff
@@ -54,12 +54,22 @@
...
+from six.moves import http_client
+from six.moves import input
+from six.moves import urllib
 
 import google.auth.transport.requests
 import google.oauth2.credentials
 import google.oauth2.oauthlib
 
 
+_LOGGER = logging.getLogger(__name__)
```
jonparrott: This is so the the request handler below doesn't write to stderr, but instead uses our named logger.

jonparrott:
@dhermes @lukesneeringer should be good to review again.

google/oauth2/flow.py
dhermes: I meant "having this code period" (and `webbrowser` was the "aw crap" moment)

google/oauth2/flow.py
dhermes: You could add the logger in a separate, less large, less controversial PR?

jonparrott:
> I meant "having this code period" 

The good news is that it's in a separate package now, woohoo. :)

additional_packages/google_auth_oauthlib/google_auth_oauthlib/flow.py
```diff
@@ -253,3 +264,195 @@ class using this flow's :attr:`credentials`.
...
+            google.oauth2.credentials.Credentials: The OAuth 2.0 credentials
+                for the user.
+        """
+        kwargs.setdefault('prompt', 'consent')
+
+        self.redirect_uri = self._OOB_REDIRECT_URI
+
+        auth_url, _ = self.authorization_url(**kwargs)
+
+        print(authorization_prompt_message.format(url=auth_url))
```
proppy: I wonder if this could be implemented with a callback. That could the calling code to fire a `webbrowser` instead of just printing the URL (and maybe be re-used by `run_local_server`).

PR #131, opened 2017-03-16
Fix inclusion of tests package in setup.py

jonparrott:
@lukesneeringer @dhermes FYI, this probably needs to be double-checked in google-cloud as well.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,2 @@
-include README.rst
-graft google
-global-exclude *.pyc __pycache__
+include README.rst LICENSE.txt CHANGELOG.rst
+recursive-include tests *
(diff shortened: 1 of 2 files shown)


google/oauth2/flow.py
```diff
@@ -54,12 +54,22 @@
...
+from six.moves import http_client
+from six.moves import input
+from six.moves import urllib
 
 import google.auth.transport.requests
 import google.oauth2.credentials
 import google.oauth2.oauthlib
 
 
+_LOGGER = logging.getLogger(__name__)
```
dhermes: This change isn't related to `dist`

jonparrott: Fixed.

jonparrott:
> LGTM but like squash those commits, eh? 🇨🇦

Fo' sho'

PR #139, opened 2017-03-22
Remove oauthlib modules from documentation, update user guide

jonparrott:


jonparrott:
@dhermes @lukesneeringer FYI, docs-only change.

PR #140, opened 2017-03-22
Add service_account.Credentials.with_claims

### LINKED ISSUE #129: Can't update additional claims with service account credentials

jonparrott:
Resolves #129.

### DIFF
--- a/google/auth/jwt.py
+++ b/google/auth/jwt.py
@@ -426,13 +426,15 @@ class Credentials(google.auth.credentials.Signing,
         Returns:
             google.auth.jwt.Credentials: A new credentials instance.
         """
+        new_additional_claims = self._additional_claims.copy()
+        new_additional_claims.update(additional_claims or {})
+
         return Credentials(
             self._signer,
             issuer=issuer if issuer is not None else self._issuer,
             subject=subject if subject is not None else self._subject,
             audience=audience if audience is not None else self._audience,
-            additional_claims=self._additional_claims.copy().update(
-                additional_claims or {}))
+            additional_claims=new_additional_claims)
 
     def _make_jwt(self):
         """Make a signed JWT.
--- a/google/auth/jwt.py
+++ b/google/auth/jwt.py
@@ -42,6 +42,7 @@ You can also skip verification::
 
 import base64
 import collections
+import copy
 import datetime
 import json
 
@@ -426,7 +427,7 @@ class Credentials(google.auth.credentials.Signing,
         Returns:
             google.auth.jwt.Credentials: A new credentials instance.
         """
-        new_additional_claims = self._additional_claims.copy()
+        new_additional_claims = copy.deepcopy(self._additional_claims)
         new_additional_claims.update(additional_claims or {})
 
         return Credentials(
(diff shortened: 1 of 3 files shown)


google/auth/jwt.py
```diff
@@ -426,13 +426,15 @@ def with_claims(self, issuer=None, subject=None, audience=None,
         Returns:
             google.auth.jwt.Credentials: A new credentials instance.
         """
+        new_additional_claims = self._additional_claims.copy()
```
dhermes: You worried about nesting? `deepcopy`?

google/auth/jwt.py
```diff
@@ -426,13 +426,15 @@ def with_claims(self, issuer=None, subject=None, audience=None,
...
+        new_additional_claims.update(additional_claims or {})
```
dhermes: If you made `additional_claims=()` by default you could ditch the `or {}` (FWIW `()` is not strictly needed, any "empty" iterable will do)

google/oauth2/service_account.py
```diff
@@ -246,6 +246,29 @@ def with_subject(self, subject):
             subject=subject,
             additional_claims=self._additional_claims.copy())
 
+    def with_claims(self, additional_claims):
```
dhermes: This is mostly / totally copied from the `jwt` module?

jonparrott: Yep.

google/auth/jwt.py
jonparrott: Eh, that borders on too clever for my likes.

google/auth/jwt.py
jonparrott: Nah, not horribly concerned about deep nesting, it's pretty rare to do that in the jwt payload.

jonparrott: But if you want to be extra sure I can do deepcopy.

google/auth/jwt.py
dhermes: That would be the least clever thing I've heard called too clever.

google/oauth2/service_account.py
dhermes: Possibly to DRY?

google/auth/jwt.py
dhermes: Up to you

google/oauth2/service_account.py
jonparrott: Nah, jwt accepts a different set of arguments here (like audience, etc.). One one hand I know there's some duplicating here, but I'm not sure if there's a good way to refactor without messiness.

google/auth/jwt.py
jonparrott: 😆

dhermes:
LGTM I suppose. It seems you want to eat the teeny cost of `or {}` every time, so it's all good?

jonparrott:
>It seems you want to eat the teeny cost of or {} every time, so it's all good?

The cost of that vs. the cost of copy/deepcopy is insignificant.

I'm going to update this to use deepcopy, but I'm finishing something else up first. Thanks for the review!

PR #142, opened 2017-03-23
Add jwt.OnDemandCredentials

### LINKED ISSUE #136: Add on-demand jwt credentials
`jwt.Credentials` only works for a single audience at a time and must have knowledge of the audience beforehand. gRPC APIs, however, may not be able to determine the audience beforehand so JWTs should be generated on-demand during the `before_request` callback.

We previously had this behavior as part of `jwt.Credentials`, but it was removed because it made the behavior of the class ambiguous. This feature request is to bring that functionality back as a separate class `jwt.OnDemandCredentials`.

This class:
* Doesn't accept an `audience` argument in its constructor.
* Can be created from existing credentials using [`from_signing_credentials`](http://google-auth.readthedocs.io/en/latest/reference/google.auth.jwt.html#google.auth.jwt.Credentials.from_signing_credentials).
* Holds a cache of JWTs for specific audiences using [`cachetools.LRUCache`](http://pythonhosted.org/cachetools/#cachetools.LRUCache).
* Generates a new JWT for the specific audience in `before_request` or uses an existing cached JWT.
* Is always valid and never expires but never has a `token`.

[Context](https://github.com/GoogleCloudPlatform/google-auth-library-python/issues/93#issuecomment-288285350). 

@dhermes @lukesneeringer any concerns about the name or dependency on `cachetools`?
@jboeuf any concerns on the behavior here?

jonparrott:
Resolves #136.

Note: this adds a new dependency on `cachetools`.

### DIFF
--- a/google/auth/jwt.py
+++ b/google/auth/jwt.py
@@ -46,13 +46,17 @@ import copy
 import datetime
 import json
 
+import cachetools
+from six.moves import urllib
+
 from google.auth import _helpers
 from google.auth import _service_account_info
 from google.auth import crypt
+from google.auth import exceptions
 import google.auth.credentials
 
-
-_DEFAULT_TOKEN_LIFETIME_SECS = 3600  # 1 hour in seconds
+_DEFAULT_TOKEN_LIFETIME_SECS = 3600  # 1 hour in sections
+_DEFAULT_MAX_CACHE_SIZE = 10
 
 
 def encode(signer, payload, header=None, key_id=None):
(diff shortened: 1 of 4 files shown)


jboeuf:
waow. Awesome. thanks much for the quick turnaround!

jonparrott:
> waow. Awesome. thanks much for the quick turnaround!

We had most of this code already, we just needed justification for it.

jboeuf:
Very much appreciate either way :).

On Thu, Mar 23, 2017 at 1:02 PM, Jon Wayne Parrott <notifications@github.com

jonparrott:
@dhermes @lukesneeringer re-based and should be ready for review. :)

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+            ValueError: If the info is not in the expected format.
+        """
+        kwargs.setdefault('subject', info['client_email'])
+        kwargs.setdefault('issuer', info['client_email'])
+        return cls(signer, **kwargs)
+
+    @classmethod
+    def from_service_account_info(cls, info, **kwargs):
+        """Creates an OnDemandCredentials instance from a dictionary containing
+        service account info in Google format.
```
dhermes: Would it be so heinous if we tried to be PEP257 here?

```python
        """Creates an OnDemandCredentials instance from a dictionary.

        Expects the dictionary to contain service account info in
        Google format.
        """
```

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+            kwargs: Additional arguments to pass to the constructor.
+
+        Returns:
+            google.auth.jwt.Credentials: A new Credentials instance.
+        """
+        kwargs.setdefault('issuer', credentials.signer_email)
+        kwargs.setdefault('subject', credentials.signer_email)
+        return cls(
+            credentials.signer,
+            **kwargs)
```
dhermes: Overkill continuation?

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+                claim will be used.
+            additional_claims (Mapping[str, str]): Any additional claims for
+                the JWT payload. This will be merged with the current
+                additional claims.
+
+        Returns:
+            google.auth.jwt.OnDemandCredentials: A new credentials instance.
+        """
+        new_additional_claims = copy.deepcopy(self._additional_claims)
+        new_additional_claims.update(additional_claims or {})
```
dhermes: You and your extra branchy work (`or` and `not`)

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+            issuer=issuer if issuer is not None else self._issuer,
+            subject=subject if subject is not None else self._subject,
+            additional_claims=new_additional_claims,
+            max_cache_size=self._cache.maxsize)
+
+    @property
+    def valid(self):
+        """Checks the validity of the credentials.
+
+        These credentials are always valid.
```
dhermes: ORLY?

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+        the audience, that token is used. Otherwise, a new token will be
+        created.
+
+        Args:
+            audience (str): The intended audience.
+
+        Returns:
+            bytes: The encoded JWT.
+        """
+
```
dhermes: Empty for fun?

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+        refreshed.
+
+        Args:
+            request (Any): Unused.
+
+        Raises:
+            google.auth.RefreshError
+        """
+        # pylint: disable=unused-argument
+        # (pylint doesn't correctly recognize overridden methods.)
```
dhermes: Shall we file an issue with Pylint? @PCManticore has in general been pretty great in situations like this.

tests/test_jwt.py
```diff
@@ -343,3 +344,135 @@ def test_before_request_refreshes(self):
         self.credentials.before_request(
             None, 'GET', 'http://example.com?a=1#3', {})
         assert self.credentials.valid
+
+
+class TestOnDemandCredentials:
```
dhermes: Are you a fan of creating old-style classes (in Python 2)?

tests/test_jwt.py
```diff
@@ -343,3 +344,135 @@ def test_before_request_refreshes(self):
...
+    SERVICE_ACCOUNT_EMAIL = 'service-account@example.com'
+    SUBJECT = 'subject'
+    ADDITIONAL_CLAIMS = {'meta': 'data'}
+    credentials = None
```
dhermes: Why not `CREDENTIALS`?

jonparrott:
> What should I know about cachetools (never having used it)? Worth making it an optional dependency?

It's a very well tested, maintained, and focused library that provides a pretty damn good implementation of an LRU cache. We could alternatively just pull in the LRUCache class (and associated tests) into this library.

dhermes:
Your call

google/auth/jwt.py
```diff
@@ -46,13 +46,17 @@
...
+
 from google.auth import _helpers
 from google.auth import _service_account_info
 from google.auth import crypt
+from google.auth import exceptions
 import google.auth.credentials
 
-
-_DEFAULT_TOKEN_LIFETIME_SECS = 3600  # 1 hour in seconds
+_DEFAULT_TOKEN_LIFETIME_SECS = 3600  # 1 hour in sections
```

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+
+    def __init__(self, signer, issuer, subject,
+                 additional_claims=None,
+                 token_lifetime=_DEFAULT_TOKEN_LIFETIME_SECS,
+                 max_cache_size=_DEFAULT_MAX_CACHE_SIZE):
+        """
+        Args:
+            signer (google.auth.crypt.Signer): The signer used to sign JWTs.
+            issuer (str): The `iss` claim.
+            subject (str): The `sub` claim.
```
jonparrott: italics. We had this same discussion when I made `jwt.Credentials`. _sub_ and _iss_ are jwt claims, not identifiers.

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+            additional_claims (Mapping[str, str]): Any additional claims for
```
jonparrott: Arguments with a default value are implicitly `Optional`. (Optional is useful for return values and properties).

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+                the JWT payload.
+            token_lifetime (int): The amount of time in seconds for
+                which the token is valid. Defaults to 1 hour.
+            max_cache_size (int): The maximum number of JWT tokens to keep in
+                cache. Tokens are cached using :class:`cachetools.LRUCache`.
+        """
+        super(OnDemandCredentials, self).__init__()
```
jonparrott: Not sure what you mean here?

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+        self._signer = signer
+        self._issuer = issuer
+        self._subject = subject
+        self._token_lifetime = token_lifetime
+
+        if additional_claims is not None:
+            self._additional_claims = additional_claims
+        else:
+            self._additional_claims = {}
```
jonparrott: Done. (in `jwt.Credentials` as well)

google/auth/jwt.py
```diff
@@ -487,3 +491,270 @@ def signer_email(self):
...
+            kwargs: Additional arguments to pass to the constructor.
+
+        Returns:
+            google.auth.jwt.OnDemandCredentials: The constructed credentials.
+
+        Raises:
+            ValueError: If the info is not in the expected format.
+        """
+        kwargs.setdefault('subject', info['client_email'])
+        kwargs.setdefault('issuer', info['client_email'])
```
jonparrott: No, generally the helper in `_service_account` loads the file information and that does validation that the fields are present.

google/auth/jwt.py
jonparrott: Nope, done. :)

google/auth/jwt.py
jonparrott: Yep, fixed.

google/auth/jwt.py
jonparrott: *insert shruggie here*

google/auth/jwt.py
jonparrott: RLY. (I added some more information about why this is).

google/auth/jwt.py
jonparrott: For fun or for business. Removed.

google/auth/jwt.py
jonparrott: Maybe.

tests/test_jwt.py
jonparrott: Fixed (although pytest didn't mind)

tests/test_jwt.py
jonparrott: It's not a constant (see the fixture below).

jonparrott:
I'm okay with the dependency. @lukesneeringer WDYT?

google/auth/jwt.py
dhermes: This class has two parents, and the `super()` MRO resolution is well-defined but not typically well-known by authors. Just wondering if you're worried that using `super()` is too unclear.

jonparrott: super is fine, remember that `Signing` is just a mixin interface.

lukesneeringer:
Just did some quick research. I prefer the dep over vendoring.

PR #147, opened 2017-03-23
Use "gcloud config config-helper" to read the project ID from the Google Cloud SDK

### LINKED ISSUE #146: Stop reading Cloud SDK's configuration to obtain the project ID
Instead, we should subprocess out to `gcloud config config-helper --format json`

Example output:

```
{
  "configuration": {
    "active_configuration": "default",
    "properties": {
      "core": {
        "account": "[ELIDED]",
        "disable_usage_reporting": "False",
        "project": "[ELIDED]"
      }
    }
  },
  "credential": {
    "access_token": "[ELIDED]",
    "token_expiry": "2017-03-23T23:09:49Z"
  },
  "sentinels": {
    "config_sentinel": "/Users/jonwayne/.config/gcloud/config_sentinel"
  }
}
```

Note: we should *not* use the auth token provided here at all. We should keep our existing logic of reading `~/.config/gcloud/application_default_credentials.json` because those credentials are populated exclusively by `gcloud auth application-default login` and intended for Application Default Credentials.

jonparrott:
Resolves #146

### DIFF
--- a/google/auth/_cloud_sdk.py
+++ b/google/auth/_cloud_sdk.py
@@ -117,8 +117,8 @@ def get_project_id():
 
     try:
         output = subprocess.check_output(
-            _CLOUD_SDK_CONFIG_COMMAND,)
-            #stderr=subprocess.STDOUT)
+            _CLOUD_SDK_CONFIG_COMMAND,
+            stderr=subprocess.STDOUT)
     except (subprocess.CalledProcessError, OSError, IOError):
         return None
 
--- a/google/auth/_cloud_sdk.py
+++ b/google/auth/_cloud_sdk.py
@@ -33,6 +33,9 @@ _WINDOWS_CONFIG_ROOT_ENV_VAR = 'APPDATA'
 # The name of the file in the Cloud SDK config that contains default
 # credentials.
 _CREDENTIALS_FILENAME = 'application_default_credentials.json'
+# The command to get the Cloud SDK configuration
+_CLOUD_SDK_CONFIG_COMMAND = (
+    'gcloud', 'config', 'config-helper', '--format', 'json')
 
 
 def get_config_path():
@@ -113,9 +116,9 @@ def get_project_id():
     """
 
     try:
-        output = subprocess.check_output([
-            'gcloud', 'config', 'config-helper', '--format', 'json'],
-            stderr=subprocess.STDOUT)
+        output = subprocess.check_output(
+            _CLOUD_SDK_CONFIG_COMMAND,)
+            #stderr=subprocess.STDOUT)
     except (subprocess.CalledProcessError, OSError, IOError):
         return None
 
--- a/google/auth/_cloud_sdk.py
+++ b/google/auth/_cloud_sdk.py
@@ -14,11 +14,11 @@
 
 """Helpers for reading the Google Cloud SDK's configuration."""
 
-import io
+import json
 import os
+import subprocess
 
 import six
-from six.moves import configparser
 
 from google.auth import environment_vars
 import google.oauth2.credentials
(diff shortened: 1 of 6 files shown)


jonparrott:
Resolves #147

google/auth/_cloud_sdk.py
```diff
@@ -166,3 +103,27 @@ def load_authorized_user_credentials(info):
...
+def get_project_id():
+    """Gets the project ID from the Cloud SDK.
+
+    Returns:
+        Optional[str]: The project ID.
+    """
+
+    try:
+        output = subprocess.check_output([
+            'gcloud', 'config', 'config-helper', '--format', 'json'])
```
dhermes: Make these args a non-public constant (as a `tuple` not a `list`)?

google/auth/_cloud_sdk.py
```diff
@@ -166,3 +103,27 @@ def load_authorized_user_credentials(info):
...
+    except (subprocess.CalledProcessError, OSError, IOError):
```
dhermes: Why catch all three?

google/auth/_cloud_sdk.py
```diff
@@ -166,3 +103,27 @@ def load_authorized_user_credentials(info):
...
+        return None
+
+    try:
+        configuration = json.loads(output.decode('utf-8'))
+    except ValueError:
+        return None
+
+    try:
+        return configuration['configuration']['properties']['core']['project']
+    except KeyError:
```
dhermes: This masks many potential issues. We don't worry about the 4 possible `KeyErrors` that could occur?

tests/data/cloud_sdk_config.json
```diff
@@ -0,0 +1,19 @@
...
+        "disable_usage_reporting": "False",
+        "project": "example-project"
+      }
+    }
+  },
+  "credential": {
+    "access_token": "don't use me",
+    "token_expiry": "2017-03-23T23:09:49Z"
+  },
+  "sentinels": {
```
dhermes: Seems foreboding: https://www.youtube.com/watch?v=m3PUjumx3ag

google/auth/_cloud_sdk.py
jonparrott: because `CalledProcessError` will happen if gcloud returns anything other than 0 (for example, any sort of un-configured state), and depending on the OS / Python version it'll raise OSError or IOError if `gcloud` isn't installed.

google/auth/_cloud_sdk.py
jonparrott: Nope, if the configuration ins't in the exact format we expect then we should fail gracefully.

tests/data/cloud_sdk_config.json
jonparrott: haha I didn't come up with this format. *shruggie*.

google/auth/_cloud_sdk.py
jonparrott: Done.

google/auth/_cloud_sdk.py
dhermes: Why could there be two **different** errors if `gcloud` isn't on `PATH`?

jonparrott:
> You definitely don't want commented out code in your commit

Trying to debug the system tests, I figured it out. Fix incoming.

google/auth/_cloud_sdk.py
jonparrott: It's a quirk between windows / unix. Ran into the same thing with Nox.

In general, it's better for this method to err on the side of being too broad in error handling than to potentially explode just because it can't talk to gcloud and figure out the project.

jonparrott:
@dhermes this should be ready for final review.

PR #150, opened 2017-03-27
Expose id_token in OAuth 2.0 credentials

### LINKED ISSUE #149: Expose the full response from the token server
`refresh` on oauth2 Credentials should store the full response from the token server. There is potentially useful data here, like the `id_token`.

jonparrott:
Resolves #149

### DIFF
--- a/google/oauth2/credentials.py
+++ b/google/oauth2/credentials.py
@@ -67,6 +67,7 @@ class Credentials(credentials.Scoped, credentials.Credentials):
         self._token_uri = token_uri
         self._client_id = client_id
         self._client_secret = client_secret
+        self._refresh_grant_response = None
 
     @property
     def refresh_token(self):
@@ -90,6 +91,17 @@ class Credentials(credentials.Scoped, credentials.Credentials):
         return self._client_secret
 
     @property
+    def refresh_grant_response(self):
+        """Optional[str]: The last response from the OAuth 2.0 token server.
+
+        This is set when :meth:`refresh` is called and will not be populated
+        otherwise. This is provided because some authorization servers will
+        send along additional information other than the access and refresh
+        tokens in the refresh grant response.
+        """
+        return self._refresh_grant_response
+
+    @property
     def requires_scopes(self):
         """False: OAuth 2.0 credentials have their scopes set when
         the initial token is requested and can not be changed."""
@@ -106,10 +118,12 @@ class Credentials(credentials.Scoped, credentials.Credentials):
 
     @_helpers.copy_docstring(credentials.Credentials)
     def refresh(self, request):
-        access_token, refresh_token, expiry, _ = _client.refresh_grant(
-            request, self._token_uri, self._refresh_token, self._client_id,
-            self._client_secret)
+        access_token, refresh_token, expiry, grant_response = (
+            _client.refresh_grant(
+                request, self._token_uri, self._refresh_token, self._client_id,
+                self._client_secret))
 
         self.token = access_token
         self.expiry = expiry
         self._refresh_token = refresh_token
+        self._refresh_grant_response = grant_response
--- a/google/oauth2/credentials.py
+++ b/google/oauth2/credentials.py
@@ -128,4 +128,4 @@ class Credentials(credentials.Scoped, credentials.Credentials):
         self.token = access_token
         self.expiry = expiry
         self._refresh_token = refresh_token
-        self._id_token = grant_response.get('id_token', self._id_token)
+        self._id_token = grant_response.get('id_token')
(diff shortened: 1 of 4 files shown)


google/oauth2/credentials.py
```diff
@@ -90,6 +91,17 @@ def client_secret(self):
         return self._client_secret
 
     @property
+    def refresh_grant_response(self):
```
dhermes: You sure sure you want this to be public?

google/oauth2/service_account.py
```diff
@@ -306,10 +319,11 @@ def _make_authorization_grant_assertion(self):
     @_helpers.copy_docstring(credentials.Credentials)
     def refresh(self, request):
         assertion = self._make_authorization_grant_assertion()
-        access_token, expiry, _ = _client.jwt_grant(
+        access_token, expiry, refresh_grant_response = _client.jwt_grant(
```
dhermes: Why not just

```python
        self.token, self.expiry, self._refresh_grant_response = _client.jwt_grant(...)
```

jonparrott: that's a good point. Hmm. I guess I wanted to use the same names that the `jwt_grant` function used for those values, then show mapping those values to properties (which have slightly different names). But I'm good to change it if you want.

google/oauth2/credentials.py
jonparrott: Yep, I actually found a use case where this is necessary. When you refresh a token, you actually get both an access token and an id token back in the response. The other option is to add an `id_token` field to `oauth2.Credentials` and only populate that, but it'd have the same caveats as this property. I'm cool with either, and the `id_token` approach has some benefits.

google/oauth2/service_account.py
dhermes: I don't care either way.

google/oauth2/credentials.py
dhermes: I was on team "no new public properties", so neither really fits that.

Though given your description, it's a bit strange to tell people to access the things they want in two slightly different ways:

```python
>>> creds.access_token
>>> creds.refresh_grant_response['id_token']
```

so if you were going to add anything, make it the ID token?

jonparrott: SG. Will update this shortly and notify when it's ready to review again. :)

jonparrott:
@dhermes updated. PTAL.

google/oauth2/credentials.py
```diff
@@ -63,10 +65,12 @@ def __init__(self, token, refresh_token=None, token_uri=None,
         super(Credentials, self).__init__()
         self.token = token
         self._refresh_token = refresh_token
+        self._id_token = id_token
         self._scopes = scopes
         self._token_uri = token_uri
         self._client_id = client_id
         self._client_secret = client_secret
+        self._refresh_grant_response = None
```
dhermes: Needs deleted

google/oauth2/credentials.py
```diff
@@ -106,10 +121,12 @@ def with_scopes(self, scopes):
...
-            self._client_secret)
+        access_token, refresh_token, expiry, grant_response = (
+            _client.refresh_grant(
+                request, self._token_uri, self._refresh_token, self._client_id,
+                self._client_secret))
 
         self.token = access_token
         self.expiry = expiry
         self._refresh_token = refresh_token
+        self._id_token = grant_response.get('id_token', self._id_token)
```
dhermes: Can't something bad happen if the `refresh_token` and the `id_token` aren't from the same response / generation?

google/oauth2/credentials.py
jonparrott: Done.

google/oauth2/credentials.py
jonparrott: It can expire. Other than that I can't think of anything. I'm happy to change this though if you thing the behavior is surprising.

dhermes: Not surprising, just worried it is incorrect?

jonparrott: After thinking a bit, these are the only scenarios:

1. The credentials are constructed with a refresh token but no id token. The credentials have the openid scope. Refresh is called, the id token is populated.
2. The credentials are constructed with a refresh token and an id token. The credentials have the openid scope. Refresh is called, the id token is updated.
3. The credentials are constructed with a refresh token and an id token. The credentials don't have the openid scope (where did that id token come from, then?). Refresh is called, the id token is preserved.

Case 3 is so weird that I'm tempted to *not* preserve the id token. So I'm gonna switch it.

dhermes: :+1:

jonparrott: Done.

PR #153, opened 2017-04-12
Move run_pylint to gcp-devrel-py-tools

jonparrott:
Towards splitting the `additional_packages` into their own repositories.


jonparrott:
(lint-only change, will merge when Travis passes)

pylint.config.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2016 Google Inc.
```
dhermes: 2017?

jonparrott: It originated in 2016, so it stays 2016. :)

PR #158, opened 2017-05-05
Fix clock skew calculations

jonparrott:
Previously, the clock skew adjusted in the wrong direction. It would cause us to consider credentials whose expiration time had already passed according to the system clock to still be sent to the server. This is the opposite of what we wanted to happen. This fixes it so that we report that credentials are expired slightly before the system clock thinks they've expired.

### DIFF
--- a/google/auth/credentials.py
+++ b/google/auth/credentials.py
@@ -58,7 +58,7 @@ class Credentials(object):
         """
         # Err on the side of reporting expiration early so that we avoid
         # the 403-refresh-retry loop.
-        adjusted_now = _helpers.utcnow() - _helpers.CLOCK_SKEW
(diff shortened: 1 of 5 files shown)


google/auth/credentials.py
```diff
@@ -58,7 +58,7 @@ def expired(self):
         """
         # Err on the side of reporting expiration early so that we avoid
         # the 403-refresh-retry loop.
-        adjusted_now = _helpers.utcnow() - _helpers.CLOCK_SKEW
+        adjusted_now = _helpers.utcnow() + _helpers.CLOCK_SKEW
```
dhermes: This is **more** lenient than before, not less.

dhermes: Update: I am wrong

PR #168, opened 2017-06-29
Make testing style more consistent

jonparrott:
Several overall style changes:

1.  Avoid plain `Mock()` and `Mock(spec=thing)`., prefer `mock.create_autospec()`.
1.  Don't `mock.patch` without `autospec`.
1.  Don't give mock instances special names. Prefer `thing` over `thing_mock` and `mock_thing`.
1.  When using `mock.patch`, use the same name as the item being patched to refer to the mock.
    ```python
    with mock.patch('module.thing') as thing:
        ...
    ```
    and
    ```
    @mock.patch('module.thing')
    def test(thing):
        ...
    ```
1.  Test helper factories should follow the naming convention `make_thing()`.
1.  Use `ThingStub` when creating semi-functioning subclasses for testing purposes.

### DIFF
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -162,10 +162,10 @@ def update_query(url, params, remove=None):
     # Update the query parameters with the new parameters.
     query_params.update(params)
     # Remove any values specified in remove.
-    query_params = {
+    query_params = dict({
         key: value for key, value
         in six.iteritems(query_params)
-        if key not in remove}
+        if key not in remove})
     # Re-encoded the query string.
     new_query = urllib.parse.urlencode(query_params, doseq=True)
     # Unsplit the url.
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -162,10 +162,10 @@ def update_query(url, params, remove=None):
     # Update the query parameters with the new parameters.
     query_params.update(params)
     # Remove any values specified in remove.
-    query_params = dict({
+    query_params = {
         key: value for key, value
         in six.iteritems(query_params)
-        if key not in remove})
+        if key not in remove}
     # Re-encoded the query string.
     new_query = urllib.parse.urlencode(query_params, doseq=True)
     # Unsplit the url.
--- a/google/auth/transport/_http_client.py
+++ b/google/auth/transport/_http_client.py
@@ -34,8 +34,8 @@ class Response(transport.Response):
     """
     def __init__(self, response):
         self._status = response.status
-        self._headers = dict({
-            key.lower(): value for key, value in response.getheaders()})
+        self._headers = {
+            key.lower(): value for key, value in response.getheaders()}
         self._data = response.read()
 
     @property
--- a/google/auth/transport/_http_client.py
+++ b/google/auth/transport/_http_client.py
@@ -34,8 +34,8 @@ class Response(transport.Response):
     """
     def __init__(self, response):
         self._status = response.status
-        self._headers = {
-            key.lower(): value for key, value in response.getheaders()}
+        self._headers = dict({
+            key.lower(): value for key, value in response.getheaders()})
         self._data = response.read()
 
     @property
--- a/tests/compute_engine/test__metadata.py
+++ b/tests/compute_engine/test__metadata.py
@@ -24,6 +24,7 @@ from six.moves import reload_module
 from google.auth import _helpers
 from google.auth import environment_vars
 from google.auth import exceptions
+from google.auth import transport
 from google.auth.compute_engine import _metadata
 
 PATH = 'instance/service-accounts/default'
@@ -31,15 +32,15 @@ PATH = 'instance/service-accounts/default'
 
 @pytest.fixture
 def mock_request():
-    request_mock = mock.Mock()
+    mock_request = mock.create_autospec(transport.Request)
 
     def set_response(data, status=http_client.OK, headers=None):
-        response = mock.Mock()
+        response = mock.create_autospec(transport.Response, instance=True)
         response.status = status
         response.data = _helpers.to_bytes(data)
         response.headers = headers or {}
-        request_mock.return_value = response
-        return request_mock
+        mock_request.return_value = response
+        return mock_request
 
     yield set_response
 
--- a/tests/compute_engine/test__metadata.py
+++ b/tests/compute_engine/test__metadata.py
@@ -30,98 +30,95 @@ from google.auth.compute_engine import _metadata
 PATH = 'instance/service-accounts/default'
 
 
-@pytest.fixture
-def mock_request():
-    mock_request = mock.create_autospec(transport.Request)
+def make_request(data, status=http_client.OK, headers=None):
+    response = mock.create_autospec(transport.Response, instance=True)
+    response.status = status
+    response.data = _helpers.to_bytes(data)
+    response.headers = headers or {}
 
-    def set_response(data, status=http_client.OK, headers=None):
-        response = mock.create_autospec(transport.Response, instance=True)
-        response.status = status
-        response.data = _helpers.to_bytes(data)
-        response.headers = headers or {}
-        mock_request.return_value = response
-        return mock_request
+    request = mock.create_autospec(transport.Request)
+    request.return_value = response
 
-    yield set_response
+    return request
 
 
-def test_ping_success(mock_request):
-    request_mock = mock_request('', headers=_metadata._METADATA_HEADERS)
+def test_ping_success():
+    request = make_request('', headers=_metadata._METADATA_HEADERS)
 
-    assert _metadata.ping(request_mock)
+    assert _metadata.ping(request)
 
-    request_mock.assert_called_once_with(
+    request.assert_called_once_with(
         method='GET',
         url=_metadata._METADATA_IP_ROOT,
         headers=_metadata._METADATA_HEADERS,
         timeout=_metadata._METADATA_DEFAULT_TIMEOUT)
 
 
-def test_ping_failure_bad_flavor(mock_request):
-    request_mock = mock_request(
+def test_ping_failure_bad_flavor():
+    request = make_request(
         '', headers={_metadata._METADATA_FLAVOR_HEADER: 'meep'})
 
-    assert not _metadata.ping(request_mock)
+    assert not _metadata.ping(request)
 
 
-def test_ping_failure_connection_failed(mock_request):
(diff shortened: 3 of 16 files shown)


google/auth/_helpers.py
```diff
@@ -162,10 +162,10 @@ def update_query(url, params, remove=None):
     # Update the query parameters with the new parameters.
     query_params.update(params)
     # Remove any values specified in remove.
-    query_params = {
+    query_params = dict({
```
dhermes: Why?

google/auth/transport/_http_client.py
```diff
@@ -34,8 +34,8 @@ class Response(transport.Response):
     """
     def __init__(self, response):
         self._status = response.status
-        self._headers = {
-            key.lower(): value for key, value in response.getheaders()}
+        self._headers = dict({
```
dhermes: Moar why

tests/compute_engine/test_credentials.py
```diff
@@ -42,8 +43,8 @@ def test_default_state(self):
         'google.auth._helpers.utcnow',
         return_value=datetime.datetime.min + _helpers.CLOCK_SKEW)
     @mock.patch('google.auth.compute_engine._metadata.get')
```
dhermes: Is there a way to just say "the patched thing can ONLY be called"?

tests/oauth2/test_credentials.py
```diff
@@ -26,40 +27,40 @@ class TestCredentials(object):
     REFRESH_TOKEN = 'refresh_token'
     CLIENT_ID = 'client_id'
     CLIENT_SECRET = 'client_secret'
-    credentials = None
 
-    @pytest.fixture(autouse=True)
-    def credentials_fixture(self):
-        self.credentials = credentials.Credentials(
+    def make_credentials(self):
+        return credentials.Credentials(
```
dhermes: I recall we disagreed about this before

tests/compute_engine/test__metadata.py
```diff
@@ -24,191 +24,189 @@
...
+    response.headers = headers or {}
 
-    def set_response(data, status=http_client.OK, headers=None):
-        response = mock.Mock()
-        response.status = status
-        response.data = _helpers.to_bytes(data)
-        response.headers = headers or {}
-        request_mock.return_value = response
-        return request_mock
+    request = mock.create_autospec(transport.Request)
```
dhermes: How does one decide between `instance=True` or not?

tests/oauth2/test_service_account.py
```diff
@@ -41,21 +41,16 @@
...
 class TestCredentials(object):
     SERVICE_ACCOUNT_EMAIL = 'service-account@example.com'
     TOKEN_URI = 'https://example.com/oauth2/token'
-    credentials = None
 
-    @pytest.fixture(autouse=True)
-    def credentials_fixture(self, signer):
-        self.credentials = service_account.Credentials(
-            signer, self.SERVICE_ACCOUNT_EMAIL, self.TOKEN_URI)
+    def make_credentials(self):
```
dhermes: `@staticmethod`? (Here and above)

tests/test__default.py
```diff
@@ -179,28 +177,38 @@ def test__get_gcloud_sdk_credentials_project_id(
...
+    def get_application_id(self):
+        raise NotImplementedError()
 
 
 @pytest.fixture
-def app_identity_mock(monkeypatch):
+def app_identity(monkeypatch):
     """Mocks the app_identity module for google.auth.app_engine."""
-    app_identity_mock = mock.Mock()
+    app_identity = mock.create_autospec(_AppIdentityModule, instance=True)
```
dhermes: 1. Hacky Hackerson using `_AppIdentityModule` like this
1. local shadowing a global, recipe for ick

tests/test__default.py
```diff
@@ -179,28 +177,38 @@ def test__get_gcloud_sdk_credentials_project_id(
...
 
     credentials, project_id = _default._get_gcloud_sdk_credentials()
 
     assert credentials == mock.sentinel.credentials
     assert project_id is None
+    assert get_project_id.called
+
+
+class _AppIdentityModule(object):
+    """The interface of the App Idenity app engine module.
```
dhermes: You just hate PEP257 so much

tests/test__default.py
```diff
@@ -179,28 +177,38 @@ def test__get_gcloud_sdk_credentials_project_id(
...
+    See https://cloud.google.com/appengine/docs/standard/python/refdocs
```
dhermes: If you `\` continue the link, life is better

tests/test_app_engine.py
```diff
@@ -21,17 +21,35 @@
...
+        raise NotImplementedError()
+
+    def sign_blob(self, bytes_to_sign, deadline=None):
+        raise NotImplementedError()
+
+    def get_service_account_name(self, deadline=None):
+        raise NotImplementedError()
+
+    def get_access_token(self, scopes, service_account_id=None):
+        raise NotImplementedError()
```
dhermes: `NotImplementedError()` or `NotImplementedError`, that is the question

tests/transport/test_requests.py
```diff
@@ -84,44 +86,44 @@ def test_constructor(self):
         assert authed_session.credentials == mock.sentinel.credentials
 
     def test_request_no_refresh(self):
-        mock_credentials = mock.Mock(wraps=MockCredentials())
-        mock_response = make_response()
-        mock_adapter = MockAdapter([mock_response])
+        credentials = mock.Mock(wraps=CredentialsStub())
```
dhermes: Why pick `wraps` over `autospec`

google/auth/_helpers.py
jonparrott: Apparently `{key: value for item in generator}` is a generator expression as well. So, coverage was complaining about these never actually being executed.

tests/compute_engine/test_credentials.py
jonparrott: Yes, I just forgot to autospec this. You can also do `spec=['__call__']`.

tests/oauth2/test_credentials.py
jonparrott: Yep, are you happier now? :)

tests/compute_engine/test__metadata.py
jonparrott: From the [docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.create_autospec):

> If a class is used as a spec then the return value of the mock (the instance of the class) will have the same spec. You can use a class as the spec for an instance object by passing instance=True. The returned mock will only be callable if instances of the mock are callable.

More concretely, without `instance` you get a mock class, with instance you get a mock instance of that class.

google/auth/_helpers.py
dhermes: Empirically this is false.

```python
>>> vals = {a: a * a for a in [1, 2, 3]}
>>> vals
{1: 1, 2: 4, 3: 9}
>>> type(vals)
<class 'dict'>
>>>
```

and I got identical output in 2.7, 3.4, 3.5 and 3.6

tests/compute_engine/test_credentials.py
dhermes: :+1: w00t

tests/oauth2/test_credentials.py
dhermes: Bigly

tests/compute_engine/test__metadata.py
dhermes: Then ISTM you want `instance=True` here too?

tests/test__default.py
jonparrott: Do you have a better recommendation?

tests/test_app_engine.py
jonparrott: I have no strong feelings. :)

tests/transport/test_requests.py
jonparrott: Because in this case I want the partial functionality of the stub as well as the ability to assert that calls were made.

tests/test__default.py
jonparrott: How did I break PEP257 here?

tests/oauth2/test_service_account.py
jonparrott: Had to be a class method. :) Done.

tests/test__default.py
dhermes: A recommendation for 1 or 2?

For 2, I recommend naming the local `app_identity_module`. For 1, I think what you've done is A-OK (it's just a hack).

tests/compute_engine/test__metadata.py
jonparrott: So `Request` is actually tricky. It describes a Callable, not a class. So in this case, it's easier to use without instance, as `assert_x` methods pick up the right set of calls.

tests/test_app_engine.py
dhermes: I only have strong feelings.

```
$ git grep -l 'NotImplementedError()' | \
> xargs sed -i s/'NotImplementedError()'/'NotImplementedError'/g
```

tests/transport/test_requests.py
dhermes: > the ability to assert that calls were made

That's not always allowed when you use autospec?

tests/compute_engine/test__metadata.py
dhermes: :+1:

tests/transport/test_requests.py
jonparrott: That is allowed when you use autospec, but I also want the stub's behavior which is harder to achieve with a plain mock.

tests/test_app_engine.py
jonparrott: lol. I'd rather not. `explicit is better than implicit`.

tests/test__default.py
jonparrott: Done

tests/transport/test_requests.py
dhermes: Ahh gotcha

tests/test_app_engine.py
dhermes: OK

"I have no strong feelings"

tests/test__default.py
dhermes: You don't have an empty-line between the one-line description and the rest of the content

Issue #170, opened 2017-07-13
Verify OAuth2 Token - Timeout


DSpeckhals:
It doesn't seem to be possible to specify a timeout when calling `google.oauth2.id_token.verify_oauth2_token`. Because, as I understand it, if no timeout parameter is given when making a request with `requests`, the timeout is infinite, and the thread or process would hang forever.

Would there be any problem with being able to provide an optional `timeout` parameter in `verify_oauth2_token`, or perhaps the other facades around HTTP requests in this library? Or do you think that a timeout parameter could somehow be set when constructing the `transport.requests.Request` object, then used for any subsequent HTTP request?

I would be happy to help if this seems like the way forward.

jonparrott:
I'd be open to adding a `timeout` parameter to the constructor for `google.auth.transport.requests.Request`.

vchudnov-g:
Makes sense. The new `timeout` parameter should probably also have a default value if not specified.

jonparrott:
Empty default (deferring to requests default) seems fine.

DSpeckhals:
Closing because of the suggest in #171.

PR #171, opened 2017-07-13
Add optional HTTP request timeout

DSpeckhals:
HTTP requests did not have the ability to time out when used as a
parameter for calls like: `google.oauth2.id_token.verify_oauth2_token`.
This commit adds the ability in the different implementations of the
`google.auth.transport.requests.Request` class to specify a
`default_timeout` to be used when no `timeout` is used in the actual
request call (like `request.get('/url', timeout=2)`.

A specific timeout can still be specified as an arg when making a
request, which overrides this new `default_timeout`.

An example for performing a `verify_oauth2_token` call could look
something like this if a two second timeout was desired:

```python
request = google.auth.transport.requests.Request(default_timeout=2)
id_info = google.oauth2.id_token.verify_oauth2_token(id_token,
                                                     request,
                                                     client_id)
```

This addresses #170.

edit: `timeout` => `default_timeout`

### DIFF
--- a/google/auth/transport/requests.py
+++ b/google/auth/transport/requests.py
@@ -74,14 +74,18 @@ class Request(transport.Request):
     Args:
         session (requests.Session): An instance :class:`requests.Session` used
             to make HTTP requests. If not specified, a session will be created.
+        default_timeout (int): A number indicating the seconds for the default
+            HTTP request timeout. If not specified, the default timeout will be
+            indefinite.
 
     .. automethod:: __call__
     """
-    def __init__(self, session=None):
+    def __init__(self, session=None, default_timeout=None):
         if not session:
             session = requests.Session()
 
         self.session = session
+        self.default_timeout = default_timeout
 
     def __call__(self, url, method='GET', body=None, headers=None,
                  timeout=None, **kwargs):
@@ -95,7 +99,7 @@ class Request(transport.Request):
             headers (Mapping[str, str]): Request headers.
             timeout (Optional[int]): The number of seconds to wait for a
                 response from the server. If not specified or if None, the
-                requests default timeout will be used.
+                ``default_timeout`` property for this instance will be used.
             kwargs: Additional arguments passed through to the underlying
                 requests :meth:`~requests.Session.request` method.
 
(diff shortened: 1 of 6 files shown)


dhermes:
ISTM this PR is a (teeny) argument **against** `default_timeout`. The only places the default ever gets used are methods which allow an explicit `timeout` argument.

If typing `timeout=MY_VALUE` every time becomes cumbersome, then users are free to subclass?

Am I missing something subtle?

DSpeckhals:
@dhermes At the moment, `Request.__call__` is the only place that allows an optional timeout, but this abstracted by methods like `verify_oauth2_token`. Yes, you could allow a timeout to be specified in those abstractions, but then it would be additions to multiple API's, which seems a little bit of an overkill. Specifying it at `Request` init makes more sense to me.

Your suggestion about sub-classing makes sense, but that would require overriding `__call__`, thus making users who do this more prone to bugs when underlying functionality changes in the super method. If that's the consensus, though, we should add some documentation on how do accomplish the sub-classing.

jonparrott:
Another alternative is doing something like this:

```python
import functools

request_with_timeout = functools.partial(request, timeout=2)
id_token.verify_oauth2_token(id_token, request, client_id)
```

What do you think about that @dhermes @DSpeckhals?

DSpeckhals:
@jonparrott That's reasonable...maybe a little "hacky", but it works.

jonparrott:
@DSpeckhals I don't think it's hacky at all- this is one of many reasons why I try to reduce classes down to callables. I'm gonna close this PR for now, but if we get some more feedback that this is needed, I'll be happy to reconsider.

DSpeckhals:
Thanks for the direction on this. I like the solution now that you explain the "why."

wontonst:
from my perspective it looks pretty hacky to have to rely on functools to do something a simple as adding a timeout to a commonly used call.

I suspect most developers are directly calling `verify_oauth2_token` and having to dig into the codebase to understand how to add a timeout is not the best developer experience. what I ended up going with is

```
from google.auth.transport import requests as google_requests
class GoogleRequestWithTimeout(google_requests.Request):
    def __call__(self, *args, **kwargs):
        return super().__call__(*args, timeout=5, **kwargs)
```

PR #179, opened 2017-08-01
Fix retrieval of default project ID on Windows

PicardParis:
- Fix for shell command call "subprocess.check_output(_CLOUD_SDK_CONFIG_COMMAND...)".
- Previously, it would trigger an error "WindowsError: [Error 2] The system cannot find the file specified". A workaround was to specify the env variable GCLOUD_PROJECT.
- With this patch, "gcloud" default project is successfully retrieved.


jonparrott:
Thanks, @PicardParis. I made a few changes and added a test. :)

PR #189, opened 2017-08-11
Split crypt into a package to allow alternative implementations

jonparrott:


google/auth/crypt/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+            to use to check the signature.
+
+    Returns:
+        bool: True if the signature is valid, otherwise False.
+    """
+    if isinstance(certs, (six.text_type, six.binary_type)):
+        certs = [certs]
+
+    for cert in certs:
+        verifier = rsa.RSAVerifier.from_string(cert)
```
dhermes: Don't you want to use this module's `RSAVerifier` here (rather than `rsa.RSAVerifier`)?

jonparrott: They're the same (and will continue to be the same when you add the cryptography based one)

dhermes: SGTM

PR #195, opened 2017-09-11
Move read-only methods of 'Scoped' into new interface, 'ReadOnlyScoped'.

### LINKED ISSUE #194: GCE Credentials should not derive from Scoped
Because `with_scopes` raises `NotImplementedError`, the class cannot be used polymorphically where instances of `Scoped` are expected (see https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3933).  For instance, in C++,`google.auth.compute_engine.credentials.Credentials` wouldn't even be instantiable, because `with_scopes` is a pure virtual method cognate.

If there is a need for *part* of the interface to be expressed (e.g., the `scopes`, `requires_scopes`, and `has_scopes` methods), that should be factored into another class (maybed `ReadonlyScoped`?), with the current `Scoped` class extending it.

tseaver:
Not all subclasses of 'Scoped' can sanely implement `with_scopes` (e.g, on GCE the scopes are hard-wired in when creating the GCE node).

Make `Scoped` derive from `ReadOnlyScoped`, adding the `with_scopes` method.

Make GCE's `Credentials` class derive from `ReadOnlyScoped`.

Closes #194.

### DIFF
--- a/google/auth/compute_engine/credentials.py
+++ b/google/auth/compute_engine/credentials.py
@@ -24,7 +24,7 @@ from google.auth import exceptions
 from google.auth.compute_engine import _metadata
 
 
-class Credentials(credentials.ReadonlyScoped, credentials.Credentials):
+class Credentials(credentials.ReadOnnlyScoped, credentials.Credentials):
     """Compute Engine Credentials.
 
     These credentials use the Google Compute Engine metadata server to obtain
--- a/google/auth/compute_engine/credentials.py
+++ b/google/auth/compute_engine/credentials.py
@@ -24,7 +24,7 @@ from google.auth import exceptions
 from google.auth.compute_engine import _metadata
 
 
-class Credentials(credentials.Scoped, credentials.Credentials):
+class Credentials(credentials.ReadonlyScoped, credentials.Credentials):
     """Compute Engine Credentials.
 
     These credentials use the Google Compute Engine metadata server to obtain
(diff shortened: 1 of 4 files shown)


tseaver:
@jonparrott PTAL.

jonparrott:
This seems pretty reasonable to me, but I'd like @dhermes to take a quick look as well.

jonparrott:
@tseaver I want to double check something - this does *not* break the current interface of the library, correct? Other than GCE credentials now being "scopeless"?

> but I'm scratching my head a bit about why this is needed.

Yeah - could we get some context? I understand the computer sciency parts, but in practice how does this make the calling code different?

lukesneeringer:
> Yeah - could we get some context? I understand the computer sciency parts, but in practice how does this make the calling code different?

Context: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3933

tseaver:
Context: Having GCE's credentials derive from `Scoped` is a promise that it has a meaningful `with_scopes`, but in fact it raises `NotImplementedError`, which means that clients can't rely on it.

jonparrott:
> If there are no objections, I will make the edit myself out of respect for others' time.

go for it

tseaver:
@lukesneeringer 07d9c3b updates spelling per your request.

lukesneeringer:
You beat me to it. Thanks. :-)

PR #197, opened 2017-09-12
Fix App Engine's expiration calculation

jonparrott:
Resolves https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3937


dhermes:
LGTM, though WDYT about adding a script here that will deploy a new GAE app and run unit tests there (with results available via `curl`, e.g.)

jonparrott:
> LGTM, though WDYT about adding a script here that will deploy a new GAE app and run unit tests there (with results available via curl, e.g.)

We have one in system_tests, it just doesn't do any introspection on the token or expiry - it just makes an authenticated request and verifies it works.

PR #200, opened 2017-09-20
Fix OAuth2 credentials to inherit from ReadOnlyScoped instead of Scoped.

mbassiouny:
This is an issue when using spanner client APIs. https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4010. Having OAuth2 credentials inherit from ReadOnlyScoped instead of Scoped will fix that.


jonparrott:
Thanks, @mbassiouny!

PR #210, opened 2017-11-08
fix comment, seconds not sections

albertjan:
I assume this is what it was supposed to be 😄


jonparrott:
Thanks, @albertjan!

albertjan:
You're welcome! And thank you! 😄

PR #212, opened 2017-11-09
Using `six.raise_from` wherever possible.

dhermes:
I'm not entirely sure how worthwhile this is, I just discovered the implementation doesn't do anything helpful in Python 2 (e.g. print the traceback from the original exc):

```python
# Python < 3.2
def raise_from(value, from_value):
    raise value
# Python == 3.2
def raise_from(value, from_value):
    try:
        if from_value is None:
            raise value
        raise value from from_value
    finally:
        value = None
# Python > 3.2
def raise_from(value, from_value):
    try:
        raise value from from_value
    finally:
        value = None
```

I was "inspired" to do this by #211.

### DIFF
--- a/tests/test__default.py
+++ b/tests/test__default.py
@@ -85,6 +85,7 @@ def test__load_credentials_from_file_authorized_user_bad_format(tmpdir):
         _default._load_credentials_from_file(str(filename))
 
     assert excinfo.match(r'Failed to load authorized user')
+    assert excinfo.match(r'missing fields')
 
 
 def test__load_credentials_from_file_service_account():
@@ -102,6 +103,7 @@ def test__load_credentials_from_file_service_account_bad_format(tmpdir):
         _default._load_credentials_from_file(str(filename))
 
     assert excinfo.match(r'Failed to load service account')
+    assert excinfo.match(r'missing fields')
 
 
 @mock.patch.dict(os.environ, {}, clear=True)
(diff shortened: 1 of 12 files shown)


tests/test__default.py
```diff
@@ -85,7 +85,6 @@ def test__load_credentials_from_file_authorized_user_bad_format(tmpdir):
         _default._load_credentials_from_file(str(filename))
 
     assert excinfo.match(r'Failed to load authorized user')
-    assert excinfo.match(r'missing fields')
```
dhermes: @jonparrott Before merging, are you OK with dropping this? It disappears because the "raised from" is no longer **inside** the exception caught by this test. I suppose if we wanted to be fancy we could check in `stderr` for the traceback, which would have "missing fields" in it.

jonparrott: As long as the user still sees a helpful message, it's fine.

dhermes: Yeah that will be **less** true on Python 2 in cases where I changed

```python
raise FooExc(caught_exc)
```

to

```python
six.raise_from(FooExc(), caught_exc)
```

so we probably shouldn't merge this "as-is".

jonparrott: Yeah, in some of these cases I was okay with shadowing the original exception, but changing things up a bit to include the original exception in the new exception's message in some way while also using raise_from would be fine.

dhermes:
@jonparrott I updated so that the wrapped exceptions were still wrapped.

PR #205, opened 2017-10-16
Retry OAuth 2.0 token refresh requests at network errors on layers lower than HTTP

somay:
`https://accounts.google.com/` does NOT send TCP FIN packets nor RST when it closes connections (see the tcpdump's output below).
This makes the requests library (precisely the underlying urllib3 library) mistakenly reuse an actually-closed connection from its connection pools for OAuth 2.0 token refresh HTTP requests, resulting in long time blocking or the `104: Connection reset by peer` error.

My first commit handles `Connection reset by peer` using requests' adapters.

The second commit is for the blocking behaviours and sets the `timeout` parameter to a HTTP request function to make them raise `requests.Timeout` exception and retry refresh requests when catches it.
Simply setting the `timeout` parameter is not enough, because the users of this library cannot determine whether the exception is from the token refresh HTTP requests (which I think can be safely retried) or from their own ones (which might not be idempotent nor able to be retried).
This commit might not be good though because it changes the default timeout behaviour of this library with the HTTP libraries other than requests.

I also encourage you to stop using `requests.Session` (and thus the urllib3's connection pool) as a better enhancement since it is nonsense to pool connections for OAuth token refresh which occurs as infrequently as every one hour in most cases.

### How to Reproduce

The code to reproduce this behaviour is below (it does not use the google-auth library but rather emulates its internal behaviour of version 1.1.1):

```python
#!/usr/bin/env python
import requests
import logging
import sys
from time import sleep

url = sys.argv[1]

logging.basicConfig()
logging.getLogger().setLevel(logging.DEBUG)
requests_log = logging.getLogger("urllib3")
requests_log.setLevel(logging.DEBUG)

s = requests.Session()
s.get(url, allow_redirects=False)
print("sleep for 200 sec.")
sleep(200)
print("try GET now")
s.get(url, allow_redirects=False)  # Block forever!
```

console output:

```console
$ ./request_twice.py https://accounts.google.com/
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): accounts.google.com
DEBUG:urllib3.connectionpool:https://accounts.google.com:443 "GET / HTTP/1.1" 302 272
sleep for 200 sec.
try GET now
```

### Environments

Whether blocking or raising the error depends on clients' environments.
I have not investigate the reason.

I observed blocking behaviour on my Linux laptop:

```console
$ uname -a
[...]

PR #220, opened 2017-11-21
google.auth.transport.requests: make credential refresh robust

somay:
Setting timeout only to the requests library. Backward compatible.
Much simpler than my previous PR (#205) because I've made my misuderstanding clear.

I said:

> the users of this library cannot determine whether the exception is from the token refresh HTTP requests (which I think can be safely retried) or from their own ones (which might not be idempotent nor able to be retried).

But it is not true. `refresh()` raises `google.auth.exceptions.TransportError` wrapping a requests' exception.
So users can write retry logic like

```python
try:
    request()
except google.auth.exceptions.TransportError as e:
    if isinstance(e.args[0], requests.exceptions.ReadTimeout):
        retry()
```

### DIFF
--- a/google/auth/transport/requests.py
+++ b/google/auth/transport/requests.py
@@ -16,6 +16,7 @@
 
 from __future__ import absolute_import
 
+import functools
 import logging
 
 try:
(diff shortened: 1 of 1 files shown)


jonparrott:
Thanks, @somay!

PR #226, opened 2017-12-01
API for parsing google authorized_user json files

hiranya911:
This is supported in other languages (e.g. [Java](http://google.github.io/google-auth-library-java/releases/0.6.0/apidocs/com/google/auth/oauth2/UserCredentials.html)). I believe the Python library can also benefit from it.

Typical Usage:
```
from google.oauth2 import credentials

cred = credentials.Credentials.from_authorized_user_file('path/to/file.json')

# or...
scopes = [...]
scoped_cred = credentials.Credentials.from_authorized_user_file('path/to/file.json', scopes)
```

### DIFF
--- a/google/oauth2/credentials.py
+++ b/google/oauth2/credentials.py
@@ -36,7 +36,7 @@ from google.auth import credentials
 from google.oauth2 import _client
 
 
-class Credentials(credentials.Scoped, credentials.Credentials):
+class Credentials(credentials.ReadOnlyScoped, credentials.Credentials):
     """Credentials using OAuth 2.0 access and refresh tokens."""
 
     def __init__(self, token, refresh_token=None, id_token=None,
@@ -108,15 +108,6 @@ class Credentials(credentials.Scoped, credentials.Credentials):
         """False: OAuth 2.0 credentials have their scopes set when
         the initial token is requested and can not be changed."""
         return False
-
-    def with_scopes(self, scopes):
-        """Unavailable, OAuth 2.0 credentials can not be re-scoped.
-
-        OAuth 2.0 credentials have their scopes set when the initial token is
-        requested and can not be changed.
-        """
-        raise NotImplementedError(
-            'OAuth 2.0 Credentials can not modify their scopes.')
 
     @_helpers.copy_docstring(credentials.Credentials)
     def refresh(self, request):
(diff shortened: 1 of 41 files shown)


jonparrott:
I have mixed feels about this, as this isn't a publicly documented/standardized format and is *very* specific to the Cloud SDK. 

@vchudnov-g what do you think?

hiranya911:
This could benefit the Firebase Python SDKs immediately, and perhaps other gcloud SDK related tooling as well. But I do understand the concern, and don't feel strongly either way.

jonparrott:
It's just we're kind of discouraging re-using gcloud's auth, and this just keep re-enforcing that behavior.

google/oauth2/credentials.py
```diff
@@ -120,3 +126,55 @@ def refresh(self, request):
...
+        Raises:
+            ValueError: If the info is not in the expected format.
+        """
+        try:
+            file_type = info['type']
+            client_id = info['client_id']
+            client_secret = info['client_secret']
+            refresh_token = info['refresh_token']
+        except KeyError as error:
+            raise ValueError('Failed to initialize am authorized_user credential. '
```
dhermes: s/am authorized/an authorized

google/oauth2/credentials.py
```diff
@@ -120,3 +126,55 @@ def refresh(self, request):
...
+                credentials.
+
+        Returns:
+            google.oauth2.credentials.Credentials: The constructed
+                credentials.
+
+        Raises:
+            ValueError: If the info is not in the expected format.
+        """
+        try:
```
jonparrott: We already have all of this logic [here](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/_cloud_sdk.py#L85). Can you think of de-duplicating it?

hiranya911: Done

google/oauth2/credentials.py
hiranya911: Done

hiranya911:
@jonparrott Addressed the comments, and fixed affected tests and lint errors.

jonparrott:
Thanks for doing this @hiranya911. I still feel weird about it, but oh well.

PR #227, opened 2017-12-07
Check for project_id envvar before warning

JacobHayes:
Resolves GoogleCloudPlatform/google-auth-library-python#221

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -147,7 +147,7 @@ def _get_explicit_environ_credentials():
         credentials, project_id = _load_credentials_from_file(
             os.environ[environment_vars.CREDENTIALS])
 
-        if not project_id:
+        if not os.environ.get(environment_vars.PROJECT):
             _LOGGER.warning(
                 'No project ID could be determined from the credentials at %s '
                 'Consider setting the %s environment variable',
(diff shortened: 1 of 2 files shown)


jonparrott:
This kind of indicates to me that we should remove the individual warnings and just warn in default()?

JacobHayes:
Agreed. The messages were _just_ different enough that I didn't tackle that originally. Would you prefer to make the message the same (ie: not reporting where the `credentials` came from) or something like adding a name to [`checkers`](https://github.com/JacobHayes/google-auth-library-python/blob/f42d53519b6ecf2a4ce235f3c04e34e54b6fa9da/google/auth/_default.py#L281) to fill into the message? Ex:

```python
    checkers = {  # or tuple of two-tuples
        os.environ.get(environment_vars.CREDENTIALS): _get_explicit_environ_credentials,
        'Cloud SDK Credentials': _get_gcloud_sdk_credentials,
        'App Engine metadata service': _get_gae_credentials,
        'Compute Engine metadata service': lambda: _get_gce_credentials(request),
    }
```

jonparrott:
> prefer to make the message the same (ie: not reporting where the credentials came from)

That would be fine. :)

JacobHayes:
Any ideas on fixing the coverage test?

jonparrott:
Write a test that mocks all the checkers to return no project?

jonparrott:
Thanks for doing this, @JacobHayes!

JacobHayes:
Absolutely; thanks for the guidance!

PR #230, opened 2017-12-15
Drop direct pyasn1 dependency

jonparrott:
Instead, rely on pyasn1-modules to pull the appropriate version.


jonparrott:
Merging despite lint, which I guess I should just turn off.

dhermes:
> which I guess I should just turn off

Boo

jonparrott:
Just our franken-pylint.

Issue #130, opened 2017-03-14
Consider implementing OpenID Connect ID Token-based credentials for service accounts


jonparrott:
This would satisfy the Identity-aware Proxy use case as well as a few other niche cases that have come up.

Context:
https://gist.github.com/jonparrott/cffca2fa7881e03fbe6ff7c25773c9cf
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/iap/make_iap_request.py#L121

/cc @matthewg

jonparrott:
I have no idea what to name this thing:

* `google.oauth2.oidc.Credentials` could work, but since this are specific to service accounts that seems weird.
* `google.oauth2.service_account.IDTokenCredentials`
* `google.oauth2.service_account.OIDCCredentials` makes me hurt.
* `google.oauth2.open_id_connect.ServiceAccountCredentials` ew.
* `google.oauth2.odic.service_acccount.Credentials`?

IDK. Thoughts @dhermes @lukesneeringer @matthewg?

dhermes:
`google.oauth2.service_account.IDTokenCredentials` :+1:

ensonic:
+1. This is right now not straightforward at all and is required for
https://cloud.google.com/endpoints/docs/openapi/service-to-service-auth#enabling_a_calling_service_to_make_authenticated_calls

FranOis:
@jonparrott I'm not sure I understand all the consequences but I was playing around thinking open id would work and it seems like the only thing that needs to change is change from `access_token` to `id_token` in https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/oauth2/_client.py#L147

Like I said, no idea what is the other uses of that class, but I did something very simple to get there. The `credentials.json` is for a service account in my project that is `roles/servicemanagement.serviceConsumer`

```
from google.oauth2 import service_account
from google.auth.transport.requests import AuthorizedSession

credentials = service_account.Credentials.from_service_account_file(
    'credentials.json')

scoped_credentials = credentials.with_scopes(
    ['http://URL_FOR_MY_CLOUD_ENDPOINTS'])

authed_session = AuthorizedSession(scoped_credentials)

response = authed_session.get('http://URL_FOR_MY_CLOUD_ENDPOINTS/hello')

print(response.json())
```

jonparrott:
@FranOis that's basically the gist of it. We'd just want a new credentials class to avoid messing with any of the existing ones.

[...]

PR #234, opened 2018-01-13
Add Credentials implementation supplying an ID token.

kryzthov:
@jonparrott Here's a primitive implementation for #130.
This adds everything in `google.auth.id_token` as I don't know any better.
Could you advise on how to structure this better?
Thanks!


jonparrott:
I changed a significant portion of this so some else needs to review. @dhermes?

google/oauth2/_client.py
```diff
@@ -155,6 +156,49 @@ def jwt_grant(request, token_uri, assertion):
...
+            HTTP requests.
+        token_uri (str): The OAuth 2.0 authorizations server's token endpoint
+            URI.
+        assertion (str): JWT token signed by a service account. The assertion
+            must include a ``target_audience`` claim.
+    Returns:
+        Tuple[str, Optional[datetime], Mapping[str, str]]:
+            The (encoded) Open ID Connect ID Token, expiration, and additional
+            data returned by the endpoint.
+    Raises:
```
dhermes: Should be an empty line before `Raises:`?

google/oauth2/service_account.py
```diff
@@ -336,3 +336,207 @@ def signer(self):
...
+            :meth:`from_service_account_info` are used instead of calling the
+            constructor directly.
+        """
+        super(IDTokenCredentials, self).__init__()
+        self._signer = signer
+        self._service_account_email = service_account_email
+        self._token_uri = token_uri
+        self._target_audience = target_audience
+
+        if additional_claims is not None:
```
dhermes: My "take-it-or-leave-it" review suggestion here is always to drop the `not` from your conditional check and just swap. The why is because the `not` is sort of doing an extra "computation" (it isn't actually if you `dis.dis()` the code).

google/oauth2/service_account.py
```diff
@@ -336,3 +336,207 @@ def signer(self):
...
+`
+    The credentials are considered immutable. If you want to modify the scopes
+    or the subject used for delegation, use :meth:`with_scopes` or
+    :meth:`with_subject`::
+
+        scoped_credentials = credentials.with_scopes(['email'])
+        delegated_credentials = credentials.with_subject(subject)
+
+    """
+    def __init__(self, signer, service_account_email, token_uri,
```
dhermes: It seems like this constructor (and some of the factories / factory helpers below) is **very** similar to existing code. Is this a problem @jonparrott?

google/oauth2/service_account.py
```diff
@@ -336,3 +336,207 @@ def signer(self):
...
+
+        Args:
+            target_audience (str): The intended audience for these credentials,
+            used when requesting the ID Token.
+
+        Returns:
+            google.auth.service_account.IDTokenCredentials: A new credentials
+                instance.
+        """
+        return IDTokenCredentials(
```
dhermes: This isn't friendly to subclasses, any reason not to use `self.__class__`?

google/oauth2/_client.py
```diff
@@ -155,6 +156,49 @@ def jwt_grant(request, token_uri, assertion):
     return access_token, expiry, response_data
 
 
+def id_token_jwt_grant(request, token_uri, assertion):
+    """Implements the JWT Profile for OAuth 2.0 Authorization Grants, but
+    requests an OpenID Connect ID Token instead of a access token.
```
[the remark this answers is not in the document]
jonparrott: Fixed `an`, leaving summary as-is because I don't want it to be identical to the `jwt_grant` function.

google/oauth2/_client.py
```diff
@@ -155,6 +156,49 @@ def jwt_grant(request, token_uri, assertion):
...
+
+    This is a variant on the standard JWT Profile that is currently unique
+    to Google. This was added for the benefit of authenticating to services
+    that require ID Tokens instead of access tokens or JWT bearer tokens.
+
+    Args:
+        request (google.auth.transport.Request): A callable used to make
+            HTTP requests.
+        token_uri (str): The OAuth 2.0 authorizations server's token endpoint
```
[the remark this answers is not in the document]
jonparrott: done.

google/oauth2/_client.py
```diff
@@ -155,6 +156,49 @@ def jwt_grant(request, token_uri, assertion):
...
+            URI.
+        assertion (str): JWT token signed by a service account. The assertion
+            must include a ``target_audience`` claim.
```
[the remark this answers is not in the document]
jonparrott: Clarified.

google/oauth2/_client.py
```diff
@@ -155,6 +156,49 @@ def jwt_grant(request, token_uri, assertion):
...
+    Returns:
```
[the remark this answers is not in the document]
jonparrott: Yah, fixed.

google/oauth2/_client.py
jonparrott: Fixed.

google/oauth2/service_account.py
jonparrott: I really struggled with this. I considered quite a few options here:

* making a common class for `service_account.Credentials` and `service_account.IDTokenCredentials`.
* making `IDTokenCredentials` a sub-class of `service_account.Credentials`, but it doesn't work because `IDTokenCredentials are actually scopeless.
* making `IDTokenCredentials` a sub-class of `jwt.Credentials` (as in the original PR, but it doesn't work because certain methods don't make sense and there was one method (which I can't remember) that would break the substitution principle - basically, `IDTokenCredentials` could not reasonably ever take the place of `jwt.Credentials`.
* Just make `IDTokenCredentials` its own class with similar methods.

The only viable options were 1 and 4, and I ended up with four, but I'm happy to reconsider (1) or some alternative I didn't think of.

google/oauth2/service_account.py
jonparrott: *leaves it*

google/oauth2/service_account.py
jonparrott: Oof, this is actually in several classes, will fix throughout.

google/oauth2/service_account.py
jonparrott: I'm going to merge this as-is. If this becomes onerous we can refactor out without breaking the interface.

PR #235, opened 2018-01-18
Includes Examples

Barathwaja:
Felt it lacking some examples, so added few samples on **Google Spreadsheet via Google Auth Library** using Service Account.


jonparrott:
Thanks for putting this together, @Barathwaja, but I do not feel like this is the right place to put this. If anywhere, it would go in `google-api-python-client`.

Barathwaja:
oh ... 👍  okay mate.. !!!

Barathwaja:
hmmm ok

PR #268, opened 2018-06-07
Fix some typos in test_urllib3.py

felixonmars:


tseaver:
@felixonmars Thanks for the patch!

PR #277, opened 2018-07-11
Making the dependency on app_engine unnecessary.

mfomitchev:
For code running in GCE, not GAE it doesn't make sense to force the
dependency on app_engine just to use GCE auth.

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -172,13 +172,13 @@ def _get_explicit_environ_credentials():
 
 def _get_gae_credentials():
     """Gets Google App Engine App Identity credentials and project ID."""
-    from google.auth import app_engine
-
     try:
+        from google.auth import app_engine
+
         credentials = app_engine.Credentials()
         project_id = app_engine.get_project_id()
         return credentials, project_id
-    except EnvironmentError:
+    except (ImportError, EnvironmentError):
         return None, None
 
 


mfomitchev:
Thanks for the quick review!
I am not sure how to write a unit case for this case.. I guess I'd need to create a separate test configuration in tox.ini where app_engine.py wasn't available? That seems messy.. Is there a simple way to do it?
I'd rather not invest too much time into this, particularly given this is a deprecated library.

tseaver:
@mfomitchev One strategy would be to move the import out to module scope, and use success / failure to set a variable, e.g.:

```python
try:
    from google.auth import app_engine
except ImportError:
    _HAVE_GAE_SUPPORT = False
else:
    _HAVE_GAE_SUPPORT = True
```

Then, `_get_gae_credentials` could check that variable:

```python
def _get_gae_credentials():
    if not _HAVE_GAE_SUPPORT:
        return None, None
    try:
        credentials = app_engine.Credentials()
    except EnvironmentError:
        return None, None
    project_id = app_engine.get_project_id()
    return credentials, project_id
```

Then the test could use `mock.patch` to replace the `_HAVE_GAE_SUPPORT` variable.

mfomitchev:
I don't think that would test the issue I am fixing though. The issue is that app_engine module may not be available, the fix is to catch the ImportError. The test should then verify that the ImportError is caught. Mocking _HAVE_GAE_SUPPORT doesn't test that the ImportError is caught. If we remove the try/catch around "from google.auth import app_engine" such test would still pass.

mfomitchev:
(Re-posting)
@tseaver I don't think that would test the issue I am fixing though. The issue is that app_engine module may not be available, the fix is to catch the ImportError. The test should then verify that the ImportError is caught. Mocking _HAVE_GAE_SUPPORT doesn't test that the ImportError is caught. If we remove the try/catch around "from google.auth import app_engine" such test would still pass.

tseaver:
@mfomitchev The module being imported is `google.auth.app_engine`, i.e. the [one in this package](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/app_engine.py).  It does *not* raise an ImportError if the `google.appengine.api` module is missing -- it just sets `app_identity` to `None`, which is what triggers raising EnvironmentError in `app_engine.get_project_id`.

Thanks for you effort, but can't see that your patch solves the problem you are intending to address.

mfomitchev:
@tseaver app_engine.py is in a separate BUILD target in google3 though. I guess I could just add a dependency on it to my code, but I'd rather not considering my code has nothing to do with app engine.

tseaver:
@mfomitchev `google-auth` does not split itself up that way.  The issue lies in whatever code chooses to split the `app_engine.py` module maintained *here* out from this package.

tseaver:
@mfomitchev I don't have access to Google-internal libraries:  if `google3` is Google's big, internal mono-repo, then this is a bug in *its* fork / packaging of this repo, which is the authoritative source for the publicly-released `google-auth` Python package.

mfomitchev:
@tseaver Ah I see, okay, fair enough. Thanks for elaborating.

theacodes:
The `app_engine` module is already well-insulated against the lack of the App Engine SDK being around - it just does so when you try to use the credentials instead of at import time. I'm with @tseaver here, if you depend on the `google.auth.default` module you sort of have to depend on *all* of the other modules (in this package) that default needs to support Application Default Credentials.

mfomitchev:
@theacodes If so, then perhaps //third_party/py/google/cloud:core should include //third_party/py/google/auth:app_engine in its deps (should probably be added to COMMON_DEPS). Otherwise all clients need to add that dep separately, which seems sub-optimal.

theacodes:
Maybe, I'm not a build expert though. Probably work mentioning to @nathanielmanistaatgoogle.

PR #284, opened 2018-08-13
Remove punctuation which becomes part of the url

sorensenjs:
In colab and likely other applications, the period is incorrectly bound into the url, which is then invalid and produces a 404

DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or
explicitly create credential and re-run the application. For more
information, please see
https://developers.google.com/accounts/docs/application-default-credentials.


theacodes:
Thank you, @sorensenjs!

PR #298, opened 2018-10-01
Enable static type checking with pytype

Solumin:
This PR adds pytype as a tox environment.

Type checking is only enabled for Python 3.6, due to limitations with the type information available for the libraries in 2.7.


Solumin:
@theacodes Do you have time to take a look at this?

Issue #275, opened 2018-06-29
Enable caller provide customized request session for authentication request.


shileiyu:
Hardcode a new session dedicated for authentication request make the user hard to customize the requests session object.
 
https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/edfe24602051969e32917e82bcedd2bace43e260/google/auth/transport/requests.py#L167

brianbao:
@theacodes I ran into this issue when working with a proxy and can send a fix, but I wasn't sure how to provide the correct fix.

We could stay consistent with what `urllib3.py` does and allow a Session object to be passed through to `__init__` and use that as `self._auth_request`. We can then just default to the original base `Session` object.

As an aside, the `requests.Session` object doesn't allow `kwargs`, so I'm not sure if we want to allow that or if we do, then we can also consider propagating `kwargs` into `self` and `auth_request_session`.

PR #306, opened 2018-11-12
Allow user to pass Session object to AuthorizedSession

brianbao:
Addresses #275. 

I also removed `**kwargs` into the `AuthorizedSession` object, as the `Session` object does not allow `**kwargs` (https://github.com/requests/requests/issues/1034).

If you'd rather keep `kwargs` we can initialize the parameters in `__init__`, something like:
```
for k, v in kwargs.items():
    setattr(self, k, v)
```


google/auth/transport/requests.py
```diff
@@ -150,27 +150,31 @@ class AuthorizedSession(requests.Session):
             refresh the credentials and retry the request.
         refresh_timeout (Optional[int]): The timeout value in seconds for
             credential refresh HTTP requests.
-        kwargs: Additional arguments passed to the :class:`requests.Session`
-            constructor.
+        auth_request_session (:class:`requests.Session`): (Optional)
+            HTTP object to make requests.
```
theacodes: Instead of `HTTP object to make requests.` can you update this to `A Requests session used when refreshing credentials.`?

Also would it make sense for this to just accept a `google.auth.transport.Request` object instead?

brianbao: :+1:

Hmm, from a user perspective it would be easier for me to just pass in the `Session` object. If it accepts a `Request` object, it'd just add an additional step of wrapping the `Session` in a `Request` object. I think both options are sensible though, so I'll defer to you to decide.

theacodes: Looking over the other implementations I think it makes the most sense to make this a Request object.

brianbao: Sounds good!

brianbao:
Hey @theacodes sorry for the bump, is there anything else I should add to this PR to get it merged?

PR #307, opened 2018-11-13
Link all the PR's in CHANGELOG

adamchainz:
The 1.6.0 release included Markdown link syntax for the PR's, which doesn't work in reStructuredText. Whilst fixing that I converted all the reference numbers to be links.


theacodes:
Thank you so much for doing this! 🎇

PR #309, opened 2018-11-27
Add downscoping to ouath2 credentials

eugenewf:


eugenewf:
fixed cla (theoretically)

eugenewf:
tests pass on python 2.7, 3.5, and 3.6.  I'm having trouble installing 3.4, but will update tomorrow.

docs/reference/google.auth.app_engine.rst
```diff
@@ -1,5 +1,5 @@
-google.auth.app_engine module
-=============================
+google.auth.app\_engine module
+==============================
```
tseaver: That is not valid Sphinx markup.  Please revert.  Note the [current, correctly-rendered page](https://google-auth.readthedocs.io/en/latest/reference/google.auth.app_engine.html).

docs/reference/google.auth.compute_engine.credentials.rst
```diff
@@ -1,5 +1,5 @@
-google.auth.compute_engine.credentials module
-=============================================
+google.auth.compute\_engine.credentials module
+==============================================
```
tseaver: Likewise here.  Note the [current, correctly-rendered page](https://google-auth.readthedocs.io/en/latest/reference/google.auth.compute_engine.html).

docs/reference/google.auth.compute_engine.rst
```diff
@@ -1,5 +1,5 @@
-google.auth.compute_engine package
-==================================
+google.auth.compute\_engine package
+===================================
```
tseaver: Likewise.

docs/reference/google.auth.crypt.rst
```diff
@@ -1,7 +1,16 @@
-google.auth.crypt module
-========================
+google.auth.crypt package
+=========================
```
tseaver: Good spot!

docs/reference/google.auth.environment_vars.rst
```diff
@@ -1,5 +1,5 @@
-google.auth.environment_vars module
-===================================
+google.auth.environment\_vars module
+====================================
```
tseaver: Not needed.  Please revert.

docs/reference/google.oauth2.id_token.rst
```diff
@@ -1,5 +1,5 @@
-google.oauth2.id_token module
-=============================
+google.oauth2.id\_token module
+==============================
```
tseaver: Please revert.

docs/reference/google.oauth2.service_account.rst
```diff
@@ -1,5 +1,5 @@
-google.oauth2.service_account module
-====================================
+google.oauth2.service\_account module
+=====================================
```
tseaver: Please revert.

google/oauth2/credentials.py
```diff
@@ -71,6 +71,10 @@ def __init__(self, token, refresh_token=None, id_token=None,
                 to obtain authorization. This is a purely informative parameter
                 that can be used by :meth:`has_scopes`. OAuth 2.0 credentials
                 can not request additional scopes after authorization.
+            downscope (bool): Whether to reduce the requested scopes from those
+                of the refresh token to those listed in scopes. Useful if
+                refresh token has a wild card scope (e.g.
+                'https://www.googleapis.com/auth/any-api').
```
tseaver: I'm ambivalent on the name 'downscope':  is it widely used elsewhere?  For a boolean, I prefer an obvious verb phrase, e.g., `reduce_scopes`.   @theacodes WDYT?

tseaver:
@eugenewf Note that the tests are passing on CI, so you don't have any work to do to get Python 3.4 working.

google/oauth2/credentials.py
eugenewf: I don't know about widely used, but I've also never seen the concept referred to in any other way (ex. https://developers.google.com/identity/protocols/OAuth_ref).

docs/reference/google.auth.app_engine.rst
eugenewf: This was added by the tox doc functions (not sure why). Removed.

docs/reference/google.auth.compute_engine.credentials.rst
eugenewf: Removed.

docs/reference/google.auth.environment_vars.rst
eugenewf: Removed.

docs/reference/google.oauth2.id_token.rst
eugenewf: Removed.

docs/reference/google.auth.compute_engine.rst
eugenewf: Removed.

docs/reference/google.auth.crypt.rst
eugenewf: No credit to me :) the auto doc functions caught this.

docs/reference/google.oauth2.service_account.rst
eugenewf: Removed.

docs/reference/google.auth.app_engine.rst
eugenewf: Should I add an issue about this?

tseaver: What is "tox doc"? Ahh, I guess you must have run `tox -e docgen`?

I see that `CONTRIBUTING.rst` suggests running it, but I'm not sure why, since the generated files typically don't contain stuff relevant to "ordinary" fixes like the one in this PR:  it should matter only if new modules are being added.  @theacodes can you comment?

eugenewf: Yes, that's correct. I basically ran all the suggested tox commands, including 'tox -e docgen'.  First time I've ever used it, so I've got to admit I don't really understand everything it does.

tseaver: Given that `sphinx-apidoc` produces that markup, I guess I was wrong in my assertion that the `\_` is invalid markup.  But it *is* unnecessary.

eugenewf:
@theacodes Are there any further actions you'd like me to take here? (e.g. with regards to https://github.com/googleapis/google-auth-library-python/pull/309#discussion_r236813911 or https://github.com/googleapis/google-auth-library-python/pull/309#discussion_r236777477)

theacodes:
Not at the moment. I just need to find time to review.

google/oauth2/credentials.py
```diff
@@ -71,6 +71,10 @@ def __init__(self, token, refresh_token=None, id_token=None,
                 to obtain authorization. This is a purely informative parameter
                 that can be used by :meth:`has_scopes`. OAuth 2.0 credentials
                 can not request additional scopes after authorization.
+            downscope (bool): Whether to reduce the requested scopes from those
```
theacodes: Hmm. This class is explicitly `ReadOnlyScoped` but `downscope` allows scopes to be modified (albeit not directly), breaking the constraint.

I'd prefer a new method, `credentials.downscope(new_scopes="...")` that explicitly creates a new copy of the credentials with modified scopes. This retains the read-only constraint and makes the downscoping process explicit.

For example:

```
broad_credentials = oauth2.Credentials.from_authorized_user_info(...)
downscoped_credentials = broad_credentials.downscope(["one", "two"])
```

eugenewf: The scopes are set at initialization, and changing them afterwords would require changing private state (one or both of _scopes or _downscope). I don't see any way to get around storing that state either, since the refresh function of the new credentials object would need to be aware that it needs to request reduced scopes. It needs to ask for reduced scope every time, since access token request scope defaults to the full scope granted by the refresh token (https://tools.ietf.org/html/rfc6749#section-6). I can't think of any way to implement a downscope function that doesn't circle back around to creating an object that has basically the same implementation as the one proposed here.

Side Note: The credentials as written could already be considered to have mutable scopes, since (per https://tools.ietf.org/html/rfc6749#section-5.1) the server is not required to return an access token with all the scopes of the refresh token.

eugenewf:
Apologies for dropping this for so long, I've been rather ill.

I had a conversation with Thea back in February resulting in a recommendation to always explicitly request the scopes if they have been provided.  I've written that up and added a failure message for the case where the server only grants some of the scopes requested (which is allowed by the oauth spec).  Comments appreciated if anyone thinks there's a better way to handle that case :)

PR #313, opened 2018-12-17
_default.default() now handles ImportErrors for optional dependencies.

jwwilson-12:
Handle ImportError when importing optional dependencies such as app_engine and compute_engine.

No unit tests yet pending input from maintainers.

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -173,9 +173,9 @@ def _get_explicit_environ_credentials():
 def _get_gae_credentials():
     """Gets Google App Engine App Identity credentials and project ID."""
     try:
-      from google.auth import app_engine
+        from google.auth import app_engine
     except ImportError:
-      return None, None
+        return None, None
 
     try:
         credentials = app_engine.Credentials()
@@ -192,10 +192,10 @@ def _get_gce_credentials(request=None):
     # uses http.client. This is only acceptable because the metadata server
     # doesn't do SSL and never requires proxies.
     try:
-      from google.auth import compute_engine
-      from google.auth.compute_engine import _metadata
+        from google.auth import compute_engine
+        from google.auth.compute_engine import _metadata
     except ImportError:
-      return None, None
+        return None, None
 
     if request is None:
         request = google.auth.transport._http_client.Request()
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -172,8 +172,8 @@ def _get_explicit_environ_credentials():
 
 def _get_gae_credentials():
     """Gets Google App Engine App Identity credentials and project ID."""
-    # While this library is normally bundled with app_engine, there are some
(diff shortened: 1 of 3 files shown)


tseaver:
@jwwilson-12 `google.auth.app_engine` and `google.auth.compute_engine` are part of this package, not dependencies, and both are designed *not* to raise any `ImportError` exceptions (they raise `RuntimeError` if you try to *use* an API outside the appropriate environment, but the code you are patching handles that case already).

jwwilson-12:
There is infrastructure inside Google that makes these dependencies optional. In that context, it's important to be able to handle this condition.

google/auth/_default.py
```diff
@@ -172,7 +172,10 @@ def _get_explicit_environ_credentials():
 
 def _get_gae_credentials():
     """Gets Google App Engine App Identity credentials and project ID."""
-    from google.auth import app_engine
+    try:
+        from google.auth import app_engine
```
tseaver: This case is odd enough to warrant a comment explaining (to our future selves) how the import could possibly fail.

google/auth/_default.py
```diff
@@ -188,8 +191,11 @@ def _get_gce_credentials(request=None):
     # to require no arguments. So, we'll use the _http_client transport which
     # uses http.client. This is only acceptable because the metadata server
     # doesn't do SSL and never requires proxies.
-    from google.auth import compute_engine
-    from google.auth.compute_engine import _metadata
+    try:
+        from google.auth import compute_engine
+        from google.auth.compute_engine import _metadata
```
tseaver: Ditto here.

jwwilson-12: Done

google/auth/_default.py
jwwilson-12: Done

tests/test__default.py
```diff
@@ -235,6 +235,15 @@ def test__get_gae_credentials(app_identity):
     assert project_id == mock.sentinel.project
 
 
+def test__get_gae_credentials_no_app_engine():
+    import sys
+    with mock.patch.dict('sys.modules'):
```
jwwilson-12: Thanks! Any idea why Travis is still failing the coverage check? https://travis-ci.org/googleapis/google-auth-library-python/jobs/469551157

crwilcox: `.tox/cover/lib/python3.6/site-packages/google/auth/_default.py                       104      2     26      0    98%   179-180` so it thinks it isn't testing an import error being thrown from `_get_gae_credentials`

It does seem to me this tests should be covering that line.

jwwilson-12: OK. Should I submit this CL? My concern is that Travis will continue to fail for future PRs if I do.

crwilcox:
This was showing a bug in mocking. I changed an import style so it is now being mocked. Coverage should be 100% now.

PR #322, opened 2019-02-12
Fix `pex` warning

jeffwidman:
I was using `pex` on an application that lists `google-auth` as a dependency and got the following warning:

    ..../pex/environment.py:330 UserWarning: The `pkg_resources` package was loaded from a pex
    vendored version when declaring namespace packages defined by google-auth 1.6.2. The 
    google-auth 1.6.2 distribution should fix its `install_requires` to include `setuptools`

So adding `setuptools` as a listed dependency to fix this.

Also making the listing alphabetical.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,10 +19,11 @@ from setuptools import setup
 
 
 DEPENDENCIES = (
+    'cachetools>=2.0.0',
     'pyasn1-modules>=0.2.1',
     'rsa>=3.1.4',
+    'setuptools>=40.3.0',
     'six>=1.9.0',
-    'cachetools>=2.0.0',
 )
 
 


jeffwidman:
Note: 
I am a bit unclear on the actual root cause of why `pex` is complaining about this.

But it seems like a harmless change.

theacodes:
This is fine, but can we set a minimum version on setuptools? It's unlikely this will work with *any* setuptools.

jeffwidman:
> This is fine, but can we set a minimum version on setuptools? It's unlikely this will work with _any_ setuptools.

So I thought about that originally myself, but decided against it. 

I agree that it's unlikely to work with all versions. However, let's think about what's happening here:
1. The code is already calling `pkg_resources`, so if an invalid version of setuptools was used, this would already be failing.
2. Therefore, all this PR does is make an implicit dependency explicit to silence a warning. 
3. So the practical reality is that all the versions of `setuptools` that your users are fetching will work. 
4. And if not, you'd be already seeing other bug reports.

Given the above, it would be arguably more confusing to pin a specific minimum version, because that implies there's something _different_ about whatever arbitrary version we pick.

That said, this is your library, so if you want me to pick a specific version, I'm happy to, I just am not sure it makes sense here unless you already have reports of known bugs with older versions of `setuptools`.

theacodes:
There are certainly versions of setuptools with buggy support for namespace
packages.

jeffwidman:
Is there a minimum `setuptools` version that you want to require?

I'm not very familiar with those bugs (or the packaging of namespaced packages in general), so happy to defer to your expertise. 

Also, I'm not clear what level of "bugginess" you're comfortable tolerating. For example, if it's a bug in a recent version, but an edge case that only manifests in certain rare environments, you may want to allow it so that users who aren't able to upgrade their environment but also aren't affected by the edge-case bug aren't excluded from using this library. Or you may not, I don't know...

theacodes:
(apologies, I'm traveling so I'm not able to pull the exact versions)

There were some old versions that would crash due to namespace packages
being present. We can try to track them down, but I think any version newer
than two years old is likely fine.

jeffwidman:
Gotcha. 

Grep'ing https://setuptools.readthedocs.io/en/latest/history.html for "namespace" shows `40.3.0` solved a problem with `pkg_resource`-style namespaces (https://github.com/pypa/setuptools/issues/1321).

That said, `40.3.0` was released Sept 16, 2018 which is fairly recent... 

If that's too new, `38.2.2` from Nov 27, 2017 fixed another bug (https://github.com/pypa/setuptools/issues/1214 solved by https://github.com/pypa/setuptools/pull/1215).

Thoughts?

theacodes:
I'd say let's go with 40.3.0, thank you for doing this research!

jeffwidman:
Sounds good. Updated.

PR #323, opened 2019-02-13
Add retry to _metadata.ping()

rmbarron:
This is to address https://github.com/googleapis/google-auth-library-python/issues/211.

I was able to reproduce the issue in App Engine Flex - my repro is an autoscaling application with minimum 1 instance and max 10 instances that makes calls to google.auth.default() on GET requests. Using apache benchmark to scale it up, there was a brief period of time '[Errno 111] Connection Refused' would be intermittently thrown (during the scale up event). Adding this retry has resolved this for me.

This has been tested manually. Please let me know if there are any integration tests I should run first, though this change should be low risk and largely invisible to all users of _metadata.py.

Many Thanks!

### DIFF
--- a/google/auth/compute_engine/_metadata.py
+++ b/google/auth/compute_engine/_metadata.py
@@ -51,13 +51,15 @@ except ValueError:  # pragma: NO COVER
     _METADATA_DEFAULT_TIMEOUT = 3
 
 
-def ping(request, timeout=_METADATA_DEFAULT_TIMEOUT):
+def ping(request, timeout=_METADATA_DEFAULT_TIMEOUT, retry_count=3):
     """Checks to see if the metadata server is available.
 
     Args:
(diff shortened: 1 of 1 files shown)


theacodes:
The reason we previously did not do this is because we want to prevent the program from hanging for a long time while attempting to contact the metadata service. How long will this hang before raising a fatal error?

rmbarron:
The default timeout is already 3 seconds. So adding a retry default of 3, the worst case scenario of timeout in this lib is 9 seconds. 

The average case is much quicker - in my reproduction, the fatal error would be thrown within a second of a successful one, and another would succeed within the same second. If desired, I can gather deeper specifics - stackdriver likes to group stdout logs as an aggregate timestamp in 1 second intervals.

theacodes:
Yeah this seems fine. Thanks!

PR #333, opened 2019-03-30
Support for `?scopes=` in metadata server

asciimike:
Adds support for requesting access tokens with arbitrary metadata scopes from the compute metadata server.

Note: even though this feature is implemented in the metadata server, this feature is only available in the implementations on GAE Second Generation and GCF, not on GCE or GKE (where this functionality is already available by setting instance scopes).

### DIFF
--- a/google/auth/compute_engine/_metadata.py
+++ b/google/auth/compute_engine/_metadata.py
@@ -179,7 +179,7 @@ def get_service_account_info(request, service_account='default'):
         recursive=True)
 
 
-def get_service_account_token(request, service_account='default'):
+def get_service_account_token(request, service_account='default', scopes=['cloud-platform']):
     """Get the OAuth 2.0 access token for a service account.
 
     Args:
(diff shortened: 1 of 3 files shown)


sqrrrl:
ping

google/auth/compute_engine/_metadata.py
```diff
@@ -198,7 +201,7 @@ def get_service_account_token(request, service_account='default'):
     """
     token_json = get(
         request,
-        'instance/service-accounts/{0}/token'.format(service_account))
+        'instance/service-accounts/{0}/token?scopes={1}'.format(service_account, ",".join(scopes)))
```
sqrrrl: Should URI encode values here.

asciimike:
I've since left Google, so I don't know what the internal status of this feature is or where the GCE or Client Lib team wants to go with it. Unless I get clear direction on that, I'll probably just close the PR and let them sort it out when they've prioritized it appropriately.

busunkim96:
Closing as this is superseded by #376

PR #334, opened 2019-04-02
Flip setup.py dependency to python-cryptography.

wenottingham:
That's the library the implementation prefers, so list that as the dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ from setuptools import setup
 DEPENDENCIES = (
     'cachetools>=2.0.0',
     'pyasn1-modules>=0.2.1',
-    'rsa>=3.1.4',
+    'cryptography>=1.4.0',
     'setuptools>=40.3.0',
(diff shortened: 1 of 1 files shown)


wenottingham:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

signed

busunkim96:
RSA is in the dependencies list of setup.py because it is the "fallback" (though not the preferred). I suspect that's because RSA is pure-Python.

To take this change some code would also have to be modified (one example below, although there may be others)
https://github.com/googleapis/google-auth-library-python/blob/9eec091cd4340f8e57dd0e837a7e8df808f9f5bd/google/auth/crypt/rsa.py#L15-L30

Closing for now, but feel free to open an issue to discuss further.

PR #340, opened 2019-05-09
feat: Implement ES256 for JWT verification

### LINKED ISSUE #157: Support Elliptic curve signatures in JWT module
Currently

[https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/crypt.py][1]

Only supports RSA

It would be great to add support for ES* such as ES256 signing algorithms as well

[1]: https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/709953d34c0c93744606a38e84abce557e73dfcb/google/auth/crypt.py

### LINKED ISSUE #336: JWT can't be verified
I am trying to validate a JWT generated by Google IAP based on this [example](https://cloud.google.com/iap/docs/signed-headers-howto) but it seems that verifying the token causes problems on the google.auth.jwt module

I created this token as an example using the tool in https://jwt.io::

```python
from google.auth import jwt
token = 'eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IkxZeVAyZyJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlhdCI6MTUxNjIzOTAyMn0.OjJokSnHIdMxqTlpT055GJDe72-zoTZBE5NISmrDPx0dletHBTnlbl1wwr0EhWaxgKIesZ7N7eLd4XW-TgX-vA'
cert = '-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEEVs/o5+uQbTjL3chynL4wXgUg2R9\nq9UU8I5mEovUf86QZ7kOBIjJwqnzD1omageEHWwHdBO6B+dFabmdT9POxg==\n-----END PUBLIC KEY-----\n'
jwt.decode(token, certs=cert)
```
And got the following error:
```
/tests/jwt/google_auth/venv/lib/python3.5/site-packages/rsa/pem.py", line 92, in load_pem
    raise ValueError('No PEM start marker "%s" found' % pem_start)
ValueError: No PEM start marker "b'-----BEGIN RSA PUBLIC KEY-----'" found
```
It seems weird that the jwt module can't decode a token created with the EC256 algorithm since is the algorithm used by Google IAP.

Any idea of why this is happening?

#### Environment
  - OS: GNU/Linux
  - Python version: 3.5
  - pip version: 19.1
  - `google-auth` version: 1.6.3

Thanks!

theacodes:
Closes #336 
Closes #157


google/auth/jwt.py
```diff
@@ -59,8 +59,18 @@
 from google.auth import exceptions
 import google.auth.credentials
 
+try:
+    from google.auth.crypt import es256
+except ImportError:  # pragma: NO COVER
+    es256 = None
```
bshaffer: Just to be clear on the flow here... 

To use `es256`, a user will include `from google.auth.crypt import es256`, at which point an error is thrown and they can import the dependent library. Otherwise, everything functions as expected.

arithmetic1728: yes this is the behavior

arithmetic1728:
> This all looks fantastic. One clarification question, and also, seems like we should add a bit to [the ID Token docs](https://github.com/googleapis/google-auth-library-python/blob/master/docs/user-guide.rst#identity-tokens)

Yes. docs are now added.

Issue #344, opened 2019-05-31
compute_engine.IDTokenCredentials should use Metadata server and not IAMCredentials API


salrashid123:
`compute_engine.IDTokenCredentials` currently uses the IAM API to get a google-issued OIDC token here:
- https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/compute_engine/credentials.py#L128

THat flow is not necessary since we can get an `id_token` directly from the metadta server.

For example, internally within `compute_engine.IDTokenCredentials`  we should "just get it" directly like so:

```python
from google.oauth2 import id_token
from google.auth import compute_engine
import google.auth
import google.auth.transport.requests
from google.auth.transport.requests import AuthorizedSession
from google.oauth2.credentials import Credentials
import requests

target_audience = 'https://myapp-6w42z6vi3q-uc.a.run.app'
url = 'https://myapp-6w42z6vi3q-uc.a.run.app'
certs_url='https://www.googleapis.com/oauth2/v1/certs'
metadata_url = "http://metadata/computeMetadata/v1/instance/service-accounts/default/identity?audience=" + target_audience

r = requests.get(metadata_url, headers={"Metadata-Flavor":"Google"})

idt=''
creds = google.oauth2.credentials.Credentials(idt)
authed_session = AuthorizedSession(creds)

# make authenticated request
r = authed_session.get(url)
print r.status_code
print r.text
````

salrashid123:
i can submit a PR for this if this change is ok.  i've verified the following works

-  [google.auth.compute_engine.credentials.py](https://gist.github.com/salrashid123/9f34b1a8eda39e0d1bc94115879a109f#file-credentials-py-L123)
-  [google.auth.compute_engine._metadata.py](https://gist.github.com/salrashid123/6297fc63029f853be5d8143e88554f96#file-_metadata-py-L214)

but in looking at the code, the existing implementation did also implement the Singer though i dont' know how it'd be used on its own.

meredithslota:
Closing this issue; see comments in the linked closed PRs for additional context, and see https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/compute_engine/credentials.py#L216 for the implemented changes.

PR #347, opened 2019-06-10
ComputeEngine ID Tokens to use metadata server

salrashid123:
Fixes:
  - https://github.com/googleapis/google-auth-library-python/issues/344

Refactors [google.auth.compute_engine](https://google-auth.readthedocs.io/en/latest/_modules/google/auth/compute_engine/credentials.html#IDTokenCredentials)  implementation to use the  metadata server (which is present in GCP's compute implementations (gce, gke*, cloud run, gcf, gae new runtimes)

### DIFF
--- a/google/auth/compute_engine/credentials.py
+++ b/google/auth/compute_engine/credentials.py
@@ -19,17 +19,13 @@ Engine using the Compute Engine metadata server.
 
 """
 
-import datetime
-
 import six
 
 from google.auth import _helpers
 from google.auth import credentials
 from google.auth import exceptions
 from google.auth import iam
-from google.auth import jwt
 from google.auth.compute_engine import _metadata
-from google.oauth2 import _client
 
 
 class Credentials(credentials.ReadOnlyScoped, credentials.Credentials):
@@ -131,7 +127,9 @@ class IDTokenCredentials(credentials.Credentials, credentials.Signing):
     def __init__(self, request, target_audience,
                  token_uri=_DEFAULT_TOKEN_URI,
                  additional_claims=None,
-                 service_account_email=None):
+                 service_account_email=None,
+                 token_format='standard',
+                 include_license=False):
         """
         Args:
             request (google.auth.transport.Request): The object used to make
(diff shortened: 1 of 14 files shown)


google/auth/compute_engine/credentials.py
```diff
@@ -120,7 +115,7 @@ def requires_scopes(self):
 _DEFAULT_TOKEN_URI = 'https://www.googleapis.com/oauth2/v4/token'
 
 
-class IDTokenCredentials(credentials.Credentials, credentials.Signing):
+class IDTokenCredentials(credentials.Credentials):
```
theacodes: Removing `Signing` breaks the existing interface of this class. Why can't these credentials continue to sign using the IAM API?

salrashid123: Yeah, but the capability shouldn't be there anyway, IMO.  GCE itself doens' thave the capability to sign: it can get `access_tokens` and even an `id_token` but not sign arbitrary blobs by itself.  IMO the computeCredential object should just reflect its baseline capabilities the underlying provider gives here anyway (in this case, the provider is the metadata service).  

The `credentials.Signing` capability is baked into the existing class as really just a side effect  too interms of gettig an IDtoken:  if you were using the IAM api, you could 'just get' an id_token [directly](https://cloud.google.com/iam/credentials/reference/rest/v1/projects.serviceAccounts/generateIdToken)  without signing

Anyway, her'es the biggest problem i see,  that in the existing implementation, the singer and sign_bytes was already _exposed_  directly [here](https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/compute_engine/credentials.py#L229)

```python
    def signer(self):
        return self._signer

    def sign_bytes(self, message):
        return self._signer.sign(message)
```

I don't think people are using that capability to direct sign via IAM but it is there.

---

What the existing implementation is doing is basically allowing the GCE to "impersonate itself" and use the IAM API to get an id token and along the way also allow blob signing.   I would say if someone wants to sigin via the IAM api, the `impersonated_credential` object could do that (i'll file a separate PR for something like [this](https://gist.github.com/salrashid123/5caaa6178e786bfb48515a2cfb7631af#file-impersonated_credentials-py-L242) later

theacodes: > don't think people are using that capability to direct sign via IAM but it is there.

This isn't a good enough reason for us to break a 1.0 interface. this will be a breaking change to users and it will cause issues.

google/auth/compute_engine/credentials.py
```diff
@@ -129,8 +124,6 @@ class IDTokenCredentials(credentials.Credentials, credentials.Signing):
     a service account that has access to the IAM Cloud API.
```
theacodes: You'll need to update these docs as well.

google/auth/compute_engine/credentials.py
salrashid123: no, thats not the reason (its just a statement).  the reasons are outlined above as well as the ability to sign isn't privilege a GCE instance has.  

per offline, we'll can keep the singer capability as well as using the IAM API to do that bit.  Getting the ID token will be through the metadta sever per the most recent commit.

salrashid123:
note, i kept the `additional_claims` parameter still in though it doens't really serve any function
since its not even something oidc will allow you via the external interface (though i can't say if someday we'd allow oidc w/ custom claims)

```python
    def __init__(self, request, target_audience,
                 token_uri=_DEFAULT_TOKEN_URI,
                 additional_claims=None,
                 service_account_email=None):
```

i can also add in warning messages to this PR if someone tries to user the `signer` methods

salrashid123:
@googlebot I consent.

PR #352, opened 2019-07-16
setup.py: set an upper bound on dependencies version

JordanP:
This should make this library more robust in case of breaking changes in one of
its dependencies.


JordanP:
I signed it!

JordanP:
CI failed because it's now running with pytest version 5 released a few days ago and it introduced breaking changes... I guess that proves the point of setting an upper bound :)

PR #353, opened 2019-07-22
Use new pytest api to keep building with pytest 5

scarabeusiv:
Matches was changed to match with at least pytest4 and from pytest 5.0+ it does not work.


busunkim96:
@scarabeusiv Thank you! Could you also change the version pin in tox.ini? 

https://github.com/googleapis/google-auth-library-python/blob/38166b28a6e1614e10b7b279021df28ae40747d7/tox.ini#L9

scarabeusiv:
Sure, commit added.

PR #360, opened 2019-07-29
added load_credentials_from_env implementation

yhuang:
The SDKs on other cloud providers have always supported in-memory account credentials for authentication.  This PR should enable GCloud SDK to do the same.

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -67,7 +67,7 @@ def _warn_about_problematic_credentials(credentials):
         warnings.warn(_CLOUD_SDK_CREDENTIALS_WARNING)
 
 
-def load_credentials_from_env(filename=None):
+def _load_credentials_from_env(filename=None):
     _LOGGER.info("Invoking snowflake safe google credential loader")
     
     """Loads credentials from the environment in a key called "GOOGLE_CREDENTIALS".
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -35,6 +35,7 @@ _LOGGER = logging.getLogger(__name__)
 _AUTHORIZED_USER_TYPE = 'authorized_user'
 _SERVICE_ACCOUNT_TYPE = 'service_account'
 _VALID_TYPES = (_AUTHORIZED_USER_TYPE, _SERVICE_ACCOUNT_TYPE)
+_ENVIRONMENT_KEY = 'GOOGLE_CREDENTIALS'
 
 # Help message when no credentials can be found.
 _HELP_MESSAGE = """\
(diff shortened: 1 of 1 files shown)


google/auth/_default.py
```diff
@@ -66,6 +67,72 @@ def _warn_about_problematic_credentials(credentials):
...
+                _ENVIRONMENT_KEY)
+            new_exc = exceptions.DefaultCredentialsError(msg, caught_exc)
+            six.raise_from(new_exc, caught_exc)
+        return credentials, info.get('project_id')
+
+    else:
+        raise exceptions.DefaultCredentialsError(
+            'The file {file} does not have a valid type. '
+            'Type is {type}, expected one of {valid_types}.'.format(
+                file=_ENVIRONMENT_KEY, type=credential_type, valid_types=_VALID_TYPES))
```
tseaver: Rather than copying in all this logic from `_load_credentials_from_file`, please separate it out into a helper function, e.g.:

```python
def _get_credentials_project_from_json(info, source):
    """Parse JSON credential info into credentials and project.

    Args:
        info (dict): JSON credentials from file / environment.
        source (str): label for source of credentials (filename,
            env var)
    Raises:
        google.auth.exceptions.DefaultCredentialsError: if the file is in the
            wrong format or is missing.
    """
    credential_type = info.get('type')

    if credential_type == _AUTHORIZED_USER_TYPE:

        try:
            credentials = _cloud_sdk.load_authorized_user_credentials(info)
        except ValueError as caught_exc:
            msg = 'Failed to load authorized user credentials from {}'.format(
                source)
            new_exc = exceptions.DefaultCredentialsError(msg, caught_exc)
            six.raise_from(new_exc, caught_exc)
        # Authorized user credentials do not contain the project ID.
        _warn_about_problematic_credentials(credentials)
        return credentials, None

    elif credential_type == _SERVICE_ACCOUNT_TYPE:
        from google.oauth2 import service_account

        try:
            credentials = (
                service_account.Credentials.from_service_account_info(info))
        except ValueError as caught_exc:
            msg = 'Failed to load service account credentials from {}'.format(
                source)
            new_exc = exceptions.DefaultCredentialsError(msg, caught_exc)
            six.raise_from(new_exc, caught_exc)
        return credentials, info.get('project_id')

    else:
        raise exceptions.DefaultCredentialsError(
            '{source} does not have a valid type. '
            'Type is {type}, expected one of {valid_types}.'.format(
                source=source, type=credential_type, valid_types=_VALID_TYPES))
```

and call it from both `_load_credentials_from_file` and `_load_credentials_from_env`.

busunkim96:
Hi @yhuang - thank you for submitting a PR.

We have credential loading from the environment. This is what happens when you do 

```
import google.auth

credentials, project_id = google.auth.default()
```

The library searches through a list of known common locations. [see here](https://google-auth.readthedocs.io/en/latest/_modules/google/auth/_default.html#default). This is consistent across auth library implementations in all the languages we support. We'd prefer to maintain as much consistency as possible for a predictable developer experience.

Is there a part of the current implementation that does not meet your needs?

tseaver:
@busunkim96
> Is there a part of the current implementation that does not meet your needs?

This PR intends to allow for the actual credentials to be stored in the environment as a JSON string, rather than as a filename pointing to a filename containing them.

busunkim96:
@tseaver I see -- I was confused by the docstrings that mention filenames. :sweat_smile: 

@broady @tswast Is loading credentials from environment variables supported in the other auth libraries?

tswast:
> Is loading credentials from environment variables supported in the other auth libraries?

No. We'd have to update the Application Default Credentials spec if we wanted to support such an option.

In-memory credentials are supported by explicitly constructing objects with the following constructors:

- [google.oauth2.credentials.Credentials.from_authorized_user_info](https://google-auth.readthedocs.io/en/latest/reference/google.oauth2.credentials.html#google.oauth2.credentials.Credentials.from_authorized_user_info)
- [google.oauth2.service_account.Credentials.from_service_account_info](https://google-auth.readthedocs.io/en/latest/reference/google.oauth2.service_account.html#google.oauth2.service_account.Credentials.from_service_account_info)

I would be interested in a method that created credentials from either kind of JSON object, but wary of introducing another environment variable for such a purpose.

erikwebb:
> I would be interested in a method that created credentials from either kind of JSON object, but wary of introducing another environment variable for such a purpose.

@tswast Just to be clear, if we re-used `GOOGLE_APPLICATION_CREDENTIALS` and overloaded it to use *either* a file path *or* JSON string, would that be an acceptable solution?

busunkim96:
Hi @erikwebb, I believe not. Tim referenced the [Application Default Credentials Spec](https://cloud.google.com/docs/authentication/production), which says:

> First, ADC checks to see if the environment variable GOOGLE_APPLICATION_CREDENTIALS is set. If the variable is set, ADC uses the *service account file* that the variable points to. The next section describes how to set the environment variable.
> 
> If the environment variable isn't set, ADC uses the default service account that Compute Engine, Kubernetes Engine, App Engine, and Cloud Functions provide, for applications that run on those services.
> 
> If ADC can't use either of the above credentials, an error occurs.

joaonetto:
Hello. I change the function `from_filename` inside the `_service_account_info.py` to use a dictionary instead to read a .json file.

In this idea, they will check if `filename` variable it is a dict, if they are, they will return a `filename` and `key`, please take a look.

```
def from_filename(filename, require=None):
    """Reads a Google service account JSON file or Dictionary and returns its parsed info.

    Args:
        filename (dict): Test if filename is a dictionary, if they are return filename and get the key.
        filename (str): The path to the service account .json file.
        require (Sequence[str]): List of keys required to be present in the
            info.

    Returns:
        Tuple[ Mapping[str, str], google.auth.crypt.Signer ]: The verified
            info and a signer instance.
    """
    if type(filename) is dict:
        return filename, from_dict(filename, require=require)
    else:
        with io.open(filename, "r", encoding="utf-8") as json_file:
            data = json.load(json_file)
            return data, from_dict(data, require=require)
```

So, this is a simple way to use Dict or File in auth method.

Can anyone give me a feedback about this solution ?

All the best

Netto

joaonetto:
Hello, I changed the `_service_account_info.py` to support a Dict and file, please take a look.

```
def from_filename(filename, require=None):
    """Reads a Google service account JSON file or Dictionary and returns its parsed info.

    Args:
        filename: Test if filename is a dictionary, if they return json.decoder.JSONDecodeError, load a file
        require (Sequence[str]): List of keys required to be present in the
            info.

    Returns:
        Tuple[ Mapping[str, str], google.auth.crypt.Signer ]: The verified
            info and a signer instance.
    """
    try:
        return json.loads(filename), from_dict(json.loads(filename), require=require)
    except json.decoder.JSONDecodeError as error:
        with io.open(filename, "r", encoding="utf-8") as json_file:
            data = json.load(json_file)
            return data, from_dict(data, require=require)
```

So, if we send `google_key` like a dict or file directory, the snipped above will understand the difference between them and will use the right method.

In my call I use:

```
        scopes = ['https://www.googleapis.com/auth/drive']
        credentials = service_account.Credentials.from_service_account_file(
            google_key,
            scopes=scopes
        )
```

Where `google_key` it's a `environment variable`.

All the best,

JNetto

busunkim96:
Closing as this would be against the current ADC spec.

Issue #359, opened 2019-07-29
bug in `with_target_audience` method of `google.auth.compute_engine.IDTokenCredentials`


lmiglio:
Hi, I think I found a problem with `google.auth.compute_engine.credentials.IDTokenCredentials`. When the methods `with_target_audience` and then `refresh` are called, a `TypeError: 'Signer' object is not callable` is raised.

#### Environment details
  - OS: Debian GNU/Linux 9 (stretch) (but I believe it does not depend on the os)
  - Python version: 2.7/3.7
  - `google-auth` version: 1.6.3

#### Steps to reproduce

1. have access either to dev shell or to a `vm` shell
2. run `python`
```python
>>> import google.auth
>>> cred, _ = google.auth.default()
>>> cred
<google.auth.compute_engine.credentials.Credentials object at 0x7fcfb4612810>
>>> from google.auth.transport.requests import Request
>>> from google.auth.compute_engine.credentials import IDTokenCredentials                                                                                                                                            
>>> id_credentials = IDTokenCredentials(Request(), None)
>>> id_credentials
<google.auth.compute_engine.credentials.IDTokenCredentials object at 0x7fcfb3dd4650>
>>> cred_with_audience = id_credentials.with_target_audience("some-existing-client-id") # client id has been removed, put an existing one
>>> cred_with_audience.refresh(Request())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/google/auth/compute_engine/credentials.py", line 217, in refresh
    assertion = self._make_authorization_grant_assertion()
  File "/usr/local/lib/python2.7/dist-packages/google/auth/compute_engine/credentials.py", line 211, in _make_authorization_grant_assertion
    token = jwt.encode(self._signer, payload)
  File "/usr/local/lib/python2.7/dist-packages/google/auth/jwt.py", line 97, in encode
    signature = signer.sign(signing_input)
  File "/usr/local/lib/python2.7/dist-packages/google/auth/iam.py", line 101, in sign
    response = self._make_signing_request(message)
  File "/usr/local/lib/python2.7/dist-packages/google/auth/iam.py", line 78, in _make_signing_request
    self._credentials.before_request(self._request, method, url, headers)
  File "/usr/local/lib/python2.7/dist-packages/google/auth/credentials.py", line 122, in before_request
    self.refresh(request)
[...]

PR #325, opened 2019-02-15
Release 1.6.3

busunkim96:

PR #361, opened 2019-07-30
fix compute engine IDTokenCredentials.with_target_audience method

lmiglio:
Hi everybody, 

Here is the PR for this issue: https://github.com/googleapis/google-auth-library-python/issues/359

It should be backward compatible and it should not break anything.

I added an integration test of `IDTokenCredentials` and `Signer` where I used [`responses`](https://github.com/getsentry/responses) for mocking REST responses, I do not know if that is fine.

I followed [these guidelines](https://github.com/googleapis/google-auth-library-python/blob/master/CONTRIBUTING.rst). 

Hope everything is ok, if not let me know :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,8 +31,11 @@ tests/data/user-key.json
 
 # PyCharm configuration:
 .idea
+venv/
 
 # Generated files
 pylintrc
 pylintrc.test
 pytype_output/
+
+.python-version
--- a/google/auth/compute_engine/credentials.py
+++ b/google/auth/compute_engine/credentials.py
@@ -131,7 +131,8 @@ class IDTokenCredentials(credentials.Credentials, credentials.Signing):
     def __init__(self, request, target_audience,
                  token_uri=_DEFAULT_TOKEN_URI,
                  additional_claims=None,
-                 service_account_email=None):
+                 service_account_email=None,
+                 signer=None):
         """
         Args:
             request (google.auth.transport.Request): The object used to make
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -32,7 +32,7 @@
 
 setup(
     name='google-auth',
-    version='1.6.3',
+    version='1.6.4',
```
busunkim96: We cut separate release PRs to just bump the version and add changelog notes. See #325 for an example.

salrashid123:
cross referencing with PR
https://github.com/googleapis/google-auth-library-python/pull/362

that'll remove using the iam signing just to get ID tokens on compute...
 you can easily an id_token through the metadata server

setup.py
lmiglio: OK, sorry, did not know that. I will revert the version to the one present before.

lmiglio: I squashed the commits in order to have a single commit with the fix (thus avoiding a commit just for reverting the version back, I do not know if this is compliant with your branch strategy). Let me know if there is other I can do. Thanks!

bshaffer:
Hi @lmiglio ! 
Could you fix the merge conflicts so we can merge this PR? I've resolved them [here](https://github.com/googleapis/google-auth-library-python/pull/438/commits/64e9c03dc59b072725799a654fddd4e5d31e0394) for example (see https://github.com/googleapis/google-auth-library-python/pull/438)

bshaffer:
Superceded by #438

PR #362, opened 2019-08-10
Refactor IDToken source for compute_engine

salrashid123:
Fixes https://github.com/googleapis/google-auth-library-python/issues/344

and as a continuation of https://github.com/googleapis/google-auth-library-python/pull/347

_ do not merge_

This PR alters the internal behavior of how an IDToken is acquired through `compute_engine` credentials.  Previously, an idtoken was acquired through (basically) the IAM api and impersonation.  This credential type shoudl get the id_token from the metadata server.   This change alters the source of the token but keeps the IAM signer capability for now.

The token format returned by metadata may include additional fields but in its defaults, is in the same IDToken format provided by the existing IAM credentials signer mechanism.  This pr basically ignores one of the exiting parameters provided, `additioal_claims=`, since that would not work anyway with any credential type.

### DIFF
--- a/google/auth/compute_engine/_metadata.py
+++ b/google/auth/compute_engine/_metadata.py
@@ -213,7 +213,7 @@ def get_service_account_token(request, service_account='default'):
 
 
 def get_id_token(request, service_account='default', target_audience=None,
-                 token_format='standard', include_license="FALSE"):
+                 token_format='standard', include_license=False):
     """Get the OAuth 2.0 access token for a service account.
 
     Args:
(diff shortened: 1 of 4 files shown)


salrashid123:
Any thought about this  proposed change?

the impact as i see it shouldn't be much.  Existing users on compute would've had to enable the IAM api, grant impersonation and then get an id_token via signing and exchanging.  The idtoken would represent the source service account from the GCE instance itself (like what w'ere doing here).

I'm leaving the `Signer` capability in now since people would be using it to do stuff like signURL...eventually, that shoudl also get deprecated but that'll be for later

parthea:
Hi @salrashid123 , I'm going to close this PR due to inactivity but please feel free to re-open it.

PR #368, opened 2019-09-07
Add retry loop  for fetching authentication token if any 'Internal Failure' occurs

doneria-anjali:
This is to prevent errors when client tries to connect with server and fails to fetch OAuth2 token throwing 'Internal Failure'.

Bug - https://bugs.chromium.org/p/chromium/issues/detail?id=1001206

### DIFF
--- a/google/oauth2/_client.py
+++ b/google/oauth2/_client.py
@@ -103,13 +103,11 @@ def _token_endpoint_request(request, token_uri, body):
     }
 
     retry = 0
-
     # retry to fetch token for maximum of two times if any internal failure
     # occurs.
     while True:
         response = request(
             method='POST', url=token_uri, headers=headers, body=body)
-
         response_body = response.data.decode('utf-8')
 
         if response.status == http_client.OK:
--- a/google/oauth2/_client.py
+++ b/google/oauth2/_client.py
@@ -102,13 +102,25 @@ def _token_endpoint_request(request, token_uri, body):
         'content-type': _URLENCODED_CONTENT_TYPE,
     }
 
-    response = request(
-        method='POST', url=token_uri, headers=headers, body=body)
+    retry = 0
 
-    response_body = response.data.decode('utf-8')
+    # retry to fetch token for maximum of two times if any internal failure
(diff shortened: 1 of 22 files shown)


google/oauth2/_client.py
```diff
@@ -102,13 +102,23 @@ def _token_endpoint_request(request, token_uri, body):
...
+        response = request(
+            method='POST', url=token_uri, headers=headers, body=body)
+        response_body = response.data.decode('utf-8')
+
+        if response.status == http_client.OK:
+            break
+        else:
+            error_desc = json.loads(
+                response_body).get('error_description') or ''
+            if error_desc == 'internal_failure' and retry < 1:
```
busunkim96: Will `error_description` always be exactly `internal_failure`?

doneria-anjali: Yup. The exact failure message is:
'RefreshError: ('internal_failure: None', u'{\n  "error": "internal_failure"\n}')'
And we want to retry connection in case of this error specifically.

georgysavva: Hi. I have the same issue and a retry would help. But I think there is a mistake in the current implementation. You are right, the error looks like that: 
```python
RefreshError: ('internal_failure: None', u'{\n "error": "internal_failure"\n}')
``` 
But this means that response body looks like this:
```python
{"error": "internal_failure"}
```
And `error_description` isn't specified, so it will be `None` and the condition won't match.
You should use `error` filed instead. Because the `error` field actually contains string 'internal_failure'
What do you think? @busunkim96  @doneria-anjali

georgysavva:
Hi. I have the same issue and a retry would help. But I think there is a mistake in the current implementation. You are right, the error looks like that:

RefreshError: ('internal_failure: None', u'{\n "error": "internal_failure"\n}')
But this means that response body looks like this:

{"error": "internal_failure"}
And error_description isn't specified, so it will be None and the condition won't match.
You should use error filed instead. Because the error field actually contains string 'internal_failure'
What do you think? @busunkim96 @doneria-anjali

PR #372, opened 2019-09-19
build: fix system tests, move to Kokoro

busunkim96:
* Fix system tests
* Use nox for everything (remove tox.ini).
* Add Kokoro configs.
* Add black.
* Use same lint standards as googleapis/google-cloud-python

You can ignore changes to source code and test files in `google/` and `tests/`, they are the result of running black.


.flake8
```diff
@@ -0,0 +1,9 @@
+# Generated by synthtool. DO NOT EDIT!
```
tswast: Is this actually generated by synthtool? Let's add a comment with a link to the template file if not.

.kokoro/publish-docs.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# install a json parser
+sudo apt-get update
+sudo apt-get -y install software-properties-common
+sudo add-apt-repository universe
+sudo apt-get update
+sudo apt-get -y install jq
+
+# create metadata
+python3 -m docuploader create-metadata \
```
tswast: Does this mean the google-auth docs are moving to google.dev?

.repo-metadata.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "google-auth-library-python",
+  "name_pretty": "Google Auth Python Library",
+  "client_documentation": "https://googleapis.dev/python/google-auth-library-python/latest",
+  "issue_tracker": "https://github.com/googleapis/google-auth-library-python/issues",
+  "release_level": "beta",
```
tswast: Is it really only beta? I would hope we'd have GAed our auth lib by now.

google/auth/_cloud_sdk.py
```diff
@@ -23,20 +23,21 @@
 
 
 # The ~/.config subdirectory containing gcloud credentials.
-_CONFIG_DIRECTORY = 'gcloud'
+_CONFIG_DIRECTORY = "gcloud"
```
tswast: It'd be nice if we did the blacken step as a separate PR.

busunkim96: I'll open a separate PR for the blacken to clean up this PR

.repo-metadata.json
busunkim96: Looks like it is GA (according to the setup.py). Thank you for catching that! https://github.com/googleapis/google-auth-library-python/blob/master/setup.py

.kokoro/publish-docs.sh
busunkim96: Yes, we're looking to get everything moved over for consistency. I'll let the docs publish both places for a bit before I make the redirects. It should be simpler than it was for the google-cloud-python docs.

google/auth/_cloud_sdk.py
busunkim96: #375 375

.flake8
busunkim96: This was not copied in using the template. Removed the comment.

noxfile.py
```diff
@@ -0,0 +1,120 @@
...
+    "flask",
+    "mock",
+    "oauth2client",
+    "pytest",
+    "pytest-cov",
+    "pytest-localserver",
+    "requests",
+    "requests-oauthlib",
+    "urllib3",
+    "cryptography",
```
tswast: Some of these seem required for the actual library to function. Those should be defined in `setup.py`, not here, right?

scripts/obtain_user_auth.py
```diff
@@ -1,65 +0,0 @@
-# Copyright 2016 Google Inc.
```
tswast: Is it intended to delete this file? How are we system testing user-based credentials files?

system_tests/noxfile.py
```diff
@@ -167,66 +169,88 @@ def configure_cloud_sdk(session, application_default_credentials, project=False)
 
 # Test sesssions
 
+TEST_DEPENDENCIES = ["pytest", "requests"]
 
-def session_service_account(session):
-    session.virtualenv = False
+
+@nox.session(python=["2.7", "3.7"])
```
tswast: `["2.7", "3.7"]` is pretty redundant throughout this file. Let's make a constant for the python versions.

tswast (changes requested):

noxfile.py
busunkim96: This was a straight copy from the [tox.ini](https://github.com/googleapis/google-auth-library-python/blob/master/tox.ini). Some of them might not be needed anymore though. I'll go through and check

scripts/obtain_user_auth.py
busunkim96: The removal of `scripts/obtain_user_auth.py` was intentional. The [current script](https://github.com/googleapis/google-auth-library-python/blob/master/scripts/obtain_user_auth.py) uses oauth2client to make the credentials. I added instructions on how to get the credentials using gcloud in the 'Running system tests' section of the `CONTRIBUTING.rst`.

https://github.com/googleapis/google-auth-library-python/blob/use-new-nox/CONTRIBUTING.rst#running-system-tests

tswast (approved):

PR #376, opened 2019-10-23
feat: support scopes for metadata credentials on GCF/Cloud Run/App Engine

sqrrrl:
Based on PR #333 which has been sitting for a while and has some unresolved issues blocking merging. 

Updates that code to be more conservative in terms of compatibility -- default case is same as before and no scopes param is included if with_scopes() never used.

Only thing I'd *like* to change, but has some compatibility risks, is how `google.auth.default(scopes)` works. Currently it only applies the scopes iff the credential requires scoping. This case is a little different -- the scoping is optional as the credential has default scopes.

So instead of:

```
credentials, project_id = google.auth.default(scopes)
```

Code currently needs to be:

```
credentials, project_id = google.auth.default()
credentials = credentials.with_scopes(scopes)
```

Minor difference, but kind of annoying that the short form doesn't work. Either requires changing the behavior of `default()` to always attempting scoping (which sounds like the right behavior anyway) or falsely reporting that scoping is required for the credentials.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,8 @@
+[flake8]
+ignore = E203, E266, E501, W503
+exclude =
+  # Standard linting exemptions.
+  __pycache__,
+  .git,
+  *.pyc,
+  conf.py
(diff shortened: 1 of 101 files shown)


tseaver:
Superseded by #633.

PR #387, opened 2019-11-13
fix: fix check for 'internal_failure' condition

georgysavva:
As I was saying in [this comment](https://github.com/googleapis/google-auth-library-python/pull/368#issuecomment-553357310)
In some cases response doesn't contain `error_description` field. So the retry won't work. The solution is to look at both `error` and `error_description` fields.

### DIFF
--- a/google/oauth2/_client.py
+++ b/google/oauth2/_client.py
@@ -104,17 +104,20 @@ def _token_endpoint_request(request, token_uri, body):
     while True:
         response = request(method="POST", url=token_uri, headers=headers, body=body)
         response_body = response.data.decode("utf-8")
+        response_data = json.loads(response_body)
 
         if response.status == http_client.OK:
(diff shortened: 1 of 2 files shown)


georgysavva:
@googlebot I signed it!

georgysavva:
I fixed the tests, but I can't fight the commit linter :) I think it's okay to merge as is if everything else will be ok

busunkim96:
@georgysavva Thank you for submitting a patch!

ziliangpeng:
@georgysavva Do you think it's possible that, sometimes if we get errors from the server, the response will not be in json format, thus loading json from the response before checking ok would cause exception?

georgysavva:
@ziliangpeng totally makes sense. Should be a good improvement. Fill free to open a PR with the fix.

Issue #351, opened 2019-07-15
AuthMetadataPlugin: gRPC requires it to not block; causes timeouts to not be respected


evanj:
The [gRPC documentation for `AuthMetadataPlugin.__call__`](https://grpc.github.io/grpc/python/grpc.html#grpc.AuthMetadataPlugin.__call__) states: "Implementations of this method must not block." Unfortunately, `google.auth.transport.grpc.AuthMetadataPlugin` violates this requirement. I believe this causes all gRPC calls using one instance of AuthMetadataPlugin to be blocked as long as token refreshing is happening. This may be causing occasional errors where the timeout we pass to Google Cloud API calls is not being respected, since the calls are all blocked waiting on the OAuth token to be refreshed.

## Possible fix

It seems likely this plugin needs to do OAuth token refreshing using a thread pool, like [the `grpc._auth.GoogleCallCredentials` implementation does](https://github.com/grpc/grpc/blob/8cd016f5977f8ae807ab6a990795c51cc58cbd41/src/python/grpcio/grpc/_auth.py#L53). This will have the side benefit of not having multiple threads "race" to refresh the credentials quite as much.

A quick-and-dirty "fix" that would reduce the impact of this error would be to add a timeout on the token refreshing requests. This is probably a good idea anyway. However, this would at least ensure the "impact" of slow/blocked token refreshes is minimized. On our case, we have seen processes that may have been "stuck" for ~15 minutes due to this issue.

## Blocking call path

When using `google.api_core.grpc_helpers.create_channel`, which seems to be what is used by Google's generated GAPIC clients:

1. gRPC calls `google.auth.transport.grpc.AuthMetadataPlugin.__call__(...)`
2. Calls `google.auth.transport.grpc.AuthMetadataPlugin._get_authorization_headers(...)`
3. Calls `self._credentials.before_request(...)` which is `google.auth.credentials.Credentials.before_request`
4. If the credentials have expired, it calls `self.refresh(...)`. In the case of service accounts, this is a `google.oauth2.service_account.Credentials` object.
5. Calls `_client.jwt_grant(...)` which calls `_token_endpoint_request(...)`
6. Calls `request(...)` which is an instance of `google.auth.transport.requests.Request`
7. This calls in to the `requests` API, which eventually makes a *blocking* HTTP request.

michaelawyu:
Thanks, @evanj!

A PR has been sent (#389).

PR #390, opened 2019-11-20
fix: make gRPC auth plugin non-blocking + add default timeout value f…

michaelawyu:
…or requests transport

This commit includes the following changes:
- `transport.grpc.AuthMetadataPlugin` is now non-blocking as gRPC requires
- `transport.requests.Request` now has a default timeout value of 120 seconds so that token refreshing will not be stuck

Resolves: #351

### DIFF
--- a/tests/transport/test_grpc.py
+++ b/tests/transport/test_grpc.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import datetime
+import time
 
 import mock
 import pytest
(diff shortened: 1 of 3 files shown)


tests/transport/test_grpc.py
```diff
@@ -58,6 +59,8 @@ def test_call_no_refresh(self):
 
         plugin(context, callback)
 
+        time.sleep(2)
```
busunkim96: Are these timeouts required for the tests to run?

michaelawyu: No, they are not required; I added them so that the thread pool has some (more) time to finish the work in the rare case that the main thread performs the check before the callback is actually called.

busunkim96:
Please merge when you're ready. The release-please bot should open up a PR soon afterwards that you can also review and merge. :)

PR #393, opened 2019-11-26
fix: in token endpoint request, do not decode the response data if it is not encoded

CodingFanSteve:
The interface of the underlying transport
'google.auth.transport.Request' that makes the token
request does not guarantee the response is encoded. In Python 3, the
non-encoded strings do not have 'decode' attribute. Blindly decoding all
the response could have the token refresh throw here.


CodingFanSteve:
> LGTM ;)

Thanks!

PR #400, opened 2019-12-05
feat: send quota project id in x-goog-user-project for OAuth2 credentials

busunkim96:
When the 3LO credentials are used, the quota project ("quota_project_id") should be sent on every outgoing request in the x-goog-user-project HTTP header/grpc metadata.

### DIFF
--- a/google/oauth2/credentials.py
+++ b/google/oauth2/credentials.py
@@ -58,7 +58,7 @@ class Credentials(credentials.ReadOnlyScoped, credentials.Credentials):
         client_id=None,
         client_secret=None,
         scopes=None,
-        quota_project=None,
+        quota_project_id=None,
     ):
         """
         Args:
@@ -82,7 +82,7 @@ class Credentials(credentials.ReadOnlyScoped, credentials.Credentials):
                 token if refresh information is provided (e.g. The refresh
                 token scopes are a superset of this or contain a wild card
                 scope like 'https://www.googleapis.com/auth/any-api').
-            quota_project (Optional[str]): The project ID used for quota and billing.
+            quota_project_id (Optional[str]): The project ID used for quota and billing.
                 This project may be different from the project used to
                 create the credentials.
         """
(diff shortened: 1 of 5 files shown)


tests/oauth2/test_credentials.py
```diff
@@ -294,6 +294,33 @@ def test_credentials_with_scopes_refresh_failure_raises_refresh_error(
...
         assert creds.valid
 
+    def test_apply_with_quota_project(self):
+        creds = credentials.Credentials(
+            token="token",
+            refresh_token=self.REFRESH_TOKEN,
+            token_uri=self.TOKEN_URI,
+            client_id=self.CLIENT_ID,
+            client_secret=self.CLIENT_SECRET,
+            quota_project="quota-project-123",
```
tmatsuo: I think the field has changed to `quota_project_id`. @broady can confirm

google/oauth2/credentials.py
```diff
@@ -202,6 +218,7 @@ def from_authorized_user_info(cls, info, scopes=None):
             scopes=scopes,
             client_id=info["client_id"],
             client_secret=info["client_secret"],
+            quota_project=info.get("quota_project"),  # quota project may not exist
```
broady: we changed it to `quota_project_id` sorry

busunkim96: Updated!

broady:
LGTM, PTAL @shinfan / @CodingFanSteve

shinfan:
LGTM. Thanks.

Just wanna mention we will also need this quota project as a client-level option.

busunkim96:
@shinfan @broady Thank you for the fast reviews! Is sometime late in Q1 an acceptable timeframe for adding this as a client-level option? I'd prefer to be done with the transition to the Python microgenerator to avoid doing the work twice. :smile:

shinfan:
late Q1 sounds good to me :) Thank you!

srulkapara:
@busunkim96 just a quick note that this code is not fully backward-compatible (not talking about constrcutors - i mean in terms of already-pickled creds-token-files, who are missing the new member - who has no default on `__setstate__`). 
Some of Google's best-practices/demo-quickstarts include pickling these tokens to files, so they will be broken after this merge (while not changing the minor module-version). maybe we should cover those cases as well..
(happened to me, thought maybe you should know)

busunkim96:
@srulkapara Thank you for the heads up! I'll continue to track this in #405

Issue #405, opened 2019-12-10
AttributeError in credentials.py


ryoichitaniguchi:
Hello,

after updating `google-auth`  package from `v1.7.2` to `v1.8.0`  error occurred from `credentials.py`

stack-trace:
```
    range=RANGE).execute()
  File "/home/vagrant/py36/lib/python3.6/site-packages/googleapiclient/_helpers.py", line 130, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "/home/vagrant/py36/lib/python3.6/site-packages/googleapiclient/http.py", line 851, in execute
    method=str(self.method), body=self.body, headers=self.headers)
  File "/home/vagrant/py36/lib/python3.6/site-packages/googleapiclient/http.py", line 165, in _retry_request
    resp, content = http.request(uri, method, *args, **kwargs)
  File "/home/vagrant/py36/lib/python3.6/site-packages/google_auth_httplib2.py", line 187, in request
    self._request, method, uri, request_headers)
  File "/home/vagrant/py36/lib/python3.6/site-packages/google/auth/credentials.py", line 125, in before_request
    self.apply(headers)
  File "/home/vagrant/py36/lib/python3.6/site-packages/google/oauth2/credentials.py", line 185, in apply
    if self.quota_project_id is not None:
  File "/home/vagrant/py36/lib/python3.6/site-packages/google/oauth2/credentials.py", line 134, in quota_project_id
    return self._quota_project_id
AttributeError: 'Credentials' object has no attribute '_quota_project_id'
```

Trigger is googleapiclient.http.execute():
```
from googleapiclient.discovery import build
build('sheets', 'v4', credentials=creds).spreadsheets().values().get(spreadsheetId=SPREADSHEET_ID,range=RANGE).execute()
```

Downgrading google-auth to previous version (or avoid to refer this attribute) , error dissapeared. 

Some compatibility issue behind this?  Appreciate if someone let me know..

requirements condition
```
google-api-python-client
google-auth-httplib2
google-auth-oauthlib
google-auth
```

installed version from pip-list
```
google-api-python-client (1.7.11)
google-auth (1.8.0)     # or 1.8.1
google-auth-httplib2 (0.0.3)
google-auth-oauthlib (0.4.1)
```

maxiller:
@ryoichitaniguchi thanks for the suggestion - rolling back to google-auth==1.7.2 workarounded for me

busunkim96:
[...]

PR #407, opened 2019-12-10
Revert "feat: send quota project id in x-goog-user-project for OAuth2 credentials"

busunkim96:
Reverts googleapis/google-auth-library-python#400

In addition to a comment on the original PR, folks have filed issue #405. Credentials pickled with the previous class definition (with no quota project) cannot be correctly unpickled. Many official docs pages have code to pickle credentials files, so we should make sure this use case is not broken.

### DIFF
--- a/docs/reference/google.auth.crypt.base.rst
+++ b/docs/reference/google.auth.crypt.base.rst
@@ -1,7 +1,0 @@
-google.auth.crypt.base module
-=============================
-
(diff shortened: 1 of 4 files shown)


tswast:
Sounds like we need to implement `__getstate__ ` and `__setstate__ ` if we're going to be able to add any new properties and still support pickle files. https://stackoverflow.com/a/1939384/101923

PR #424, opened 2020-01-17
feat: distinguish transport and execution time timeouts

### LINKED ISSUE #423: Make a clear distinction between the transport timeout and the total method time timeout
It turned out that the concern from #406 was valid, unfortunately:

> A downstream library that is timeout-sensitive might still be affected by the change, because the new explicit timeout parameter changes the semantics of AuthorizedSession.request() . The timeout represents the total allowed time for all requests made behind the scenes (e.g. credentials refresh attempts), as opposed to a previous per-request timeout.

The changed semantics of `timeout` caused an [issue](https://github.com/googleapis/google-cloud-python/issues/10137) in the `storage` library. A user reported that if a large file upload takes, say, minutes _and succeeds_, a ``Timeout`` error is still raised, because the total execution time takes more than ~60 seconds, which the default transport timeout configured here in this library here.

### Proposed solution
Leave the `timeout` parameter as is and still pass it to the transport, but introduce and expose a new optional parameter `max_allowed_time` that can be used instead to configure the "total timeout".

> NOTE: The method itself might still take longer than `max_allowed_time`, because the underlying request might take longer, which is the nature of the `requests` library's timeout ([docs](https://2.python-requests.org/en/master/user/quickstart/#timeouts)).

### Alternatives
Modifying the library code so that it supports "absolute total timeouts", but that would require interrupting requests in-flight, which does not seem to be a straightforward task -https://stackoverflow.com/q/21965484/5040035

plamut:
Closes #423. 
Towards https://github.com/googleapis/google-cloud-python/issues/10137

This PR makes a distinction between the Requests transport timeout, i.e. the ``timeout`` parameter, and the total allowed execution time for the `request()` method, i.e. the ``max_allowed_time`` argument. These two are not mixed together anymore.

In other words, if `timeout` is passed in (set to, say, 60 seconds) and an underlying request takes longer, but the times between the bytes received by the socket are all under 60 seconds, the `TimeoutGuard` should **not** kick in and cause a timeout error (the cause of [Storage #10137](https://github.com/googleapis/google-cloud-python/issues/10137)).

`TimeoutGuard` is only affected by ``max_allowed_time``, which is `None` by default, making the guard a no-op.

### DIFF
--- a/google/auth/transport/requests.py
+++ b/google/auth/transport/requests.py
@@ -340,7 +340,7 @@ class AuthorizedSession(requests.Session):
             remaining_time = guard.remaining_timeout
 
             # Recurse. Pass in the original headers, not our modified set, but
-            # do pass the adjusted timeout (i.e. the remaining time).
+            # do pass the adjusted max allowed time (i.e. the remaining total time).
             return self.request(
                 method,
                 url,
--- a/google/auth/transport/requests.py
+++ b/google/auth/transport/requests.py
@@ -239,18 +239,38 @@ class AuthorizedSession(requests.Session):
         # credentials.refresh).
         self._auth_request = auth_request
 
-    def request(self, method, url, data=None, headers=None, timeout=None, **kwargs):
+    def request(
+        self,
+        method,
+        url,
+        data=None,
+        headers=None,
+        max_allowed_time=None,
+        timeout=None,
+        **kwargs
+    ):
         """Implementation of Requests' request.
 
         Args:
-            timeout (Optional[Union[float, Tuple[float, float]]]): The number
-                of seconds to wait before raising a ``Timeout`` exception. If
-                multiple requests are made under the hood, ``timeout`` is
-                interpreted as the approximate total time of **all** requests.
+            timeout (Optional[Union[float, Tuple[float, float]]]):
+                The amount of time in seconds to wait for the server response
+                with each individual request.
 
-                If passed as a tuple ``(connect_timeout, read_timeout)``, the
-                smaller of the values is taken as the total allowed time across
-                all requests.
+                Can also be passed as a tuple (connect_timeout, read_timeout).
+                See :meth:`requests.Session.request` documentation for details.
+
+            max_allowed_time (Optional[float]):
+                If the method runs longer than this, a ``Timeout`` exception is
+                automatically raised. Unlike the ``timeout` parameter, this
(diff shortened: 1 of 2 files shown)


google/auth/transport/requests.py
```diff
@@ -239,18 +239,38 @@ def __init__(
...
+                with each individual request.
+
+                Can also be passed as a tuple (connect_timeout, read_timeout).
+                See :meth:`requests.Session.request` documentation for details.
+
+            max_allowed_time (Optional[float]):
+                If the method runs longer than this, a ``Timeout`` exception is
+                automatically raised. Unlike the ``timeout` parameter, this
+                value applies to the total method execution time, even if
+                multiple requests are made under the hood.
```
busunkim96: Does 'multiple requests' mean retries?

google/auth/transport/requests.py
```diff
@@ -332,6 +346,7 @@ def request(self, method, url, data=None, headers=None, timeout=None, **kwargs):
                 url,
```
busunkim96: Could you modify the comment on Line 343 to say 'adjusted max_allowed_time' (or something along those lines) rather than 'adjusted timeout'?

GitHub won't let me comment on the line itself :sweat_smile:

tests/transport/test_requests.py
```diff
@@ -284,7 +304,10 @@ def test_request_timeout_w_refresh_timeout_and_tuple_timeout(self, frozen_time):
         )
         authed_session.mount(self.TEST_URL, adapter)
 
-        # The shortest timeout will prevail and cause a Timeout error, despite
-        # other timeouts being quite long.
+        # An UNATHORIZED response triggers a refresh (an extra request), thus
```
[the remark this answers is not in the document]
plamut: oops, thanks

google/auth/transport/requests.py
plamut: Usually this will mean retries, yes, e.g. when the initial response is UNAUTHORIZED and the token needs to be refreshed.

In addition, an extra request can also be sent by `self.credentials.before_request()`, if the latter determines that the current token is not valid - say, at the very first request in the session, when there's no token yet.

google/auth/transport/requests.py
plamut: Done, thanks.

PR #426, opened 2020-01-17
test: add no cover for py2 collections import

busunkim96:
The coverage report is generated from the unit test session that runs last (3.7), which won't invoke the 'except' clause.

### DIFF
--- a/google/auth/jwt.py
+++ b/google/auth/jwt.py
@@ -42,7 +42,8 @@ You can also skip verification::
 
 try:
     from collections.abc import Mapping
-except ImportError: # Python 2.7 compatibility
+# Python 2.7 compatibility
+except ImportError:  # pragma: NO COVER
     from collections import Mapping
 import copy
 import datetime


google/auth/jwt.py
```diff
@@ -42,7 +42,8 @@
 
 try:
     from collections.abc import Mapping
-except ImportError: # Python 2.7 compatibility
+# Python 2.7 compatibility
+except ImportError:  # pragma: NO COVER
```
plamut: @busunkim96 We had a similar case in BigQuery - an optional import that might fail - and worked around the coverage by mimicking an import failure and covering that case in an actual test.

[failed import fixture](https://github.com/googleapis/google-cloud-python/blob/ae54dc347351c713ac7d343bb77210f3f7d013af/bigquery/tests/unit/test_magics.py#L71-L81)

The magic is performed by the common test helpers `maybe_fail_import` that patches the built-in `__import__` function.

If you believe there is value in this, we can add an extra test covering the missing case, but otherwise a pragma comment would do, too.

busunkim96: I wasn't aware this was a possibility, thank you for raising it here. :+1:  I opened #427 to track.

PR #435, opened 2020-01-23
feat: add non-None default timeout to AuthorizedSession.request()

### LINKED ISSUE #434: Set the default request timeout to something other than `None`
The [AuthorizedSession.requests()](https://github.com/googleapis/google-auth-library-python/blob/1b9de8dfbe4523f3170e47985ab523cb7865de48/google/auth/transport/requests.py#L242-L251) method does not specify a default timeout.

Since production code would almost always want to have some sort of a timeout, let's add one. The libraries that rely on `google-auth`, but do (yet) not specify a timeout themselves, would benefit from that.

Example: [storage issue](https://github.com/googleapis/google-cloud-python/issues/10182)

plamut:
Closes #434.

This PR adds a non-`None` default timeout to `AuthorizedSession.request()` to prevent requests from hanging indefinitely in a default case.

Should help with https://github.com/googleapis/google-cloud-python/issues/10182

### DIFF
--- a/google/auth/transport/requests.py
+++ b/google/auth/transport/requests.py
@@ -42,6 +42,8 @@ from google.auth import transport
 
 _LOGGER = logging.getLogger(__name__)
 
+_DEFAULT_TIMEOUT = 120  # in seconds
+
 
 class _Response(transport.Response):
     """Requests transport response adapter.
@@ -141,7 +143,13 @@ class Request(transport.Request):
         self.session = session
 
     def __call__(
-        self, url, method="GET", body=None, headers=None, timeout=120, **kwargs
+        self,
+        url,
+        method="GET",
+        body=None,
+        headers=None,
+        timeout=_DEFAULT_TIMEOUT,
+        **kwargs,
     ):
         """Make an HTTP request using requests.
 
@@ -246,8 +254,8 @@ class AuthorizedSession(requests.Session):
         data=None,
         headers=None,
         max_allowed_time=None,
-        timeout=None,
-        **kwargs
+        timeout=_DEFAULT_TIMEOUT,
+        **kwargs,
     ):
         """Implementation of Requests' request.
 
@@ -306,7 +314,7 @@ class AuthorizedSession(requests.Session):
                 data=data,
                 headers=request_headers,
                 timeout=timeout,
-                **kwargs
+                **kwargs,
             )
         remaining_time = guard.remaining_timeout
 
@@ -349,7 +357,7 @@ class AuthorizedSession(requests.Session):
                 max_allowed_time=remaining_time,
                 timeout=timeout,
                 _credential_refresh_attempt=_credential_refresh_attempt + 1,
-                **kwargs
+                **kwargs,
             )
 
         return response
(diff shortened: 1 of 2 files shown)


plamut:
Well go figure, running the `blacken` session can cause syntax errors, because a trailing comma is not allowed after **kwargs in Python <= 3.5. :smile:

google/auth/transport/requests.py
```diff
@@ -246,7 +254,7 @@ def request(
         data=None,
         headers=None,
         max_allowed_time=None,
-        timeout=None,
+        timeout=_DEFAULT_TIMEOUT,
```
plamut: This can still be overridden by `None` if `None` is the default in a higher-level library... huh. I don't see any cases when we would actually want requests to wait indefinitely?

busunkim96: The docs currently say that the transport-specific default timeout is used. https://google-auth.readthedocs.io/en/latest/reference/google.auth.transport.html?highlight=timeout#google.auth.transport.Request.__call__ Could you go through and update that?

If a higher level library is passing None I think we should fix it in that library.  Someone might have a weird reason to want a `None` timeout... I'm open to other opinions though, since I'm not very familiar with how the handwritten libraries use this.

plamut: That `Request` class is used for refreshing the `auth` credentials (in the `request()` method) - https://github.com/googleapis/google-auth-library-python/blob/1b9de8dfbe4523f3170e47985ab523cb7865de48/google/auth/transport/requests.py#L236-L240

The "main" request still goes through the `requests.Session` class (the parent class of `AuthorizedSession`).

I agree that a higher-level library should fix that, but I'm not sure if that is feasible in the short term (talking for `storage` where the issue has been reported). I err on the side of fixing the user bug sooner rather than later, with the assumption that a desire to have a `None` timeout is a very marginal use case.

I might be wrong, of course, thus any insight of the higher-level library maintainers would be much appreciated.

google/auth/transport/requests.py
```diff
@@ -281,6 +289,13 @@ def request(
         # thread-safety.
         _credential_refresh_attempt = kwargs.pop("_credential_refresh_attempt", 0)
 
+        # Until all libraries expose explicit timeouts, a rouge None timeout
+        # might sneak in - for example, a default timeout value in one of the
+        # internal classes. Since it's desired to have at least some timeout,
+        # we override any `None` values for now.
+        if timeout is None:
+            timeout = _DEFAULT_TIMEOUT
```
plamut: This is of course not ideal, but might be a reasonable short- to mid- term fix to stalled requests in higher-level libraries. I am open for feedback of course.

google/auth/transport/requests.py
busunkim96: Whoops, that was a mistake on my part. I would still like input from @crwilcox and perhaps others on where this change should be made. If it will be a faster fix here than we can go forward with the default timeout

google/auth/transport/requests.py
crwilcox: This bit is a bit more concerning to me. If a library specifies None, we should likely respect it.

plamut: Fair enough, I'll purge that commit.

PR #438, opened 2020-02-07
fix: compute engine id token credentials "with_target_audience" method

bshaffer:


bshaffer:
The user @lmiglio has signed the CLA, and the commit is unaltered from the previously submitted #361

PR #439, opened 2020-02-08
fix: set Content-Type header in the request to signBlob API to avoid Invalid JSON payload error

kimamula:
`auth.create_custom_token()` results in an error:

```
Failed to sign custom token. Error calling the IAM signBytes API:{ (...)
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"{\"bytesToSign\": \"...\"}\": Cannot bind query parameter. Field '{\"bytesToSign\": \"...\"}' could not be found in request message.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.BadRequest",
        "fieldViolations": [
          {
            "description": "Invalid JSON payload received. Unknown name \"{\"bytesToSign\": \"...\"}\": Cannot bind query parameter. Field '{\"bytesToSign\": \"...\"}' could not be found in request message."
          }
        ]
      }
    ]
  }
}
```

I have confirmed setting `'Content-Type': 'application/json'` header resolves the problem.


kimamula:
I think this is a serious bug which makes it impossible to create a custom token with Firebase Admin Python SDK.
The feature is crucial for our app which is to be released next week.
Please tell me if there is something I can do for this fix to be released.

kimamula:
I ended up using private key file.

tests/test_iam.py
```diff
@@ -86,6 +86,8 @@ def test_sign_bytes(self):
         returned_signature = signer.sign("123")
 
         assert returned_signature == signature
+        kwargs = request.call_args.kwargs
+        assert kwargs.get("headers").get("Content-Type") == "application/json"
```
tseaver, line 90:
```suggestion
        assert kwargs["headers"]["Content-Type"] == "application/json"
```

PR #448, opened 2020-02-24
feat: add SslCredentials class for mTLS ADC

arithmetic1728:
add SslCredentials class for mTLS ADC (linux only)

go/adc-for-dca-python


arithmetic1728:
For grpc mTLS ADC support.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+        _LOGGER.debug("Failed to run cert provider command with error: %s", str(e))
+        return False, None, None, None, None
+
+    # Check cert provider command execution error.
+    if stderr != b"":
+        _LOGGER.debug("Cert provider command failed with error: %s", stderr)
+        return False, None, None, stdout, stderr
+
+    # Parse stdout, it should be a cert followed by a key, both in PEM format.
+    cert_end = stdout.find(_CERTIFICATE_SURFFIX)
```
busunkim96: Please use a regex for matching the cert and key.

Something like:
```py
import re

_CERTIFICATE_REGEX = r"(?P<cert>.+)-----END CERTIFICATE-----\n(?P<key>.+)"
m = re.match(_CERTIFICATE_REGEX, stdout, re.DOTALL)
```

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+# limitations under the License.
+
+"""Helper functions for getting mTLS cert and key, for internal use only."""
+
+import json
+import logging
+from os import path
+import subprocess
+
+CONTEXT_AWARE_METADATA_PATH = "~/.secureConnect/context_aware_metadata.json"
```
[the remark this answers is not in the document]
arithmetic1728: yes, the file is always in this path

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helper functions for getting mTLS cert and key, for internal use only."""
+
+import json
```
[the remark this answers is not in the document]
arithmetic1728: These are all standard lib imports so I put them in one group, what are the correct orders?

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+import subprocess
+
+CONTEXT_AWARE_METADATA_PATH = "~/.secureConnect/context_aware_metadata.json"
+_CERT_PROVIDER_COMMAND = "cert_provider_command"
+_CERTIFICATE_SURFFIX = b"-----END CERTIFICATE-----\n"
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def read_metadata_file(metadata_path):
```
broady: `_read_dca_metadata_file`

we don't want this as part of the public API for the transport package.

also, I think this should probably move to its own package (we called it "cert" for Go)

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+            The tuple contains the following in order:
+            (1) boolean to show if client cert and key is obtained successfully
+            (2) client certificate in PEM forma if successful, otherwise None
+            (3) client key in PEM format if successful, otherwise None
+            (4) stdout from cert provider command execution
+            (5) stderr from cert provider command execution
+    """
+
+    # Check the system. For now only Linux is supported.
+    if not platform.startswith("linux"):
```
broady: I don't see anything platform-specific about this code.

If Linux is not supported, then the context_aware_metadata.json file won't exist and we won't hit this codepath, right?

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+        _LOGGER.debug("mTLS for platform: %s is not supported.", platform)
+        return False, None, None, None, None
+
+    # Execute the cert provider command in the metadata json file.
+    if _CERT_PROVIDER_COMMAND not in metadata_json:
+        _LOGGER.debug("cert_provider_command missing, skip client SSL authentication")
```
broady: In Go, this is a validation error. If the metadata file exists, it *must* have a non-zero cert_provider_command.

/cc @shinfan @andyrzhao: we should probably start a spec

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+        process = subprocess.Popen(
+            command, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        stdout, stderr = process.communicate()
+    except OSError as e:
+        _LOGGER.debug("Failed to run cert provider command with error: %s", str(e))
+        return False, None, None, None, None
+
+    # Check cert provider command execution error.
+    if stderr != b"":
```
broady: non-empty stderr output doesn't necessary mean the command failed, it might be some log output.

check the exit code is non-zero to indicate failure

google/auth/transport/_mtls_helper.py
broady: is there an actual PEM parser we can use instead? I don't think that we can assume the public key comes before the private key

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+    Args:
+        metadata_json (Dict[str]): metadata JSON file which contains the cert
+            provider command.
+        platform (str): The OS.
+
+    Returns:
+        Tuple[bool, bytes, bytes, bytes, bytes]:
+            The tuple contains the following in order:
+            (1) boolean to show if client cert and key is obtained successfully
+            (2) client certificate in PEM forma if successful, otherwise None
```
broady: s/forma/format/

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+def get_client_ssl_credentials(metadata_json, platform):
+    """Function to get mTLS client side cert and key.
+
+    Args:
+        metadata_json (Dict[str]): metadata JSON file which contains the cert
+            provider command.
+        platform (str): The OS.
+
+    Returns:
+        Tuple[bool, bytes, bytes, bytes, bytes]:
```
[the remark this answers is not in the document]
broady: probably need a class/interface.

we will also want a callback as a client option, similar to Go's WithClientCertSource:
https://godoc.org/google.golang.org/api/option#WithClientCertSource

```
type ClientCertSource = func(*tls.CertificateRequestInfo) (*tls.Certificate, error)
```

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+    return metadata
+
+
+def get_client_ssl_credentials(metadata_json, platform):
+    """Function to get mTLS client side cert and key.
+
+    Args:
+        metadata_json (Dict[str]): metadata JSON file which contains the cert
+            provider command.
+        platform (str): The OS.
```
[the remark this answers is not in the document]
arithmetic1728: no longer has this in the new version.

google/auth/transport/_mtls_helper.py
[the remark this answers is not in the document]
arithmetic1728: import was auto sorted by blacken when I ran "nox -s".

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def read_metadata_file(metadata_path):
+    """Function to load context aware metadata from the given path.
+
+    Args:
+        metadata_path (str): context aware metadata path.
+
+    Returns:
+        Dict[str]:
```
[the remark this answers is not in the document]
arithmetic1728: done

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+"""Helper functions for getting mTLS cert and key, for internal use only."""
+
+import json
+import logging
+from os import path
+import subprocess
+
+CONTEXT_AWARE_METADATA_PATH = "~/.secureConnect/context_aware_metadata.json"
+_CERT_PROVIDER_COMMAND = "cert_provider_command"
+_CERTIFICATE_SURFFIX = b"-----END CERTIFICATE-----\n"
```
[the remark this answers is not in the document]
arithmetic1728: new version no longer uses this.

google/auth/transport/_mtls_helper.py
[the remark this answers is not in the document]
arithmetic1728: new version no longer uses this.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+    if stderr != b"":
+        _LOGGER.debug("Cert provider command failed with error: %s", stderr)
+        return False, None, None, stdout, stderr
+
+    # Parse stdout, it should be a cert followed by a key, both in PEM format.
+    cert_end = stdout.find(_CERTIFICATE_SURFFIX)
+    if cert_end == -1:
+        _LOGGER.debug("Client SSL certificate is missing")
+        return False, None, None, stdout, stderr
+    private_key_start = cert_end + len(_CERTIFICATE_SURFFIX)
```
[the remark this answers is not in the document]
arithmetic1728: new version no longer uses this.

google/auth/transport/_mtls_helper.py
arithmetic1728: new version is using regex to parse PEM.

google/auth/transport/_mtls_helper.py
arithmetic1728: right, removed platform-specific logic.

google/auth/transport/_mtls_helper.py
arithmetic1728: done

google/auth/transport/_mtls_helper.py
arithmetic1728: done

google/auth/transport/_mtls_helper.py
arithmetic1728: changed the name. Let's just keep it here for now. We can refactor once everything (http and grpc) is done.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,107 @@
...
+            return None
+
+    return metadata
+
+
+def get_client_ssl_credentials(metadata_json, platform):
+    """Function to get mTLS client side cert and key.
+
+    Args:
+        metadata_json (Dict[str]): metadata JSON file which contains the cert
```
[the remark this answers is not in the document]
arithmetic1728: done

google/auth/transport/_mtls_helper.py
broady: Oh, I'm wrong, anyway - I was thinking about how Go handles packages (all files in a directory are a package). Which is different than Python, of course.

Because the code is in a file prefixed with `_`, I think that's enough for it to not show up in documentation. Just making sure we're not creating a public API that our customers will see/try to depend on.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,111 @@
...
+import json
+import logging
+from os import path
+import re
+import subprocess
+
+CONTEXT_AWARE_METADATA_PATH = "~/.secureConnect/context_aware_metadata.json"
+_CERT_PROVIDER_COMMAND = "cert_provider_command"
+_CERT_REGEX = re.compile(
+    b"-----BEGIN CERTIFICATE-----.+-----END CERTIFICATE-----\r?\n?", re.DOTALL
```
broady: should `.+` in the middle be `[^-]+`? Would matter if there are two certificates.

google/auth/transport/grpc.py
```diff
@@ -149,3 +154,44 @@ def secure_authorized_channel(
     )
 
     return grpc.secure_channel(target, composite_credentials, **kwargs)
+
+
+class SslCredentials:
+    """Class for application default SSL credentials. For Linux with endpoint
+    verification support, device certificate will be automatically loaded if
```
broady: Add doc URL:

"See https://cloud.google.com/endpoint-verification/docs/overview"

google/auth/transport/grpc.py
```diff
@@ -149,3 +154,44 @@ def secure_authorized_channel(
...
+        )
+        if context_aware_metadata:
+            try:
+                cert, key = _mtls_helper.get_client_ssl_credentials(
+                    context_aware_metadata
+                )
+                self._is_mtls = True
+
+            except (NotImplementedError, OSError, RuntimeError, ValueError) as e:
+                _LOGGER.debug(
```
broady: I don't think these errors should be caught and discarded here.

If you want consistency with the Go implementation, the call to get_client_ssl_credentials would be moved to the ssl_credentials method.

Then _is_mtls = True if context_aware_metadata != None

google/auth/transport/grpc.py
```diff
@@ -149,3 +154,44 @@ def secure_authorized_channel(
...
+            )
+        else:
+            self._ssl_credentials = grpc.ssl_channel_credentials()
+
+    @property
+    def ssl_credentials(self):
+        return self._ssl_credentials
+
+    @property
+    def is_mtls(self):
```
broady: needs docs

google/auth/transport/grpc.py
```diff
@@ -149,3 +154,44 @@ def secure_authorized_channel(
...
+
+        if self._is_mtls:
+            self._ssl_credentials = grpc.ssl_channel_credentials(
+                certificate_chain=cert, private_key=key
+            )
+        else:
+            self._ssl_credentials = grpc.ssl_channel_credentials()
+
+    @property
+    def ssl_credentials(self):
```
broady: needs docs

broady:
Looks pretty good overall... do you have it hooked up to `secure_authorized_channel`, or is that for later?

google/auth/transport/_mtls_helper.py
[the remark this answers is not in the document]
arithmetic1728: now it raises exceptions instead of returning the big tuple.

google/auth/transport/_mtls_helper.py
arithmetic1728: the regex captures all certificates, i.e. the cert chain, because grpc channel credential reads a cert chain and a private key.

google/auth/transport/grpc.py
arithmetic1728: done

google/auth/transport/grpc.py
arithmetic1728: done

google/auth/transport/grpc.py
arithmetic1728: done

arithmetic1728:
> Looks pretty good overall... do you have it hooked up to `secure_authorized_channel`, or is that for later?

Hooked it up to `secure_authorized_channel` in the new commit.

google/auth/transport/_mtls_helper.py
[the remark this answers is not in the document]
arithmetic1728: I ended up adding a new `client_cert_callback` argument to `secure_authorized_channel`, which is mutually exclusive with `ssl_credentials`.
```
def secure_authorized_channel(
    credentials,
    request,
    target,
    ssl_credentials=None,
    client_cert_callback=None,  <============= here
    **kwargs
):
```

google/auth/transport/_mtls_helper.py
arithmetic1728: now raise error to the user

google/auth/transport/grpc.py
arithmetic1728: done

google/auth/transport/grpc.py
```diff
@@ -129,23 +139,113 @@ def secure_authorized_channel(
...
+            credentials loading process on devices with endpoint verification
+            support.
+        RuntimeError: cert provider command runtime error, in application
+            default SSL credentials loading process on devices with endpoint
+            verification support.
+        ValueError:
+            if context aware metadata file is malformed, or cert provider
+            command doesn't produce both client certicate and key, in application
+            default SSL credentials loading process on devices with endpoint
+            verification support.
```
busunkim96:
```suggestion
    Raises:
        OSError: If the cert provider command launch fails during the application default SSL
            credentials loading process on devices with endpoint verification
            support.
        RuntimeError: If the cert provider command has a runtime error during the 
            application default SSL credentials loading process on devices with endpoint verification
            support.
        ValueError:
            If the context aware metadata file is malformed or if the cert provider
            command doesn't produce both client certificate and key during the application
            default SSL credentials loading process on devices with endpoint
            verification support.
```

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    metadata_path = path.expanduser(metadata_path)
+    if not path.exists(metadata_path):
+        _LOGGER.debug("%s is not found, skip client SSL authentication.", metadata_path)
+        return None
+
+    with open(metadata_path) as f:
+        try:
+            metadata = json.load(f)
+        except Exception as e:
```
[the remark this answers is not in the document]
arithmetic1728: This is outdated code. New code doesn't catch and let the user handle it.

google/auth/transport/grpc.py
```diff
@@ -129,23 +139,109 @@ def secure_authorized_channel(
...
+
+    @property
+    def ssl_credentials(self):
+        """Get the created SSL channel credentials.
+
+        For device with endpoint verification support, if device certificate
+        loading has any problems, corresponding exceptions will be raised. For
+        device without endpoint verification support, no exceptions will be
+        raised.
+
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,122 @@
...
+    absolute path; otherwise return None.
+
+    Args:
+        metadata_path (str): context aware metadata path.
+
+    Returns:
+        str: absolute path if exists and None otherwise.
+    """
+    metadata_path = path.expanduser(metadata_path)
+    print(metadata_path)
```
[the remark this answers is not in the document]
arithmetic1728: Removed.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,122 @@
...
+    stdout, stderr = process.communicate()
+    process_return_code = process.returncode
+
+    # Check cert provider command execution error.
+    if process_return_code != 0:
+        raise RuntimeError(
+            "Cert provider command returns non-zero status code %s"
+            % process_return_code
+        )
+
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/grpc.py
```diff
@@ -129,23 +139,113 @@ def secure_authorized_channel(
...
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/grpc.py
```diff
@@ -129,23 +139,113 @@ def secure_authorized_channel(
...
+        Raises:
+            OSError: cert provider command launch failure
+            RuntimeError: cert provider command runtime error
+            ValueError:
+                if context aware metadata file is malformed, or cert provider
+                command doesn't produce both client certicate and key.
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/grpc.py
```diff
@@ -129,23 +139,113 @@ def secure_authorized_channel(
...
+                certificate_chain=cert, private_key=key
+            )
+        else:
+            self._ssl_credentials = grpc.ssl_channel_credentials()
+
+        return self._ssl_credentials
+
+    @property
+    def is_mtls(self):
+        """Property indicating if the created SSL channel credentials is mutual TLS."""
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/grpc.py
```diff
@@ -129,23 +139,113 @@ def secure_authorized_channel(
         target (str): The host and port of the service.
         ssl_credentials (grpc.ChannelCredentials): Optional SSL channel
             credentials. This can be used to specify different certificates.
+            This argument is mutually exclusive with client_cert_callback;
+            providing both will raise an exception.
+            If ssl_credentials is None and client_cert_callback is None or
+            fails, application default SSL credentials will be used.
+        client_cert_callback (Callable[[], (bool, bytes, bytes)]): Optional
```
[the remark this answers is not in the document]
arithmetic1728: Added examples on how to use mtls.

google/auth/transport/grpc.py
arithmetic1728: Done.

tests/transport/test__mtls_helper.py
```diff
@@ -0,0 +1,177 @@
...
+
+    key_match = re.findall(_mtls_helper._KEY_REGEX, content)
+    success = success and len(key_match) == 1 and key_match[0] == expected_key
+
+    return success
+
+
+class TestCertAndKeyRegex(object):
+    def test_cert_and_key(self):
+        # Test signle cert and single key
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,122 @@
...
+    re.DOTALL,
+)
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def _check_dca_metadata_path(metadata_path):
+    """Check the existence of context aware metadata. If exists, return the
+    absolute path; otherwise return None.
+
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,122 @@
...
+        ValueError: If failed to parse metadata as JSON.
+    """
+    with open(metadata_path) as f:
+        metadata = json.load(f)
+
+    return metadata
+
+
+def get_client_ssl_credentials(metadata_json):
+    """Function to get mTLS client side cert and key.
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/_mtls_helper.py
```diff
@@ -0,0 +1,122 @@
...
+    Returns:
+        Tuple[bytes, bytes]: client certificate and key, both in PEM format.
+
+    Raises:
+        OSError: subprocess throws OSError if failed to run cert provider command
+        RuntimeError: if cert provider command has runtime error
+        ValueError:
+            if metadata json file doesn't contain cert provider command, or the
+            execution of this command doesn't produce both client certicate and
+            client key.
```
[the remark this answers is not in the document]
arithmetic1728: Done.

google/auth/transport/grpc.py
```diff
@@ -129,23 +139,113 @@ def secure_authorized_channel(
...
     return grpc.secure_channel(target, composite_credentials, **kwargs)
+
+
+class SslCredentials:
+    """Class for application default SSL credentials.
+
+    For device with endpoint verification support, device certificate will be
+    automatically loaded and mutual TLS will be established.
+    See https://cloud.google.com/endpoint-verification/docs/overview.
+    """
```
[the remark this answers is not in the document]
arithmetic1728: Done.

arithmetic1728:
> Mostly nits on the docstrings to get the style consistent with existing functions. 😄
> 
> Do you have plans for a system test, or is that infeasible?

It is unfortunately infeasible now. mTLS endpoint currently requires provisioning and explicit project white-listing in the server side, which none of us has figured out yet. I will do manual testing after that is figured out, and add system test when mTLS endpoint is GA.

google/auth/transport/grpc.py
```diff
@@ -114,11 +124,91 @@ def secure_authorized_channel(
...
         # Create a channel.
         channel = google.auth.transport.grpc.secure_authorized_channel(
-            credentials, 'speech.googleapis.com:443', request)
+            credentials, regular_endpoint, request,
+            ssl_credentials=grpc.ssl_channel_credentials())
 
         # Use the channel to create a stub.
         cloud_speech.create_Speech_stub(channel)
 
+        # There are actually a couple of options to create a channel, depending
```
busunkim96: This looks great! Thank you for writing this. 

Could you move this out of the existing examples section and format it more like [this docstring](https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/impersonated_credentials.py#L114-L173)? The generated [html docs](https://googleapis.dev/python/google-auth/latest/reference/google.auth.impersonated_credentials.html) are easier to read.

You can view the docs locally by running `nox -s docs` and opening the directory in the browser.

busunkim96:
> It is unfortunately infeasible now. mTLS endpoint currently requires provisioning and explicit project white-listing in the server side, which none of us has figured out yet. I will do manual testing after that is figured out, and add system test when mTLS endpoint is GA.

Does that mean we should not release this feature until it is figured out on the backend?

google/auth/transport/grpc.py
```diff
@@ -129,23 +230,116 @@ def secure_authorized_channel(
...
+        success, cert, key = client_cert_callback()
+        if success:
+            ssl_credentials = grpc.ssl_channel_credentials(
+                certificate_chain=cert, private_key=key
+            )
+
     if ssl_credentials is None:
-        ssl_credentials = grpc.ssl_channel_credentials()
+        adc_ssl_credentils = SslCredentials()
+        ssl_credentials = adc_ssl_credentils.ssl_credentials
```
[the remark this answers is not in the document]
arithmetic1728: Yes you are right. Changed the code.

google/auth/transport/grpc.py
```diff
@@ -129,23 +214,118 @@ def secure_authorized_channel(
...
+        raise ValueError(
+            "Received both ssl_credentials and client_cert_callback; "
+            "these are mutually exclusive."
+        )
+
+    # If SSL credentials are not explicitly set, try client_cert_callback and ADC.
+    if not ssl_credentials:
+        if client_cert_callback:
+            # Use the callback if provided.
+            cert, key = client_cert_callback()
```
chrisdunelm: The design doc states that this callback should have a return value of type `(bool, bytes, bytes)`, which is not what's implemented here. Which is correct?

arithmetic1728: Ah, forgot to update the design doc. Now the design doc for grpc is updated.

PR #451, opened 2020-03-06
fix: only add IAM scope to credentials that can change scopes

### LINKED ISSUE #416: Breaking changes for impersonated_credentials between 1.6.3 and 1.7.0+
#### Environment details

  - OS: MacOS Mojave
  - Python version: 3.7.3
  - pip version: 19.0.3
  - `google-auth` version: 1.6.3, 1.7.0, 1.10.0

The default credentials are my user credentials (with the ServiceAccountTokenCreator role on the service account). 

#### Steps to reproduce
The following code works and produces a valid access token for the service account in version 1.6.3.

```python
from google.auth import impersonated_credentials, default
from google.auth.transport.requests import Request

sa = "<sa>@<project>.iam.gserviceaccount.com"
source_credentials, _ = default()
scopes = ["https://www.googleapis.com/auth/cloud-platform"]
creds = impersonated_credentials.Credentials(
    source_credentials=source_credentials, target_principal=sa, target_scopes=scopes
)

creds.refresh(Request())
print(creds.token)
```
After upgrading to 1.10.0, it fails to authenticate the service account due to invalid scopes.

```
  File "main.py", line 13, in <module>
    creds.refresh(Request())
  File "/google/lib/python3.7/site-packages/google/auth/impersonated_credentials.py", line 218, in refresh
    self._update_token(request)
  File "/google/lib/python3.7/site-packages/google/auth/impersonated_credentials.py", line 234, in _update_token
    self._source_credentials.refresh(request)
  File "/google/lib/python3.7/site-packages/google/oauth2/credentials.py", line 152, in refresh
    self._scopes,
  File "/google/lib/python3.7/site-packages/google/oauth2/_client.py", line 241, in refresh_grant
    response_data = _token_endpoint_request(request, token_uri, body)
  File "/google/lib/python3.7/site-packages/google/oauth2/_client.py", line 115, in _token_endpoint_request
    _handle_error_response(response_body)
  File "/google/lib/python3.7/site-packages/google/oauth2/_client.py", line 60, in _handle_error_response
    raise exceptions.RefreshError(error_details, response_body)
google.auth.exceptions.RefreshError: ('invalid_scope: Bad Request', '{\n  "error": "invalid_scope",\n  "error_description": "Bad Request"\n}')
```

I traced this back to an google-auth upgrade from 1.6.3 to 1.7.0 (same error occurs).

busunkim96:
Closes #416. 

Users reported that they were no longer able to use user credentials as the parent credential. This was happening because user credentials cannot have their scopes modified after creation. 

- [x] add system test for impersonated credentials
- [x] update unit test with user credential as donor credential case
- [x] update CONTRIBUTING.rst with new permissions necessary


google/auth/impersonated_credentials.py
```diff
@@ -205,7 +205,13 @@ def __init__(
         super(Credentials, self).__init__()
 
         self._source_credentials = copy.copy(source_credentials)
-        self._source_credentials._scopes = _IAM_SCOPE
+        # Service account source credentials must have the _IAM_SCOPE
+        # added to refresh correctly. User credentials cannot have
+        # their original scopes modified.
+        try:
+            self._source_credentials = self._source_credentials.with_scopes(_IAM_SCOPE)
```
bshaffer: I think it'd be better to check the `Credentials` type and add the `scope` depending on that, rather than catching an error. This makes the logic more obvious and less error prone.

arithmetic1728: Only the `Credentials` which inherits `google.auth.Credentials.Scoped` has the `with_scopes` method (service account credentials class inherits Scoped, and user credentials class is not). So you can use the following code instead of the try/except block:

```
if isinstance(self._source_credentials, credentials.Scoped):
    self._source_credentials = self._source_credentials.with_scopes(_IAM_SCOPE)
```

arithmetic1728:
LGTM

google/auth/impersonated_credentials.py
busunkim96: Modified, thank you for the suggestion @arithmetic1728!

PR #460, opened 2020-03-16
fix: use AuthMetadataPlugin with context manager

busunkim96:
If this is merged, [google-api-core](https://github.com/googleapis/python-api-core/blob/2b103b60ece16a1e1bc98cfda7ec375191a90f75/google/api_core/grpc_helpers.py#L204) will have to be modified to make use of the context manager.

@lidiz @gnossen Does this look reasonable?


gnossen:
@busunkim96 There are also currently no hooks in gRPC to call your `__enter__` and `__exit__` methods. We've have to make changes to the gRPC Python-level creds API to make this work. What memory leak remains after https://github.com/googleapis/google-auth-library-python/pull/461? If the creds objects are not being garbage collected *ever*, then there's likely a reference cycle issue that needs to be addressed.

Issue #455, opened 2020-03-09
Memory Leak Introduced in 1.7.2


BenRKarl:
Bug discovered via google-ads-python, [original Issue here](https://github.com/googleads/google-ads-python/issues/239).

## Environment details

  - OS: Ubuntu 18.04 and also my gLinux desktop.
  - Python version: 3.7.0
  - pip version: 10.0.1
  - `google-auth` version: 1.7.2+

#### Steps to reproduce

Not exactly sure since this is exposed in google-ads-python via requests that require user credentials, however the code in [this change](https://github.com/googleapis/google-auth-library-python/commit/0c33e9c0fe4f87fa46c8f1a5afe725a467ac5fcc) appears to be the cause. Specifically closing threads in the `__del__` method of `AuthMetadataPlugin` may be causing problems because there are no guarantees around when it's called, and in our case it seems like it may not be called at all.

busunkim96:
@lidizheng Could you provide guidance on the best way to implement gRPC's `AuthMetadataPlugin.__call__`?  I think this library's [implementation](https://github.com/googleapis/google-auth-library-python/blob/0c33e9c0fe4f87fa46c8f1a5afe725a467ac5fcc/google/auth/transport/grpc.py#L35-L91) (using threadpools) was modeled closely on [grpc/_auth.py](https://github.com/grpc/grpc/blob/c683d803306ec3b64bb2e868f8f5118bdecb27ea/src/python/grpcio/grpc/_auth.py#L40-L63).

lidizheng:
Okay. This is a bit awkward.

The code has been out for 3 years, I guess we need to improve its design. I might get back with a PR or a CL? Which one is preferred?

busunkim96:
Thanks for the quick response!  A PR would be preferred. I'm also happy to tweak this myself if you can point me in the right direction.

To be completely honest, I have not worked with threading/threadpool, so I was wondering if there was some obvious way to correct this that I was missing. 😅

lidizheng:
I have a workaround solution that instead of creating a large number of `ThreadPoolExecutor`s, we can have a global one. Like:
```python
_global_thread_pool = None

class AuthMetadataPlugin(...):
    def __init__(self, ...):
        global _global_thread_pool
        ...
        if _global_thread_pool is None:
            _global_thread_pool = futures.ThreadPoolExecutor()  # Let the system default decide number of max_workers
        self._pool = _global_thread_pool
```

[...]

PR #461, opened 2020-03-16
fix: make ThreadPoolExecutor a class var

busunkim96:
Partially addresses #455 using @lidiz's suggestion. This does not solve the memory leak issue.


gnossen:
It's worth noting though that you're now essentially setting your maximum concurrency to `5 * multiprocessing.cpu_count()`

gnossen:
Word of warning. If you want to support Python 2 users in appengine, you're going to want to explicitly specify `max_workers` in the constructor. [I'm speaking from experience.](https://github.com/grpc/grpc/pull/22385)

PR #462, opened 2020-03-16
feat: fetch id token from GCE metadata server

arithmetic1728:


google/auth/compute_engine/credentials.py
```diff
@@ -228,22 +247,76 @@ def _make_authorization_grant_assertion(self):
...
+
+        Raises:
+            google.auth.exceptions.RefreshError: If the Compute Engine metadata
+                service can't be reached or if the instance has no credentials.
+            ValueError: If extracting expiry from the obtained ID token fails.
+        """
+        try:
+            id_token = _metadata.get(
+                request,
+                "instance/service-accounts/default/identity?audience={}".format(
```
bshaffer: There is an issue with the GCE metadata server where the ID Token returned does not include `email` or `email_verified` fields in the ID Token claims.

This is an issue because IAP rejects any ID tokens without an `email` claim. So any requests from GCE to an IAP-protected resource will fail.

The workaround is to include the query parameter `format=full` in the identity endpoint request. I don't like using `format=full` because it returns a LOT of information in the ID token. Also, we expect to have this issue fixed in the coming months, and if we remove `format=full`, that will be a BC breaking change. But this is probably better than releasing a broken integration.

bshaffer:
We also want to detect if the metadata server is available when we are on GKE and Cloud Run. Can you verify this is the case?

google/auth/compute_engine/credentials.py
```diff
@@ -154,29 +160,41 @@ def __init__(
...
-        if additional_claims is not None:
-            self._additional_claims = additional_claims
+        self._additional_claims = {}
+        self._use_metadata_identity_endpoint = use_metadata_identity_endpoint
+
+        if use_metadata_identity_endpoint:
+            # If metadata identity endpoint is used, ignore all other keyword
+            # arguments
+            self._token_uri = None
+            self._signer = None
```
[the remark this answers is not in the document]
arithmetic1728: Done. Did this for `token_uri`, `signer`, `service_account_email` and `additional_claims`.

google/auth/compute_engine/credentials.py
arithmetic1728: added.

google/auth/compute_engine/credentials.py
```diff
@@ -162,39 +162,52 @@ def __init__(
...
                 signer = iam.Signer(
                     request=request,
                     credentials=Credentials(),
-                    service_account_email=service_account_email,
+                    service_account_email=self._service_account_email,
                 )
             self._signer = signer
-
-            self._token_uri = token_uri
+            self._token_uri = _DEFAULT_TOKEN_URI
```
bshaffer: Doesn't this change the previous behavior? It should be:

```py
self._token_uri = token_uri or _DEFAULT_TOKEN_URI
```

arithmetic1728: Oops, thank you for the good catch! I fixed this and added unit test for token_uri.

PR #465, opened 2020-03-20
feat: asyncio support and aiohttp transport

tan01:
Adds asyncio copies and conversions of the oauth2 libraries for
credentials and service accounts.

Most code/tests have been copy-pasted over from synchronous equivalents.
The new functional code is the aio transport abstract base
class and the aiohttp transport implementation. Asyncio requests
can be created with either the aiohttp basic API or by passing
in a ClientSession, and tokens are now refreshed asynchronously.

Asyncio requires Python >=3.5. This change adds dependencies on
aiohttp and pytest-asyncio.

Currently doesn't pass contribution guidelines because it won't work on 2.7. But I'm keeping my fingers crossed.

### DIFF
--- a/docs/reference/google.auth.aio.credentials.rst
+++ b/docs/reference/google.auth.aio.credentials.rst
@@ -1,0 +1,7 @@
+google.auth.aio.credentials module
+==================================
+
+.. automodule:: google.auth.aio.credentials
+   :members:
+   :inherited-members:
+   :show-inheritance:
--- a/docs/reference/google.auth.aio.rst
+++ b/docs/reference/google.auth.aio.rst
@@ -1,0 +1,14 @@
+google.auth.aio package
+=======================
+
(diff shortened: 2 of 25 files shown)


tseaver:
@tan01 Thank you very much for your work.  Because it will involve breaking changes (e.g., dropping Python2, etc.), we have decided to do the async stuff on a [separate branch](https://github.com/googleapis/google-auth-library-python/tree/async).  Work in  progress there includes:

- [x] #572, "feat: asyncio http request logic and asynchronous credentials logic"
- [ ] #573, " feat: all asynchronous credential types required for all authentication types (service accounts, default... etc)"
- [ ] #574, "feat: added system tests for the asyncIO auth changes"

I am therefore closing this PR with thanks for spurring us on, even if we decided we couldn't use your work directly.

PR #469, opened 2020-03-25
feat: add fetch_id_token to support id_token adc

arithmetic1728:


google/oauth2/id_token.py
```diff
@@ -159,3 +163,90 @@ def verify_firebase_token(id_token, request, audience=None):
...
+    Returns:
+        str: The ID token.
+
+    Raises:
+        ~google.auth.exceptions.DefaultCredentialsError:
+            If metadata server doesn't exist and no valid service account
+            credentials are found.
+    """
+    # 1. First try to fetch ID token from metada server if it exists. The code
+    # works for GAE and cloud run metadata server as well.
```
busunkim96:
```suggestion
    # works for GAE and Cloud Run metadata server as well.
```

PR #473, opened 2020-03-26
feat: Add debug logging that can help with diagnosing auth lib. path

hilts-vaughan:
Sometimes in support, we need to understand where things went wrong in a customers flow after the fact or during it.

It would be nice to have some debug lows to trace this down for people that are less familiar with the flow and would be able to trace it back to here.


busunkim96:
Authors all have valid CLAs

PR #474, opened 2020-03-26
feat: Show the transport exception that happened for GCE Metadata

hilts-vaughan:
Two primary motivators here:

1. Why obscure the exception? Sometimes it has some really useful information, such as the case when its being rate limited or throwing some internal error. Some of those responses have valuable information; and it's a public API so the contents of it are not a secret. 

2. The metadata server should not be spewing failures that often; I think it would be good to know with `WARNING` when it happens since `WARNING` is the default and what a lot of people run with. When Metadata servers have issues overnight, we have no logs for those who left it as default.
A good example of this is when the Metadata server is being rate limited; it would be good to know that's happening. Logs will help make that faster, though ideally of course maybe the library would have a result or something indicating the status of all checks (or something else so faults can be diagnosed faster?) but logs are cheap, easy and make diagnosing it a lot faster. 

It should be rare enough (correct me if I'm wrong?) that it should not be failing often enough except in the "resolution lag" case.


tseaver:
@googlebot I consent.

Issue #471, opened 2020-03-26
 ImportError: No module named 'requests.packages'


AAverin:
After update to latest version we get
```
 from google.auth.transport.requests import AuthorizedSession
  File "/home/travis/.local/lib/python3.4/site-packages/google/auth/transport/requests.py", line 38, in <module>
    from requests.packages.urllib3.util.ssl_ import (
ImportError: No module named 'requests.packages'
```
Were there any changes made to the dependencies of the library?

These is our setup:
```
pip3 install --user google-cloud
pip3 install --user --upgrade pip
pip3 install --user requests
pip3 install --user google-auth
```

busunkim96:
@AAverin Would you mind doing a pip freeze so we can see version #s for other dependencies?

@arithmetic1728 Is there an additional package that needs to be installed for the SSL support?

arithmetic1728:
@AAverin what is the version of your `requests` library? I tried python3.4 on my machine, it worked fine.

```
(tmp) sijunliu@penguin:~/wks/tmp$ python -m pip freeze
DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).                                          
cachetools==3.1.1
certifi==2019.11.28
chardet==3.0.4
google-auth==1.12.0
idna==2.8
pyasn1==0.4.8
pyasn1-modules==0.2.8
requests==2.21.0
rsa==4.0
six==1.14.0
urllib3==1.24.3

(tmp) sijunliu@penguin:~/wks/tmp$ python
Python 3.4.9 (default, Mar 26 2020, 14:10:47) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from google.auth.transport.requests import AuthorizedSession
>>>
```

AAverin:
I can try checking, but because this was on CI and it was important to fix fast, I just rolled back to 1.11.3 and it resolved the issue

@arithmetic1728 Because I don't force specific versions, I would assume that latest version of `requests` library is being installed

arithmetic1728:
@AAverin hmm... I am not sure what causes the problem. I tried python 3.4, 3.6, 3.7, none of them have problems on my machine. I think I will just change the lib to import ssl only when needed then.

@busunkim96 I created a PR to import ssl from `requests` only when needed. Please take a look. Thanks.

[...]

PR #475, opened 2020-03-26
fix: only import ssl if needed

arithmetic1728:
solve #471

### DIFF
--- a/google/auth/transport/requests.py
+++ b/google/auth/transport/requests.py
@@ -35,9 +35,6 @@ except ImportError as caught_exc:  # pragma: NO COVER
     )
 import requests.adapters  # pylint: disable=ungrouped-imports
 import requests.exceptions  # pylint: disable=ungrouped-imports
-from requests.packages.urllib3.util.ssl_ import (
-    create_urllib3_context,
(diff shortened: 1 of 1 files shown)


google/auth/transport/requests.py
```diff
@@ -202,6 +199,7 @@ class _MutualTlsAdapter(requests.adapters.HTTPAdapter):
     def __init__(self, cert, key):
         import certifi
         from OpenSSL import crypto
+        from requests.packages.urllib3.util.ssl_ import create_urllib3_context
```
busunkim96: Is it safe to assume that this import is available at this point?

arithmetic1728: This class needs the `create_urllib3_context` function from the package, so if the import is not available we have to throw exceptions anyway.

`requests.packages.urllib3.util.ssl_` comes from the `requests` package, I don't know why it could be unavailable... This is more likely a `requests` package issue or user's python installation issue to me. Are you able to reproduce that problem?

busunkim96:
Closing as the root cause of the linked issue is most likely a bad dependency

PR #476, opened 2020-03-27
feat: add access token credentials

arithmetic1728:
(1) Add `UserAccessTokenCredentials` class which uses `gcloud auth print-access-token` to obtain user account access token. This token is used to access api endpoint when mtls is enforced.
(2) Remove `google.auth._cloud_sdk.load_authorized_user_credentials` since it is a duplicate of `google.oauth2.credentials.from_authorized_user_info`. Also has to remove this to avoid cyclic import problem.

### DIFF
--- a/google/auth/_cloud_sdk.py
+++ b/google/auth/_cloud_sdk.py
@@ -114,7 +114,7 @@ def get_project_id():
 
 
 def get_auth_access_token(account=None):
-    """Load user access token with the `gcloud auth print-access-token` command.
+    """Load user access token with the ``gcloud auth print-access-token`` command.
 
     Args:
         account (Optional[str]): Account to get the access token for. If not
@@ -144,7 +144,7 @@ def get_auth_access_token(account=None):
 
         access_token = subprocess.check_output(command, stderr=subprocess.STDOUT)
         # remove the trailing "\n"
-        return access_token.decode("utf-8")[:-1]
+        return access_token.decode("utf-8").strip()
     except (subprocess.CalledProcessError, OSError, IOError) as caught_exc:
         new_exc = exceptions.UserAccessTokenError(
             "Failed to obtain access token", caught_exc
--- a/google/auth/_cloud_sdk.py
+++ b/google/auth/_cloud_sdk.py
@@ -18,8 +18,10 @@ import json
 import os
 import subprocess
 
+import six
+
 from google.auth import environment_vars
-import google.oauth2.credentials
+from google.auth import exceptions
 
 
 # The ~/.config subdirectory containing gcloud credentials.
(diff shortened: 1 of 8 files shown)


google/auth/_cloud_sdk.py
```diff
@@ -122,3 +111,42 @@ def get_project_id():
...
+                (command,)
+                + _CLOUD_SDK_USER_ACCESS_TOKEN_COMMAND
+                + ("--account=" + account,)
+            )
+        else:
+            command = (command,) + _CLOUD_SDK_USER_ACCESS_TOKEN_COMMAND
+
+        access_token = subprocess.check_output(command, stderr=subprocess.STDOUT)
+        # remove the trailing "\n"
+        return access_token.decode("utf-8")[:-1]
```
busunkim96: [`.strip()`](https://docs.python.org/3.6/library/stdtypes.html?highlight=strip#str.strip) will remove leading and trailing whitespace

google/oauth2/credentials.py
```diff
@@ -292,3 +293,50 @@ def to_json(self, strip=None):
             prep = {k: v for k, v in prep.items() if k not in strip}
 
         return json.dumps(prep)
+
+
+class UserAccessTokenCredentials(credentials.Credentials):
+    """Access token credentials for user account.
+
+    Obtain the access token for a given user account or the current active
+    usr account with the `gcloud auth print-access-token` command.
```
busunkim96:
```suggestion
    user account with the ``gcloud auth print-access-token`` command.
```

google/auth/_cloud_sdk.py
```diff
@@ -122,3 +111,42 @@ def get_project_id():
         return configuration["configuration"]["properties"]["core"]["project"]
     except KeyError:
         return None
+
+
+def get_auth_access_token(account=None):
+    """Load user access token with the `gcloud auth print-access-token` command.
```
busunkim96:
```suggestion
    """Load user access token with the ``gcloud auth print-access-token`` command.
```

google/oauth2/credentials.py
```diff
@@ -292,3 +293,50 @@ def to_json(self, strip=None):
...
+        Args:
+            account (str): Account to get the access token for.
+
+        Returns:
+            google.oauth2.credentials.UserAccessTokenCredentials: The created
+                credentials with the given account.
+        """
+        return self.__class__(account=account)
+
+    def refresh(self, request):
```
busunkim96: Is it possible to default this to None, or does this break something?
```suggestion
    def refresh(self, request=None):
```

google/auth/exceptions.py
```diff
@@ -28,5 +28,9 @@ class RefreshError(GoogleAuthError):
     failed."""
 
 
+class UserAccessTokenError(GoogleAuthError):
+    """Used to indicate `gcloud auth print-access-token` command failed."""
```
busunkim96:
```suggestion
    """Used to indicate ``gcloud auth print-access-token`` command failed."""
```

google/auth/_cloud_sdk.py
arithmetic1728: done

google/oauth2/credentials.py
arithmetic1728: done

google/auth/_cloud_sdk.py
arithmetic1728: done

google/oauth2/credentials.py
arithmetic1728: this function override the abstract method defined in the [base class](https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/credentials.py#L79), so `request` cannot be default to None.

google/auth/exceptions.py
arithmetic1728: done

PR #480, opened 2020-04-01
feat: consolidate mTLS channel errors

arithmetic1728:


google/auth/exceptions.py
```diff
@@ -34,3 +34,7 @@ class UserAccessTokenError(GoogleAuthError):
 
 class DefaultCredentialsError(GoogleAuthError):
     """Used to indicate that acquiring default credentials failed."""
+
+
+class MutualTlsChannelError(GoogleAuthError):
```
busunkim96, line 39: Per PEP8, please rename to `MutualTLSChannelError`

https://www.python.org/dev/peps/pep-0008/

> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError

arithmetic1728: done.

PR #481, opened 2020-04-01
fix: invalid expiry type

### LINKED ISSUE #479: Invalid type for expiry when using GCE metadata server
#### Environment details

  - OS: Linux
  - Python version: 3.8.2
  - pip version: 20.0.2
  - `google-auth` version: 1.12.0

#### Steps to reproduce

  1. Create an ID token using the GCE metadata server:
      ```python
      from google.auth import compute_engine
      target_audience = 'https://example.com'
      creds = compute_engine.IDTokenCredentials(request, target_audience, use_metadata_identity_endpoint=True)
      ```
  2. Try to use the credentials to authorize a request:
      ```python
      from google.auth.transport.requests import AuthorizedSession
      session = AuthorizedSession(credentials)
      response = session.get('https://example.com')
      ```

This results in an error:
```
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/site-packages/google/auth/credentials.py", line 66, in expired
    skewed_expiry = self.expiry - _helpers.CLOCK_SKEW
TypeError: unsupported operand type(s) for -: 'int' and 'datetime.timedelta'
```

I think the issue is on [this line](https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/compute_engine/credentials.py#L294). `payload["exp"]` should (apparently) be getting parsed into a `datetime.timedelta`

arithmetic1728:
Fix #479 

`self.expiry` should be a `datetime.datetime` object, treated similar as https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/impersonated_credentials.py#L352

### DIFF
--- a/google/auth/compute_engine/credentials.py
+++ b/google/auth/compute_engine/credentials.py
@@ -275,8 +275,7 @@ class IDTokenCredentials(credentials.Credentials, credentials.Signing):
                 HTTP requests.
 
         Returns:
-            str: The ID token.
-            datetime.datetime: Expiry of the ID token.
+            Tuple[str, datetime.datetime]: The ID token and the expiry of the ID token.
 
         Raises:
             google.auth.exceptions.RefreshError: If the Compute Engine metadata
--- a/google/auth/compute_engine/credentials.py
+++ b/google/auth/compute_engine/credentials.py
@@ -274,6 +274,10 @@ class IDTokenCredentials(credentials.Credentials, credentials.Signing):
             request (google.auth.transport.Request): The object used to make
                 HTTP requests.
 
+        Returns:
+            str: The ID token.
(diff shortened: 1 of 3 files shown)


google/auth/compute_engine/credentials.py
```diff
@@ -274,6 +274,10 @@ def _call_metadata_identity_endpoint(self, request):
             request (google.auth.transport.Request): The object used to make
                 HTTP requests.
 
+        Returns:
+            str: The ID token.
+            datetime.datetime: Expiry of the ID token.
+
```
busunkim96, line 280:
```suggestion
        Returns:
            Tuple[str, datetime.datetime]: The ID token and the expiry of the ID token.

```

PR #485, opened 2020-04-07
fix a potential json parsing error

ziliangpeng:
As discussed [here](https://github.com/googleapis/google-auth-library-python/pull/387), there's a chance (which I have experienced) that the response of the oauth request is not in json format, resulting in json parsing error.

This PR fixes it by capturing the exception and convert to a `RefreshError`.

### DIFF
--- a/google/oauth2/_client.py
+++ b/google/oauth2/_client.py
@@ -108,7 +108,13 @@ def _token_endpoint_request(request, token_uri, body):
             if hasattr(response.data, "decode")
             else response.data
         )
-        response_data = json.loads(response_body)
+        try:
+            response_data = json.loads(response_body)
+        except (KeyError, ValueError):
+            # raise if the response can't be json decoded.
+            error_details = response_body
+            raise exceptions.RefreshError(error_details, response_body)
+
 
         if response.status == http_client.OK:
             break


ziliangpeng:
@georgysavva @busunkim96

georgysavva:
Why not to just move `json.loads()` after the `http_client.OK` check. As it was before my change?

ziliangpeng:
@georgysavva The decoded json needs to be used in the error handling as well as being a return value.

I believe keeping it here is the most appropriate place to decode it. Where else exactly do you see is a better place?

alvyjudy:
Hi, please correct me if I'm wrong but it appears that this problem (response not in json format) kind of occur by chance? (I get this impression after reading the discussion you mentioned)

So when it arises wouldn't it be a better idea to re-fetch, rather than giving the error back to the client? I would imagine the error message to be quite obscure. 

Also, if it is indeed the API that wouldn't return a json response, would it be better to place your proposed functionality inside ``_handle_error_message``? I feel that would make the code look cleaner.

google/oauth2/_client.py
```diff
@@ -108,7 +108,13 @@ def _token_endpoint_request(request, token_uri, body):
             if hasattr(response.data, "decode")
             else response.data
         )
-        response_data = json.loads(response_body)
+        try:
+            response_data = json.loads(response_body)
+        except (KeyError, ValueError):
```
mik-laj, line 113:
```suggestion
        except json.DecodeError
```

tseaver: That would be a Python3-only change, i think.

parthea:
Hi @ziliangpeng , I'm going to close this PR due to inactivity but please feel free to re-open it.

PR #488, opened 2020-04-15
fix(deps): use mock from stdlib in tests

maxwell-k:
This change removes an unnecessary dependency in the tests.

The [mock package] is a backport of functionality from the Python 3.3 standard library to Python 2.7. This library is [not supported] on Python 2.7. Using `import mock` instead of `from unittest import mock` adds an unnecessary dependency, for no apparent gain. 

[mock package]: https://pypi.org/project/mock/
[not supported]: https://github.com/googleapis/google-auth-library-python#deprecated-python-versions


tseaver:
@maxwell-k Thanks for the patch.  However, Python 2.7 support is deprecated, but still present.  Until we drop it entirely (including from CI / nox), we cannot rely on the stdlib to have `unittest.mock`.

tseaver:
@maxwell-k I'm sorry not to be able to merge this PR.  While working on #778, I found that we cannot drop the `mock` compatibility library, because we rely on `mock.AsyncMock`, which is not present in the stdlib version before Python 3.8.

PR #490, opened 2020-04-21
fix: wrong es256 signature format

arithmetic1728:
es256 signature should have the raw format instead of asn1 encoded format. Raw format is base64url encoded bytes array (64 bytes) of two integers r, s (each contributes 32 bytes to the array).

Verified the fix works with iap issued id_token in PR: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3444

### DIFF
--- a/google/auth/crypt/es256.py
+++ b/google/auth/crypt/es256.py
@@ -15,12 +15,15 @@
 """ECDSA (ES256) verifier and signer that use the ``cryptography`` library.
 """
 
+from cryptography import utils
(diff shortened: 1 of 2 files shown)


busunkim96:
@arithmetic1728 Thanks for fixing this!

Would you mind tweaking the PR title to be slightly more descriptive? The title is used directly in the changelog.

arithmetic1728:
@busunkim96 Updated the title and commit message to include more details.

Issue #494, opened 2020-05-04
Credentials to_json method could have clearer documentation and API


adelevie:
**Is your feature request related to a problem? Please describe.**

I need to serialize the `Credentials` data as JSON, and I just discovered the `to_json` method. It's very helpful, as the pickle-related techniques in the quickstart (e.g. https://developers.google.com/docs/api/quickstart/python) rely on the file system and binary data, which means the token cannot be persisted in an environment variable (useful for Heroku and other server deployments).

Anyways, the `to_json` method has the following in the doc string:

```
        Returns:
            str: A JSON representation of this instance, suitable to pass to
                 from_json().
```

I looked and looked for `from_json()` and couldn't find it in this library. If it does exist, I can't find it, and maybe should be easier to find. Assuming it doesn't exist, I found (mostly) what I needed in the class method `from_authorized_user_info`:

```python
    if os.path.exists('token.json'):
        with open('token.json', 'r') as token:
            token_file = token.read()
            token_json = json.loads(token_file)
            creds = Credentials.from_authorized_user_info(token_json, scopes=SCOPES)
            creds.token = token_json['token']
```

That last line is used to ensure `creds.valid` returns `True`.

 **Describe the solution you'd like**

The documentation and/or API for serializing/de-serializing as JSON should be clearer. If there is a `from_json` method, it should be mentioned somewhere in the docs. If there isn't, the doc string in `to_json` should specify the (class) method most suitable to ingest the serialized JSON. If that method happens to be `from_authorized_user_info`, it should ingest the `token` key as well, avoiding the need for the sort of hacky value set after instantiation/building.

 **Describe alternatives you've considered**

The alternative I considered is the code sample for my workaround mentioned earlier in this issue.

 **Additional context**

None I can think of.

alvyjudy:
The ``from_authorized_user_file`` seems to provide that functionality. It reads a json file and returns a credential instance by calling ``from_authorized_user_info``:

```
def from_authorized_user_file(cls, filename, scopes=None):
    """Creates a Credentials instance from an authorized user json file."""
    with io.open(filename, "r", encoding="utf-8") as json_file:
        data = json.load(json_file)
        return cls.from_authorized_user_info(data, scopes)
```

I agree that the documentation should be clearer, though

adelevie:
Thanks for the quick note, @alvyjudy.

> Is it a good idea to just replace from_json with from_authorized_user_file in the docstring?

[...]

PR #495, opened 2020-05-04
update docstring of to_json method

alvyjudy:
from #494 
The original docstring of ``to_json`` referenced a method called ``from_json`` which actually exists in the form of ``from_authorized_user_info``. I also added some explanation for the functionality provided by ``from_authorized_user_info``.


adelevie:
One minor thing, maybe worth pointing out, is `to_json` returns a string, while `from_authorized_user_info` expects a dictionary.

alvyjudy:
I might have messed up a few commit messages and formatting so I'll just close this and make another one:(

PR #503, opened 2020-05-06
chore: support string type response.data for gcloud

arithmetic1728:
gcloud uses httplib2 for the underlying http where response.data is string instead of bytes.

### DIFF
--- a/google/auth/impersonated_credentials.py
+++ b/google/auth/impersonated_credentials.py
@@ -88,13 +88,17 @@ def _make_iam_token_request(request, principal, headers, body):
 
     response = request(url=iam_endpoint, method="POST", headers=headers, body=body)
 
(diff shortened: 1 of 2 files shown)


arithmetic1728:
> Would you mind changing this to a `fix` since it changes the implementation?

Ah, sorry I didn't notice your comment.

busunkim96:
@arithmetic1728 That's alright! You won't see release-please open a PR (it does patch bumps for `fix`, minor bumps for `feat`, and major bumps for breaking changes). 

If this is urgent you can open a new PR and add the following to the commit body:

`Release-As: 1.14.2`

arithmetic1728:
@busunkim96 gcloud has its own google-auth copy in g3, so I just need to update in github and pull it into gcloud. I think I don't need to release it this time. Thank you!

PR #519, opened 2020-06-02
feat: add quota_project_id to service accounts; add with_quota_project methods

busunkim96:
Adds `quota_project_id` to service account credentials, making it possible to set `quota_project_id` on OAuth2 credentials and service account credentials.

This PR also adds the method `with_quota_project` to both classes.

### DIFF
--- a/google/oauth2/service_account.py
+++ b/google/oauth2/service_account.py
@@ -333,7 +333,6 @@ class Credentials(credentials.Signing, credentials.Scoped, credentials.Credentia
             additional_claims=self._additional_claims.copy(),
         )
 
-
     def _make_authorization_grant_assertion(self):
         """Create the OAuth 2.0 assertion.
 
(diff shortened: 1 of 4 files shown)


busunkim96:
@broady @bshaffer I *think* I should expand this PR so all credentials produced by the default auth flow have `quota_project` (+ App Engine and Compute Engine credentials).

Is it appropriate to add it to the base credentials class? I'm curious what other languages did.

google/oauth2/service_account.py
```diff
@@ -335,6 +373,12 @@ def refresh(self, request):
         self.token = access_token
         self.expiry = expiry
 
+    @_helpers.copy_docstring(credentials.Credentials)
+    def apply(self, headers, token=None):
+        super(Credentials, self).apply(headers, token=token)
```
software-dov, line 378: Shouldn't this come after the headers manipulation?

busunkim96: I think either order works - the `headers` object is passed into the function and modified.

https://github.com/googleapis/google-auth-library-python/blob/37141e4dffc2ba3f3f57c5914544fb8b9cf7d017/google/auth/credentials.py#L94-L104

PR #528, opened 2020-06-11
fix: narrow acceptable RSA versions to maintain Python 2 compatability

crwilcox:
Version 4.0 was the last version to support Python 2 and 3.4. Version 4.1 is compatible with Python 3.5+ only.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ from setuptools import setup
 DEPENDENCIES = (
     "cachetools>=2.0.0,<5.0",
     "pyasn1-modules>=0.2.1",
-    "rsa>=3.1.4,<5.0",
+    "rsa>=3.1.4,<4.1.0",
     "setuptools>=40.3.0",
     "six>=1.9.0",
 )
--- a/setup.py
+++ b/setup.py
@@ -22,8 +22,8 @@ DEPENDENCIES = (
     "cachetools>=2.0.0,<5.0",
     "pyasn1-modules>=0.2.1",
     # rsa 4.1, 4.1.1, 4.2 are broken on Py2: https://github.com/sybrenstuvel/python-rsa/issues/152
-    'rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version < 3',
-    'rsa>=3.1.4,<5; python_version >= 3'
+    "rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version < 3",
+    "rsa>=3.1.4,<5; python_version >= 3",
     "setuptools>=40.3.0",
     "six>=1.9.0",
 )
--- a/setup.py
+++ b/setup.py
@@ -21,8 +21,9 @@ from setuptools import setup
 DEPENDENCIES = (
     "cachetools>=2.0.0,<5.0",
     "pyasn1-modules>=0.2.1",
-    # TODO: rsa >= 4.1.0 doesn't support Python 2. Remove pinning after dropping support for Python 2. 
-    "rsa>=3.1.4,<4.1.0",
+    # rsa 4.1, 4.1.1, 4.2 are broken on Py2: https://github.com/sybrenstuvel/python-rsa/issues/152
+    'rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version < 3',
+    'rsa>=3.1.4,<5; python_version >= 3'
     "setuptools>=40.3.0",
     "six>=1.9.0",
 )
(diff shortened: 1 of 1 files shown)


crwilcox:
Also, related to https://issues.apache.org/jira/browse/BEAM-10244

crwilcox:
Also https://github.com/sybrenstuvel/python-rsa/issues/152

setup.py
```diff
@@ -21,7 +21,7 @@
 DEPENDENCIES = (
     "cachetools>=2.0.0,<5.0",
     "pyasn1-modules>=0.2.1",
-    "rsa>=3.1.4,<5.0",
+    "rsa>=3.1.4,<4.1.0",
```
mik-laj, line 24:
```suggestion
    # TODO: rsa v4.10 doesn't support Python 2. Remove pinning after dropping support for Python 2. 
    "rsa>=3.1.4,<4.1.0",
```

tvalentyn: Consider the following instead:

# rsa 4.1, 4.1.1, 4.2 are broken on Py2: https://github.com/sybrenstuvel/python-rsa/issues/152
'rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version < 3',
'rsa>=3.1.4,<5; python_version >= 3',

crwilcox: Sure. modified your comment slightly and accepted.

tvalentyn: Reopening just to make sure you saw my comment before this was resolved.

setup.py
```diff
@@ -21,7 +21,9 @@
 DEPENDENCIES = (
     "cachetools>=2.0.0,<5.0",
     "pyasn1-modules>=0.2.1",
-    "rsa>=3.1.4,<5.0",
+    # rsa 4.1, 4.1.1, 4.2 are broken on Py2: https://github.com/sybrenstuvel/python-rsa/issues/152
+    'rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version < 3',
+    'rsa>=3.1.4,<5; python_version >= 3'
```
busunkim96, line 26:
```suggestion
    # rsa 4.1, 4.1.1, 4.2 are broken on Py2: https://github.com/sybrenstuvel/python-rsa/issues/152
    "rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version < 3",
    "rsa>=3.1.4,<5; python_version >= 3",
```

setup.py
```diff
@@ -22,8 +22,8 @@
     "cachetools>=2.0.0,<5.0",
     "pyasn1-modules>=0.2.1",
     # rsa 4.1, 4.1.1, 4.2 are broken on Py2: https://github.com/sybrenstuvel/python-rsa/issues/152
-    'rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version < 3',
-    'rsa>=3.1.4,<5; python_version >= 3'
+    "rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version < 3",
```
tvalentyn, line 25: I think "3" may need to be in quotes as in:
`'rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2,<5; python_version<"3"'`
`'rsa>=3.1.4,<5; python_version>="3"'`

crwilcox: Currently fixing and testing locally

sybrenstuvel:
I fixed it in the Python-RSA package so that package managers now know which version they can use. Python-RSA 4.3 is now the last one to support Python 2.7, version 4.4 explicitly requires Python 3.5+.

Hope this is now resolved for you, if there are still issues let me know.

tvalentyn:
> version 4.4 explicitly requires Python 3.5+.
Actually, 4.4 is still installable on Py2. Perhaps intended? We can continue this on https://github.com/sybrenstuvel/python-rsa/issues/152.

setup.py
tvalentyn: My motivation for rsa>=3.1.4,!=4.1,!=4.1.1,!=4.2  was that there maybe another Py2 release (which we had) with additional fixes (there were security backports), and the range would be able to pick it up.
I also was hoping that new releases won't be broken on Py2, however, looks like 4.4 is installable and broken on Py2, so we'd have to exclude it.

busunkim96: Narrowed in in #532 

https://github.com/googleapis/google-auth-library-python/blob/b74168b710da38c0cf06cdf8685c49af4748e3a0/setup.py#L25-L26

busunkim96: Released `1.17.2` with the pin above

PR #530, opened 2020-06-12
feat: make ``load_credentials_from_file`` a public method

busunkim96:
Turns `_load_credentials_from_file` into a public method `load_credentials_from_file`. Also allow scopes to be passed.

```py
import google.auth

credentials, project_id = google.auth.load_credentials_from_file(
    "path/to/credentials.json",
    scopes=["https://www.google.com/calendar/feeds"]
)
```

On second thought, I'm not sure exposing this method this way is appropriate. Might be deceptive given that it isn't a magical method that can load all credentials 🤔

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -77,7 +77,7 @@ def load_credentials_from_file(filename, scopes=None):
 
     Args:
         filename (str): The full path to the credentials file.
-        scopes (Sequence[str]): The list of scopes for the credentials. If
+        scopes (Optional[Sequence[str]]): The list of scopes for the credentials. If
             specified, the credentials will automatically be scoped if
             necessary.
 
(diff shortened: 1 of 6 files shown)


google/auth/_default.py
```diff
@@ -69,14 +69,17 @@ def _warn_about_problematic_credentials(credentials):
...
-    """Loads credentials from a file.
+def load_credentials_from_file(filename, scopes=None):
+    """Loads Google credentials from a file.
 
     The credentials file must be a service account key or stored authorized
     user credentials.
 
     Args:
         filename (str): The full path to the credentials file.
+        scopes (Sequence[str]): The list of scopes for the credentials. If
```
software-dov, line 80: Should there also be an `Optional`, as in `Optional[Sequence[str]]`? I can't remember exactly how the syntax works.

google/auth/_default.py
```diff
@@ -105,37 +108,41 @@ def _load_credentials_from_file(filename):
...
+                new_exc = exceptions.DefaultCredentialsError(msg, caught_exc)
+                six.raise_from(new_exc, caught_exc)
+            return credentials, info.get("project_id")
+
+    raise exceptions.DefaultCredentialsError(
+        "The file {file} does not have a valid type. "
+        "Type is {type}, expected one of {valid_types}.".format(
+            file=filename, type=credential_type, valid_types=_VALID_TYPES
         )
+    )
```
software-dov, line 145: The control flow is a little tricky to eyeball grep. It may be easier if it were changed to be something like
```python
if credential_type is None:
    raise exceptions.DefaultCredentialsError(...)
# dedented control flow
```

arithmetic1728:
@busunkim96 This happens to be exactly what I need for my gcloud auth file override work (where gcloud reads cred json files to create google-auth creds)

arithmetic1728:
@software-dov Please take another look, thanks!

PR #545, opened 2020-06-24
feat: asyncio aiohttp functionality

AniBadde:
Implemented asyncio http requests functionality in the auth library. All components tested except for refreshing credentials in request() function in AuthorizedSession() class in aiohttp_req.py. 

To be resolved:
- Increase coverage from 99% to 100%
- Adding test for credential refresh in request coroutine in AuthorizedSession() class - contained in aiohttp_req.py
- Only functional for Python 3.6, 3.7 due to aiohttp compatibility.


crwilcox:
Looks like there are a few failures, for instance, looks like black should be run. If you run `nox` locally it should run all the necessary checks and fix up formatting. I also saw what looks like a test failure.

```
black --check google tests noxfile.py setup.py docs/conf.py
would reformat /tmpfs/src/github/google-auth-library-python/google/oauth2/_client.py
would reformat /tmpfs/src/github/google-auth-library-python/tests/test_credentials.py
would reformat /tmpfs/src/github/google-auth-library-python/tests/transport/test_aiohttp_req.py
All done! 💥 💔 💥
3 files would be reformatted, 72 files would be left unchanged.
Command black --check google tests noxfile.py setup.py docs/conf.py failed with exit code 1
Session lint failed.
```

PR #546, opened 2020-06-24
feat: add quota project to base credentials class

busunkim96:
Make `quota_project_id` a property of the base credentials class and add it to the existing credentials types. 

Note that I didn't make `with_quota_project` an abstract method on the base credentials class because that would be a breaking change.

This PR also modifies functions in `default` to allow a `quota_project_id` to be passed.

```py
import google.auth

credentials, project = google.auth.default(quota_project_id='my-project')
```

```py
import google.auth

credentials, project = google.auth.load_credentials_from_file(
    'service_account.json',
    quota_project_id='my-project'
)
```

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -317,9 +317,7 @@ def default(scopes=None, request=None, quota_project_id=None):
     for checker in checkers:
         credentials, project_id = checker()
         if credentials is not None:
-            credentials = with_scopes_if_required(credentials, scopes)
-            if quota_project_id:
-                credentials = credentials.with_quota_project_id(quota_project_id)
+            credentials = with_scopes_if_required(credentials, scopes).with_quota_project_id(quota_project_id)
             effective_project_id = explicit_project_id or project_id
             if not effective_project_id:
                 _LOGGER.warning(
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -317,7 +317,9 @@ def default(scopes=None, request=None, quota_project_id=None):
     for checker in checkers:
         credentials, project_id = checker()
         if credentials is not None:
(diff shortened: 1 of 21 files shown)


hiranya911:
Isn't this a breaking change? Should it have been released as 1.19.0?

We are already seeing test failures in downstream libs like `firebase-admin` where we extend this interface.

hiranya911:
Actually this seems to be because pylint considers any method in an abstract class that throws NotImplementedError to be abstract: http://pylint-messages.wikidot.com/messages:w0223

Not quite sure if that should be considered a breaking change, or write it off as pylint being too pedantic.

busunkim96:
@hiranya911 

Hmm, maybe it makes more sense to have the function raise `ValueError`?

```py
    def with_quota_project(self, quota_project_id):
        raise ValueError("Credentials don't support quota project.")
```

google/auth/_default.py
```diff
@@ -114,7 +116,7 @@ def load_credentials_from_file(filename, scopes=None):
         try:
             credentials = credentials.Credentials.from_authorized_user_info(
                 info, scopes=scopes
-            )
+            ).with_quota_project(quota_project_id)
```
mapler, line 119: Can not find anywhere that `load_credentials_from_file` is called with a `quota_project_id` parameter.  
So `quota_project_id` seems always be the default `None`.  
At the same time, `quota_project_id` loaded in `info` (from json file) could have work well but overwrote by the `.with_quota_project(quota_project_id)`

busunkim96: Ah yes, it looks like I introduced a bug. Will open a PR shortly. Thanks for the review!

hiranya911:
@busunkim96 thanks for looking into it. Yes, perhaps the default implementation in the abstract class should raise `ValueError`.

PR #547, opened 2020-06-24
feat: asyncio aiohttp functionality

anibadde:
Implemented asyncio http requests functionality in the auth library. All components tested except for refreshing credentials in request() function in AuthorizedSession() class in aiohttp_req.py.

To be resolved:

Increase coverage from 97% to 100%
Adding test for credential refresh in request coroutine in AuthorizedSession() class - contained in aiohttp_req.py
Only functional for Python 3.5, 3.6, 3.7 due to aiohttp compatibility.


crwilcox:
It seems like 2.7 is failing and coverage is low.

AniBadde:
Requires tests for aiohttp_req lines 376-399 amd 421-430
credentials_async, _default_async contains duplicated code from the corresponding sync versions.

cert_path
```diff
@@ -0,0 +1 @@
+cert
```
crwilcox, line 1: is this file required?

google/auth/__init__.py
```diff
@@ -17,9 +17,10 @@
 import logging
 
 from google.auth._default import default, load_credentials_from_file
+from google.auth._default_async import default_async
```
crwilcox, line 20: are we reusing load_credentials_from_file? Could that be added to async?

google/auth/_default_async.py
```diff
@@ -0,0 +1,330 @@
...
+Could not automatically determine credentials. Please set {env} or \
+explicitly create credentials and re-run the application. For more \
+information, please see \
+https://cloud.google.com/docs/authentication/getting-started
+""".format(
+    env=environment_vars.CREDENTIALS
+).strip()
+
+# Warning when using Cloud SDK user credentials
+_CLOUD_SDK_CREDENTIALS_WARNING = """\
```
crwilcox, line 52: Assuming this constant matches the sync one, could we import these to each class?

google/auth/_default_async.py
```diff
@@ -0,0 +1,330 @@
...
+Your application has authenticated using end user credentials from Google \
+Cloud SDK without a quota project. You might receive a "quota exceeded" \
+or "API not enabled" error. We recommend you rerun \
+`gcloud auth application-default login` and make sure a quota project is \
+added. Or you can use service accounts instead. For more information \
+about service accounts, see https://cloud.google.com/docs/authentication/"""
+
+
+def _warn_about_problematic_credentials(credentials):
```
crwilcox, line 61: nothing in this is specifically async as far as I can tell? If copy/paste can we refactor? (similar comment throughout)

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,440 @@
...
+        return response
+
+    @property
+    def is_mtls(self):
+        """Indicates if the created SSL channel is mutual TLS."""
+        return self._is_mtls
+
+
+async def main():
+    # breakpoint()
```
crwilcox, line 425: nit: remove comment out code

google/oauth2/_client.py
```diff
@@ -257,3 +257,36 @@ def refresh_grant(
     expiry = _parse_expiry(response_data)
 
     return access_token, refresh_token, expiry, response_data
+
+
+'''
+async version of refresh token grant implemmented above:
```
crwilcox, line 263: remove or add `TODO(anibadde):` comment

noxfile.py
```diff
@@ -64,10 +90,24 @@ def blacken(session):
...
+        "--cov=google.auth",
+        "--cov=google.oauth2",
+        "--cov=tests",
+        "tests",
+        "tests_async",
+    )
+
+
+@nox.session(python="2.7")
+def unit2(session):
```
crwilcox, line 108: maybe name `unit-2-7` to better match the names made for the other unit test run?

setup.py
```diff
@@ -26,6 +26,8 @@
     'rsa>=3.1.4,<5; python_version >= "3"',
     "setuptools>=40.3.0",
     "six>=1.9.0",
+    "aiohttp>=3.6.2; python_version>='3.5'",
```
crwilcox, line 29: is this 3.5 truly? I know there are some reasons to do 3.6+ only for async

google/auth/_default_async.py
tritone: +1 to this. It would be helpful to see a diff of this file vs. _default.py and then we can give guidance on refactoring if you need it.

google/auth/credentials_async.py
```diff
@@ -0,0 +1,347 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: Would also be helpful to see a diff of this file vs. credentials.py. (I don't think there's a super straightforward way to do this in git/github unfortunately unless you go back and make the initial commit just a copy of the file and then stack subsequent commits on that, but you could also just take a diff and send it to us separately).

tests/oauth2/test__client.py
```diff
@@ -90,7 +90,7 @@ def test__token_endpoint_request():
     result = _client._token_endpoint_request(
         request, "http://example.com", {"test": "params"}
     )
-
+    # breakpoint()
```
tritone, line 93: Remove all these.

AniBadde:
The only difference between credentials.py and credentials_async.py is the before_request method, which required async functionality in the implementation. However, this can be changed as well if needed.

google/auth/credentials_async.py
```diff
@@ -0,0 +1,331 @@
...
+    def __init__(self):
+        self.token = None
+        """str: The bearer token that can be used in HTTP headers to make
+        authenticated requests."""
+        self.expiry = None
+        """Optional[datetime]: When the token expires and is no longer valid.
+        If this is None, the token is assumed to never expire."""
+
+    '''
+    @property
```
tritone, line 54: I'm wondering how this works with `expired` and `valid` commented out. @crwilcox does this make sense to you?

tests/test_credentials.py
```diff
@@ -57,6 +57,34 @@ def test_expired_and_valid():
     assert credentials.expired
 
 
+"""
+@pytest.mark.asyncio
```
tritone, line 61: Why is this commented out for now?

tests_async/test_aiohttp_req.py
```diff
@@ -0,0 +1,226 @@
...
+                credentials
+            )
+            response = await authed_session.request("GET", "http://example.com")
+            assert response.status == 200
+            assert credentials.async_before_request.called
+            assert not credentials.refresh.called
+
+            await authed_session.close()
+
+    @pytest.mark.asyncio
```
tritone, line 213: Nice, glad this approach ended up working!

tests_async/test_aiohttp_req.py
```diff
@@ -0,0 +1,226 @@
...
+            mocked.get("http://example.com", status=200)
+            authed_session = google.auth.transport.aiohttp_req.AuthorizedSession(
+                credentials
+            )
+            response = await authed_session.request("GET", "http://example.com")
+            assert response.status == 200
+            assert credentials.async_before_request.called
+            assert not credentials.refresh.called
+
+            await authed_session.close()
```
tritone, line 211: Is closing the session required for cleanup? If so, what happens when an assertion fails? Is there a way to handle this via context managers or some pytest wizardry?

tests/transport/compliance.py
```diff
@@ -61,14 +61,15 @@ def wait():
         yield server
         server.stop()
 
+    @pytest.mark.asyncio
```
tritone, line 64: Does adding this mark here mean that these tests are no longer running on the sync version?

google/auth/credentials_async.py
```diff
@@ -0,0 +1,331 @@
...
+        """
+        # pylint: disable=unused-argument
+        # (Subclasses may use these arguments to ascertain information about
+        # the http request.)
+
+        self.apply(headers)
+
+
+# We comment out the following as it is not called in our aiohttp file, but is used in other credential requirements.
+# May require future usage depending on use case, but not tested in sync requests module.
```
tritone, line 131: I'm curious about the implications of this. Does this mean that callers of async methods cannot use anonymous credentials or specific scopes, for example? Or are these things totally decoupled from each other?

crwilcox:
Should this be unmarked as a draft?

google/auth/credentials_async.py
crwilcox: +1. Pretty much anything in here that is commented out for future needed work should either be done in this pr, or have an issue explaining it.

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,414 @@
...
+
+
+_OAUTH_SCOPES = [
+    "https://www.googleapis.com/auth/appengine.apis",
+    "https://www.googleapis.com/auth/userinfo.email",
+]
+
+_LOGGER = logging.getLogger(__name__)
+
+_DEFAULT_TIMEOUT = 120  # in seconds
```
crwilcox, line 42: is this the same timeout as async? if so we should avoid copying instead sharing timeouts.

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,414 @@
...
+        self._timeout = timeout
+        self.remaining_timeout = timeout
+        self._timeout_error_type = timeout_error_type
+
+    def __enter__(self):
+        self._start = time.time()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if exc_value:
```
crwilcox, line 93: this seems to take from transport/request.py. can any of the code be shared as to not copy/paste?

Also, filename is `aiohttp_req.py`. Generally it is best to not abbreviate, but it can be acceptable. Is there a reason to abbreviate request?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,414 @@
...
+        self._refresh_lock = asyncio.Lock()
+
+        if auth_request is None:
+            self._auth_request_session = aiohttp.ClientSession()
+
+            # Using an adapter to make HTTP requests robust to network errors.
+            # This adapter retrys HTTP requests when network errors occur
+            # and the requests seems safely retryable.
+
+            # retry_adapter = requests.adapters.HTTPAdapter(max_retries=3)
```
crwilcox, line 268: There are a lot of comments in here. I feel that if this is helpful, this may need to be pulled up to docs? Or authored as such?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,414 @@
...
+
+        # Use a kwarg for this instead of an attribute to maintain
+        # thread-safety.
+        _credential_refresh_attempt = kwargs.pop("_credential_refresh_attempt", 0)
+        # Make a copy of the headers. They will be modified by the credentials
+        # and we want to pass the original headers if we recurse.
+        request_headers = headers.copy() if headers is not None else {}
+
+        # Do not apply the timeout unconditionally in order to not override the
+        # _auth_request's default timeout.
```
crwilcox, line 339: I see in a few places you have 'floating' comments, ones with whitespace on either side. I thought the linter enforced that this couldn't happen, but maybe not. It is more typical to have comments directly above the line they describe.

google/oauth2/_client.py
```diff
@@ -257,3 +257,38 @@ def refresh_grant(
     expiry = _parse_expiry(response_data)
 
     return access_token, refresh_token, expiry, response_data
+
+
+'''
+TODO(anibadde)
```
crwilcox, line 263: If you plan to do this as part of this PR, likely fine, but if you want to come back to this (if that is doable), we should likely open an issue to track the work.

key_path
```diff
@@ -0,0 +1 @@
+key
```
crwilcox, line 1: is this file required?

noxfile.py
```diff
@@ -64,10 +90,24 @@ def blacken(session):
     session.run("black", *BLACK_PATHS)
 
 
-@nox.session(python=["2.7", "3.5", "3.6", "3.7"])
+@nox.session(python=["3.6", "3.7"])
```
crwilcox, line 93: I would add 3.8 here

tests_async/test_aiohttp_req.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def test_tracks_elapsed_time_w_tuple_timeout(self, frozen_time):
+        with self.make_guard(timeout=(16, 19)) as guard:
+            frozen_time.tick(delta=datetime.timedelta(seconds=3.8))
+        assert guard.remaining_timeout == (12.2, 15.2)
+
+    def test_noop_if_no_timeout(self, frozen_time):
+        with self.make_guard(timeout=None) as guard:
+            frozen_time.tick(delta=datetime.timedelta(days=3650))
+        # NOTE: no timeout error raised, despite years have passed
```
crwilcox, line 84: ?

google/auth/transport/aiohttp_req.py
AniBadde: The asyncio timeout exception is different, so we're forced to write a new Timeout Guard to deal with that

google/auth/credentials_async.py
AniBadde: Resolved

tests/transport/compliance.py
AniBadde: Resolved

tests/test_credentials.py
AniBadde: Resolved

google/auth/credentials_async.py
AniBadde: Resolved

tests/oauth2/test__client.py
AniBadde: Resolved

google/auth/credentials_async.py
AniBadde: Resolved (refactored to remove credentials_async file - did not need async/await on refreshing credentials)

setup.py
AniBadde: Resolved, changed to 3.6+

noxfile.py
AniBadde: Resolved to include both python 3.5 and 2.7 and changed name accordingly

tests_async/test_aiohttp_req.py
AniBadde: Yes, the closing session was required to prevent the warning message saying that there is a loose session that is not closed. The tests still pass when we remove this, but was done to remove the warning message. If the assertion fails then an error is thrown before it reaches the end of the method (and hence the session.close line), so I assumed that would be okay?

google/auth/transport/aiohttp_req.py
AniBadde: Resolved

google/auth/transport/aiohttp_req.py
AniBadde: Resolved

google/oauth2/_client.py
AniBadde: Resolved - had logic for async credential refreshing, but this doesn't seem to add any benefit so removed from implementation.

key_path
AniBadde: Resolved

noxfile.py
AniBadde: Resolved

tests_async/test_aiohttp_req.py
AniBadde: This test was taken from test_requests as both timeout guards are almost identical

tests_async/test_aiohttp_req.py
tritone: It doesn't seem hugely dangerous, however in python typically if there's a resource that needs to be closed after use we'd use a context manager to create it and then do necessary cleanup when the context is done. In unit tests this can often be achieved with some kind of pytest fixture. @crwilcox may have ideas about the best way to handle this.

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+            response = await self.session.request(
+                method, url, data=body, headers=headers, timeout=timeout, **kwargs
+            )
+            return _Response(response)
+
+        except aiohttp.ClientError as caught_exc:
+            new_exc = exceptions.TransportError(caught_exc)
+            six.raise_from(new_exc, caught_exc)
+
+        except asyncio.TimeoutError as caught_exc1:
```
tritone, line 180: Don't have to use a different var name for caught_exc here I believe.

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+            new_exc1 = exceptions.TransportError(caught_exc1)
```
tritone, line 181: Does a `TransportError` also make sense for timeouts?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+            six.raise_from(new_exc1, caught_exc1)
+
+
+class AuthorizedSession(aiohttp.ClientSession):
+    """
+
+    documentation
```
tritone, line 188: Plz add :)

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+        self._refresh_timeout = refresh_timeout
+        self._is_mtls = False
+        self._auth_request = auth_request
+        self._auth_request_session = None
+        self._loop = asyncio.get_event_loop()
+        self._refresh_lock = asyncio.Lock()
+
+        if auth_request is None:
+            self._auth_request_session = aiohttp.ClientSession()
+            auth_request = Request(self._auth_request_session)
```
tritone, line 213: This and the following can just be 1 line?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+        auth_request=None,
+    ):
+        super(AuthorizedSession, self).__init__()
+        self.credentials = credentials
+        self._refresh_status_codes = refresh_status_codes
+        self._max_refresh_attempts = max_refresh_attempts
+        self._refresh_timeout = refresh_timeout
+        self._is_mtls = False
+        self._auth_request = auth_request
+        self._auth_request_session = None
```
tritone, line 207: It looks like this is never set to anything other than None if there is no request passed in. Is that a problem?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+        self._loop = asyncio.get_event_loop()
+        self._refresh_lock = asyncio.Lock()
+
+        if auth_request is None:
+            self._auth_request_session = aiohttp.ClientSession()
+            auth_request = Request(self._auth_request_session)
+            self._auth_request = auth_request
```
tritone, line 214: I notice there is no use of a retry adapter here, which there is in the non-async version. Will there be retries on network errors here?  If not, is there a way to include this functionality?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def configure_mtls_channel(self, client_cert_callback=None):
+        """Configure the client certificate and key for SSL connection.
+        If client certificate and key are successfully obtained (from the given
+        client_cert_callback or from application default SSL credentials), a
+        :class:`_MutualTlsAdapter` instance will be mounted to "https://" prefix.
+        Args:
```
tritone, line 221: Missing blank lines between docstring sections here.

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+                mtls_adapter = _MutualTlsAdapter(cert, key)
+                self.mount("https://", mtls_adapter)
+        except (
+            exceptions.ClientCertError,
+            ImportError,
+            OpenSSL.crypto.Error,
+        ) as caught_exc:
+            new_exc = exceptions.MutualTLSChannelError(caught_exc)
+            six.raise_from(new_exc, caught_exc)
+        """
```
tritone, line 251: I see this whole method is now commented out. Is it no longer needed? If not, why?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+        self,
+        method,
+        url,
+        data=None,
+        headers=None,
+        max_allowed_time=None,
+        timeout=_DEFAULT_TIMEOUT,
+        **kwargs
+    ):
+
```
tritone, line 263: Add docstring.

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+            if timeout is None
+            else functools.partial(self._auth_request, timeout=timeout)
+        )
+
+        remaining_time = max_allowed_time
+
+        with TimeoutGuard(remaining_time) as guard:
+            self.credentials.before_request(auth_request, method, url, request_headers)
+
+        with TimeoutGuard(remaining_time) as guard:
```
tritone, line 284: Just because I'm curious, can you speak to how the TimeoutGuard here would work with await? And how did you decide on the longer timeout to use?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+                url,
+                data=data,
+                headers=headers,
+                max_allowed_time=remaining_time,
+                timeout=timeout,
+                _credential_refresh_attempt=_credential_refresh_attempt + 1,
+                **kwargs
+            )
+
+        await self._auth_request_session.close()
```
tritone, line 335: As in my comment about this in the test, I wonder if this needs a context manager or something similar to handle teardown.

tests/transport/compliance.py
```diff
@@ -64,7 +64,6 @@ def wait():
     def test_request_basic(self, server):
         request = self.make_request()
         response = request(url=server.url + "/basic", method="GET")
-
```
tritone, line 67: Remove the changes in this file.

tests_async/async_compliance.py
```diff
@@ -0,0 +1,125 @@
...
+        # pylint: enable=unused-variable
+
+        server = WSGIServer(application=app.wsgi_app)
+        server.start()
+        yield server
+        server.stop()
+
+    @pytest.mark.asyncio
+    async def test_request_basic(self, server):
+        request = self.make_request()
```
tritone, line 67: I'm a little confused, what is make_request actually calling here?

tests_async/test_aiohttp_req.py
tritone: Hopefully you can remove TimeoutGuard from the async-specific code and then delete these tests.

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+class _Response(transport.Response):
+    """
+    Requests transport response adapter.
+
+    Args:
+        response (requests.Response): The raw Requests response.
+    """
+
+    def __init__(self, response):
+        self.response = response
```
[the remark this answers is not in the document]
AniBadde: No reason in particular why they are different, just thought it would be easier to implement without the underscores

tests/transport/compliance.py
AniBadde: Resolved

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+            deadline_hit = min(self.remaining_timeout) <= 0
+
+        if deadline_hit:
+            raise self._timeout_error_type()
+
+
+class Request(transport.Request):
+    """Requests request adapter.
+
+    This class is used internally for making requests using various transports
```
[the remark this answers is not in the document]
AniBadde: Resolved both issues here

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,337 @@
...
+    async def __call__(
+        self,
+        url,
+        method="GET",
+        body=None,
+        headers=None,
+        timeout=_DEFAULT_TIMEOUT,
+        **kwargs
+    ):
+        """Make an HTTP request using aiohttp.
```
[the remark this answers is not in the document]
AniBadde: I think we would need to await session.request as it is an aiohttp.ClientSession.request call, which is a coroutine and would require async/await syntax - there may be differing conventions on this though.

google/auth/transport/aiohttp_req.py
AniBadde: Resolved

google/auth/transport/aiohttp_req.py
AniBadde: The description in the auth.exceptions file was 'error during the http request, which seemed to be the most reasonable choice among the exceptions options there, and it is a wrapper for the TimeoutError exception, so the transport error returned will be the timeout error.

google/auth/transport/aiohttp_req.py
AniBadde: Resolved

google/auth/transport/aiohttp_req.py
AniBadde: I thought the flow here was easier to follow instead of self._auth_request_session = aiohttp.ClientSession() inside the parameter of the Request object, though open to changing this if we would like to minimize line count

google/auth/transport/aiohttp_req.py
AniBadde: I think in all our use cases we do not input an auth_request parameter - if we do then we go away from the default aiohttp client session request required for the Authorized Session class in the async case. That's an interesting point though as in all the storage use cases this does begin as a None type.

google/auth/transport/aiohttp_req.py
AniBadde: The adapters were only compatible with sync requests and I wasn't able to find an async aiohttp equivalent, so ended up removing the retry logic - I think that if we are able to find an alternative to add this extra check to network errors that would definitely be good. 

My reasoning for removing it is that even with the adapter logic there isn't a guarantee of always creating a network connection (the probability of not doing so with perfect independence of each retry is just cubed ^3 and the choice of 3 seemed arbitrary), so without the guarantee I felt that functionally we weren't fundamentally doing something different with the retry adapter. Adding it would be good though.

google/auth/transport/aiohttp_req.py
AniBadde: Resolved

google/auth/transport/aiohttp_req.py
AniBadde: This was the same adapter issue, where we don't have adapters with async functionality. I therefore commented it out as it was not used directly in our Auth class, but if we are able to get that adapter we can then reintegrate the code (though this does not seem to be an essential function)

google/auth/transport/aiohttp_req.py
AniBadde: Sure - the time keeps running while other asynchronous functions may be performed (due to the use of time.time() in the timeout implementation). Therefore, I added in an additional buffer on the timeout value (though this is arbitrary and can be changed) to account for the possible time delays when other functions are running asynchronously. I added in 50% of the time as this intuitively seemed reasonable in terms of possible additional time taken for this to run (since the original 120s was already a conservative timeout value).

google/auth/transport/aiohttp_req.py
AniBadde: Yes, could we discuss this in more detail? Currently I just close the sessions, but it would be good to discuss with both both of you (@crwilcox) if this can be done in a better way as you mentioned in the earlier comment.

tests/transport/compliance.py
AniBadde: Resolved

tests_async/async_compliance.py
AniBadde: make_request is calling the function in the test_aiohttp_req file. That class takes in the compliance class as a parameter so the inheritance relationship allows us to call the make_request function in the other file. The request makes an aiohttp Request object from our aiohttp_req file.

tests_async/test_aiohttp_req.py
crwilcox: I think @tritone  is on the right track. We *can* do it the way authored, but usually as some kind of last resort. context managers are more pythonic and should be used when they can. Happy to chat about this if helpful.

google/auth/transport/aiohttp_req.py
AniBadde: Resolved

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,299 @@
+# Copyright 2020 Google LLC
```
crwilcox, line 1: is there a reason to abbreviate the name of this file?

google/auth/transport/aiohttp_req.py
AniBadde: Resolved

google/auth/_default_async.py
AniBadde: Resolved

google/auth/transport/aiohttp_req.py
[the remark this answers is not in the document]
crwilcox: I see the class itself is internal, so either way should be fine, but if we wouldn't want folks relying on this if we made the class public, it may make sense to make the field internal as well.

+1 that these likely should match eachother though

google/auth/__init__.py
AniBadde: Resolved

cert_path
AniBadde: Resolved

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,299 @@
...
+
+    This class is used internally for making requests using asyncio transports
+    in a consistent way. If you use :class:`AuthorizedSession` you do not need
+    to construct or use this class directly.
+
+    This class can be useful if you want to manually refresh a
+    :class:`~google.auth.credentials.Credentials` instance::
+
+        import google.auth.transport.aiohttp_req
+        import aiohttp
```
crwilcox, line 80: Does the user have to import aiohttp or is it a dep of aiohttp_req?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,299 @@
...
+        try:
+            _LOGGER.debug("Making request: %s %s", method, url)
+            response = await self.session.request(
+                method, url, data=body, headers=headers, timeout=timeout, **kwargs
+            )
+            return _Response(response)
+
+        except aiohttp.ClientError as caught_exc:
+            new_exc = exceptions.TransportError(caught_exc)
+            six.raise_from(new_exc, caught_exc)
```
crwilcox, line 137: six is used to handle py2/3 code, but aiohttp is only py3.6+. Is it needed?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,299 @@
...
+
+        except asyncio.TimeoutError as caught_exc:
+            new_exc = exceptions.TransportError(caught_exc)
+            six.raise_from(new_exc, caught_exc)
+
+
+class AuthorizedSession(aiohttp.ClientSession):
```
crwilcox, line 144: Would it be helpful to call this `AsyncAuthorizedSession`? are there places where the distinction would be more clear for users?

google/auth/transport/aiohttp_req.py
crwilcox: I would collapse these. Not that it is a big deal either way.

google/auth/transport/aiohttp_req.py
crwilcox: Any method that is marked as public should have a docstring. I would expect the docs on https://github.com/googleapis/google-auth-library-python/pull/547/files#diff-3d65f346e41412c7d3c700ff2dd913ebR111 to be very similar

google/auth/transport/aiohttp_req.py
crwilcox: I see that we are closing the session on the instance `self`. Is this an instance member on the sync version as well?

key_path
crwilcox: So it is needed?

setup.py
```diff
@@ -26,6 +26,8 @@
     'rsa>=3.1.4,<5; python_version >= "3"',
     "setuptools>=40.3.0",
     "six>=1.9.0",
+    "aiohttp>=3.6.2; python_version>='3.6'",
+    "aioresponses>=0.6.4; python_version>='3.6'",
```
crwilcox, line 30: is aioresponses needed for test and client code?

google/auth/transport/aiohttp_req.py
AniBadde: Abbreviated due to aiohttp_request appearing lengthy when called by the user (aiohttp_request.AuthorizedSession()) and it seemed relatively clear that req denoted request. Could change this though if the longer version or a better name is more suitable.

google/auth/transport/aiohttp_req.py
AniBadde: It is a dependency of aiohttp_req, so a user calling aiohttp_req Authorized Session would not need to import aiohttp - analogous case in the sync version is that the user does not need to import the requests library

google/auth/transport/aiohttp_req.py
AniBadde: I do think these are needed as aiohttp_req is only called with python 3.6+ and the exceptions that will need to be addressed are for the aiohttp errors that could occur when making a request call. Six is used to raise the exception and to my knowledge can be used for both python 2 and 3 so is compatible with 3.6+

tests_async/test_aiohttp_req.py
AniBadde: Sounds good, I can change the docstring to have this implementation instead. As I define the client session in the init function though, would we still use context managers in the implementation of the auth class?

setup.py
AniBadde: Resolved above comments. aioresponses is used as a mocking framework for async io responses, so is only required for test code. Changed to only include in the test dependencies,

AniBadde: Yes, the key_path file is generated when nox is run (even after deleting the file), so it does seem to be required.

AniBadde: This instance member is separate to the sync version, the reason i closed it here is to prevent the issue of hanging sessions.

google/auth/_oauth2client_async.py
```diff
@@ -0,0 +1,171 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: I don't see anything async-specific in this file. Could you speak to why it was necessary to duplicate this?

google/auth/credentials_async.py
tritone: Looks like this file is still present? Is it just because of before_request?

google/auth/credentials_async.py
```diff
@@ -0,0 +1,356 @@
...
+
+import six
+
+
+from google.auth import _helpers
+
+
+@six.add_metaclass(abc.ABCMeta)
+class Credentials(object):
+    """Base class for all credentials.
```
tritone, line 29: Can we de-duplicate this with the non-async credentials using a common base class? I would think that we should be able to just have a subclass of the sync version in this file and only override methods we need for async.

google/auth/jwt_async.py
```diff
@@ -0,0 +1,846 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: Again, this is a huge file-- what changes did you make vs. the sync version that necessitated making a copy?

google/auth/transport/aiohttp_req.py
tritone: Still commented out-- please remove.

google/auth/transport/aiohttp_req.py
tritone: Please add a docstring here.

google/oauth2/service_account_async.py
```diff
@@ -0,0 +1,606 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: Again, what's the necessity of duplicating this file?

system_tests/test_service_account.py
```diff
@@ -47,19 +45,3 @@ def test_refresh_success(http_request, credentials, token_info):
             "https://www.googleapis.com/auth/userinfo.profile",
         ]
     )
-
-def test_iam_signer(http_request, credentials):
```
tritone, line 51: Why is this test removed? Please add it back in.

system_tests_async/conftest.py
```diff
@@ -0,0 +1,125 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: I don't think you should need a whole separate async system test directory and conftest. Just add it to the existing system tests dir and put any new fixtures you are adding in the existing system_tests/conftest.py . You may have to rename some fixtures to get this to work.

system_tests_async/test_service_account.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.mark.asyncio
+async def test_refresh_no_scopes(http_request, credentials):
+    with pytest.raises(exceptions.RefreshError):
+        await credentials.refresh(http_request)
+
+@pytest.mark.asyncio
+async def test_refresh_success(http_request, credentials, token_info):
+    credentials = credentials.with_scopes(["email", "profile"])
+    #breakpoint()
```
tritone, line 36: remove

system_tests_async/test_service_account.py
```diff
@@ -0,0 +1,50 @@
...
+from google.oauth2 import service_account_async
+
+
+@pytest.fixture
+def credentials(service_account_file):
+    yield service_account_async.Credentials.from_service_account_file(service_account_file)
+
+
+@pytest.mark.asyncio
+async def test_refresh_no_scopes(http_request, credentials):
```
tritone, line 29: Add short docstring-- why is this test expected to throw an error?

tests/compute_engine/test_credentials.py
```diff
@@ -106,6 +106,7 @@ def test_refresh_error(self, get):
         assert excinfo.match(r"http error")
 
     @mock.patch("google.auth.compute_engine._metadata.get", autospec=True)
+    @pytest.mark.asyncio
```
tritone, line 109: I think you would need a separate test for async here?

tests_async/data/es256_privatekey.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN EC PRIVATE KEY-----
```
tritone, line 1: Make a commit to remove all these.

tests_async/transport/test_aiohttp_req.py
```diff
@@ -0,0 +1,165 @@
...
+class TestRequestResponse(async_compliance.RequestResponseTests):
+    @pytest.mark.asyncio
+    def make_request(self):
+        return aiohttp_req.Request()
+
+    def test_timeout(self):
+        http = mock.create_autospec(aiohttp.ClientSession, instance=True)
+        request = google.auth.transport.aiohttp_req.Request(http)
+        request(url="http://example.com", method="GET", timeout=5)
+
```
busunkim96, line 35: Hi @AniBadde, I think this is the test that's getting dinged for coverage? 

```
.nox/cover/lib/python3.7/site-packages/google/auth/transport/aiohttp_req.py           76      0      8      1    99%   133->135
```

google/auth/transport/aiohttp_req.py
```py
        try:
            if self.session is None:
                self.session = aiohttp.ClientSession()
            _LOGGER.debug("Making request: %s %s", method, url)
```

It *think* you need an additional test case where no session is passed to the constructor `google.auth.transport.aiohttp_req.Request()` instead of `google.auth.transport.aiohttp_req.Request(http)`. I

google/auth/_oauth2client_async.py
AniBadde: This new file was made to run the oauth2client code with the async service account functions.

tests_async/data/es256_privatekey.pem
AniBadde: I think these are sample data/credentials that were committed in previous versions too, should I still delete them?

tests/compute_engine/test_credentials.py
AniBadde: Resolved

system_tests_async/test_service_account.py
AniBadde: Resolved

google/auth/credentials_async.py
AniBadde: Resolved

tests_async/transport/test_aiohttp_req.py
AniBadde: My tests currently make a sample request by defining aiohttp_req.Request(). Should I make a test where I do include the http parameter for completeness?

AniBadde: ^ I just did that but it still doesn't seem to see those lines in coverage, do you think we could ignore those lines in the coverage file? When I removed those lines / traced through code for the tests where the parameter isn't defined it does go through these lines which it isn't showing coverage for

tests_async/data/es256_privatekey.pem
tritone: Yes, please remove them using git rm (no need to delete the data locally but they must be removed from this PR).

google/auth/_default_async.py
```diff
@@ -0,0 +1,333 @@
+# Copyright 2015 Google Inc.
```
tritone, line 1: Please use git mv when moving files-- otherwise it's hard to see what was changed in addition to the file being renamed.

google/auth/credentials_async.py
```diff
@@ -75,42 +74,11 @@ class AnonymousCredentials(credentials.AnonymousCredentials, Credentials):
...
-
-    def before_request(self, request, method, url, headers):
-        """Anonymous credentials do nothing to the request."""
-
-    def with_quota_project(self, quota_project_id):
-        raise ValueError("Anonymous credentials don't support quota project.")
-    '''
 
 @six.add_metaclass(abc.ABCMeta)
 class ReadOnlyScoped(credentials.ReadOnlyScoped):
```
tritone, line 84: I notice that this doesn't override anything from credentials.ReadOnlyScoped. Is there a reason for having an inheritance here rather than just aliasing credentials.ReadOnlyScoped? (Probably fine to have it like this also).

tests_async/data/es256_privatekey.pem
tritone: Sorry-- I missed that this stuff is already checked in under tests/data and isn't data that you generated.

Could you make these tests use tests/data instead of tests_async/data and avoid having these files copied over?

google/auth/_oauth2client_async.py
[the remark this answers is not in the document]
AniBadde: Resolved

system_tests_async/test_service_account.py
AniBadde: The test is expected to throw an error because appropriate credential scopes aren't passed in, added this docstring;

system_tests_async/test_service_account.py
AniBadde: Resolved

google/auth/jwt_async.py
AniBadde: The difference is in the jwt_async file we inherit the service account async credentials - this should be resolved with the refactoring. The reason we still define the async version of the classes when it only uses methods from the sync classes is because when we define an async credentials we keep a separate thread of async credentials.

google/oauth2/service_account_async.py
AniBadde: The reason we would require an async file is inheritance from the async credentials, but in the latest commit I resolved the code duplication by increasing inheritance in the async methods from the sync methods.

google/auth/credentials_async.py
AniBadde: The reason I did it like this was that if we were to define an async credentials (service_account_async credentials for example), we would inherit from this file. In that case, I wanted a separate thread of all the classes through-out so that we then have an inheritance in service_account_async of credentials_async.ReadOnlyScoped instead of credentials.ReadOnlyScoped. However, since all the methods in the async version are synchronous, we do not require any added functionality in this class which is is why it is defined but left as is.

AniBadde: I am currently drawing out the dependency graph and will add it to the design doc by tonight

google/auth/_default_async.py
AniBadde: Ok, noted, will do in the next round

tests_async/data/es256_privatekey.pem
AniBadde: Resolved in the latest commit. Changed the directory paths to point to tests/data and got rid of the duplication in tests_async.

system_tests/test_service_account.py
AniBadde: These tests are no longer contained in master, I don't recall removing them from the test file.

tests_async/conftest.py
```diff
@@ -21,7 +21,9 @@
 
 def pytest_configure():
     """Load public certificate and private key."""
-    pytest.data_dir = os.path.join(os.path.dirname(__file__), "data")
+    pytest.data_dir = os.path.join(
```
tritone, line 24: Nice, glad you were able to remove the duplicated test files.

google/auth/credentials_async.py
tritone: Thanks, I saw this in your design doc-- definitely helps!

cert_path
anibadde: This file is removed.

google/auth/__init__.py
anibadde: Implemented inheritance from the sync classes with the default_async and load credentials from file to async so we do reuse the load_credentials_from_file method in both.

google/auth/_default_async.py
anibadde: Implemented re-use of the constants for use in both classes, so these constants are inherited from the sync _default file.

google/auth/_default_async.py
anibadde: Kept the methods here which required async functionality (or used an async credentials file such as service_accounts_async). This avoids code duplication and keeps only the necessary components

google/auth/transport/aiohttp_req.py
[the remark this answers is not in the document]
anibadde: Will fix this today.

google/auth/transport/aiohttp_req.py
anibadde: Discussed this offline and kept naming as is.

google/auth/transport/aiohttp_req.py
anibadde: Discussed this offline and chose to leave it as it is.

google/auth/transport/aiohttp_req.py
anibadde: Docstring will be added in the latest version

key_path
anibadde: Will remove this from git

system_tests_async/conftest.py
anibadde: Will look into doing this refactoring in the 3rd PR

tritone:
Closing in favor of 3 smaller PRs.

PR #561, opened 2020-07-15
refactor: split 'with_quota_project' into separate base class

busunkim96:
https://github.com/googleapis/google-auth-library-python/pull/546#issuecomment-657876147
> Isn't this a breaking change? Should it have been released as 1.19.0?
> 
> We are already seeing test failures in downstream libs like `firebase-admin` where we extend this interface.
> 
> ```
> firebase_admin/db.py:983:0: W0223: Method 'with_quota_project' is abstract in class 'Credentials' but is not overridden (abstract-method)
> ```

https://github.com/googleapis/google-auth-library-python/pull/546#issuecomment-657883603
> 
> Actually this seems to be because pylint considers any method in an abstract class that throws NotImplementedError to be abstract: http://pylint-messages.wikidot.com/messages:w0223
> 
> Not quite sure if that should be considered a breaking change, or write it off as pylint being too pedantic.

CC @hiranya911

### DIFF
--- a/google/auth/app_engine.py
+++ b/google/auth/app_engine.py
@@ -77,7 +77,9 @@ def get_project_id():
     return app_identity.get_application_id()
 
 
-class Credentials(credentials.Scoped, credentials.Signing, credentials.Credentials):
+class Credentials(
+    credentials.Scoped, credentials.Signing, credentials.CredentialsWithQuotaProject
+):
     """App Engine standard environment credentials.
 
     These credentials use the App Engine App Identity API to obtain access
@@ -145,7 +147,7 @@ class Credentials(credentials.Scoped, credentials.Signing, credentials.Credentia
             quota_project_id=self.quota_project_id,
         )
 
-    @_helpers.copy_docstring(credentials.Credentials)
+    @_helpers.copy_docstring(credentials.CredentialsWithQuotaProject)
     def with_quota_project(self, quota_project_id):
         return self.__class__(
             scopes=self._scopes,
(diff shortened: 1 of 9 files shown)


tseaver:
FWIW, the `NotImplementedError` in the abstract base class could be considered "correct", as it adds a contract that all concrete subclasses *must* implement.  For better backward-compatibility, it might be better to add another abstract class with that method, rather than changing it to raise `ValueError`.  E.g.:

```python
class CredentialsWithQuotaProject(Credentials):
    @abc.abstractmethod
    def with_quota_project(self, quota_project_id):
        """Returns a copy of these credentials with a modified quota project
        Args:
            quota_project_id (str): The project to use for quota and
                billing purposes
        Returns:
            google.oauth2.credentials.Credentials: A new credentials instance.
        """
        raise NotImplementedError("This class does not support quota project.")
```

hiranya911:
Any plans to implement the change suggested by @tseaver? This is currently causing a failure in our CI system. I'm thinking of adding a lint ignore directive for now.

tseaver:
@google-cla I consent.

tseaver:
@google-cla I consent

@busunkim96 I've pushed dc3ae79 with my suggested refactoring.  I don't know why the CLA bot is unhappy with that commit.  If you don't like how that looks, feel free to rebase my commit away.

tseaver:
@googlebot I consent.

busunkim96:
@tseaver Thank you! 🙏

PR #572, opened 2020-07-28
feat: asyncio http request logic and asynchronous credentials logic 

anibadde:
1st Async Auth Class PR

- asyncio http request logic.
- asynchronous credentials functionality.


google/auth/credentials_async.py
```diff
@@ -0,0 +1,168 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Interfaces for credentials."""
+
+import abc
+
+import six
```
crwilcox, line 20: Should six be needed for this? this surface is only usable for 3.6+ correct?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,297 @@
...
+import aiohttp
+import six
+
+
+from google.auth import exceptions
+from google.auth import transport
+from google.auth.transport import requests
+
+
+_OAUTH_SCOPES = [
```
crwilcox, line 33: are these shared with the other transport? Can we extract them to a common import to avoid duplication?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,297 @@
...
+    """Requests request adapter.
+
+    This class is used internally for making requests using asyncio transports
+    in a consistent way. If you use :class:`AuthorizedSession` you do not need
+    to construct or use this class directly.
+
+    This class can be useful if you want to manually refresh a
+    :class:`~google.auth.credentials.Credentials` instance::
+
+        import google.auth.transport.aiohttp_req
```
crwilcox, line 79: I may have raised this before, but I am -1 on abbreviations. is there a motivation for naming this `aiohttp_req`? Instead of perhaps `aiohttp` or `aiohttp_request`?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,297 @@
...
+        import aiohttp
```
crwilcox, line 80: This import doesn't seem to be used in this sample. Is it required?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,297 @@
...
+    Args:
+        session (aiohttp.ClientSession): An instance :class: aiohttp.ClientSession used
+            to make HTTP requests. If not specified, a session will be created.
+
+    .. automethod:: __call__
+    """
+
+    def __init__(self, session=None):
+        """
+        self.session = None
```
crwilcox, line 95: Is this commented out code?

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,297 @@
...
+        body=None,
+        headers=None,
+        timeout=_DEFAULT_TIMEOUT,
+        **kwargs
+    ):
+        """
+        Make an HTTP request using aiohttp.
+
+        Args:
+            url (str): The URI to be requested.
```
crwilcox, line 114: this may be carry-over, does the old surface use a param `url` and refer to it as a `uri` as well? If so we can leave this but it seems a little strange

google/auth/transport/aiohttp_req.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def __init__(
+        self,
+        credentials,
+        refresh_status_codes=transport.DEFAULT_REFRESH_STATUS_CODES,
+        max_refresh_attempts=transport.DEFAULT_MAX_REFRESH_ATTEMPTS,
+        refresh_timeout=None,
+        auth_request=None,
+    ):
+        super(AuthorizedSession, self).__init__()
```
crwilcox, line 195: are all of these vars unique to the implementation? It seems `_loop` and ` _refresh_lock` are. Are any of these in the super() call and duplicate? If the args were passed?

noxfile.py
```diff
@@ -28,9 +28,35 @@
     "cryptography",
     "responses",
     "grpcio",
+    "pytest-asyncio",
+    "aioresponses",
 ]
+
+TEST_DEPENDENCIES2 = [
```
crwilcox, line 35: can we share this list, composing `test_dependencies` from this list and adding the other two?

google/auth/transport/aiohttp_req.py
anibadde: The motivation was the length, could change this to aiohttp_request though if that would be better. I'm worried about the name change to aiohttp just because of naming collision to the aiohttp library.

google/auth/credentials_async.py
anibadde: I used six to change the exception types (line 143, 147) so that we could raise our defined transport error.

google/auth/transport/aiohttp_req.py
anibadde: Made this change when switching the initialization of client session, deleted the commented out code now.

google/auth/transport/aiohttp_req.py
anibadde: Fixed, changed to aiohttp_requests in all 3 PRs

google/auth/transport/aiohttp_req.py
anibadde: Yes the aiohttp import is not required. Removed from the docstring.

google/auth/transport/aiohttp_req.py
anibadde: This was what the documentation was like in the old surface and therefore carried over on to here - I do agree the discrepancy is strange - it would seem that we want this to be URL. Changed in the current version.

google/auth/transport/aiohttp_req.py
anibadde: _loop and _refresh_lock are unique to this implementation. I think that as they are class variables with the self. syntax we wouldn't need to include them in the super() call, we could talk about this on our next call.

google/auth/transport/aiohttp_req.py
anibadde: Fixed in the latest commit.

noxfile.py
anibadde: Fixed in the latest commit to only have the 1 TEST_DEPENDENCIES list.

noxfile.py
```diff
@@ -107,7 +91,7 @@ def unit(session):
 
 @nox.session(python=["2.7", "3.5"])
 def unit_prev_versions(session):
-    session.install(*TEST_DEPENDENCIES2)
+    session.install(*TEST_DEPENDENCIES[:-2])
```
crwilcox, line 94: I am a little nervous about this, expecting to see folks add more deps :D.

can you make two lists one "dependencies" and one "additional_python_3". And for python3 create a set with the two?

tritone: +1

anibadde: Changed in the latest commit to have 2 lists and then a composition of both for the python3.6+ async tests.

google/auth/credentials_async.py
tseaver: This module uses `six` only for the `six.add_metaclass(abc.ABCMeta)` class decorator.  Given that it will only ever be imported under Python >= 3.6, it would be clearer to just use the direct Py3 syntax, e.g.:

```python
class Credentials(credentials.Credentials, metaclass=abc.ABCMeta):
    ...
```

google/auth/transport/aiohttp_requests.py
```diff
@@ -0,0 +1,313 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Transport adapter for Async HTTP (aiohttp)."""
+
+from __future__ import absolute_import
```
tseaver, line 17: Redundant under Python 3.x.

google/auth/transport/aiohttp_requests.py
```diff
@@ -0,0 +1,313 @@
...
+                self.session = aiohttp.ClientSession()  # pragma: NO COVER
+            requests._LOGGER.debug("Making request: %s %s", method, url)
+            response = await self.session.request(
+                method, url, data=body, headers=headers, timeout=timeout, **kwargs
+            )
+            return _Response(response)
+
+        except aiohttp.ClientError as caught_exc:
+            new_exc = exceptions.TransportError(caught_exc)
+            six.raise_from(new_exc, caught_exc)
```
tseaver, line 127: Given Python 3 only, this can be simplified to:

```python
        except aiohttp.ClientError as caught_exc:
            raise exceptions.TransportError(caught_exc) from caught_exc
```

google/auth/transport/aiohttp_requests.py
```diff
@@ -0,0 +1,313 @@
...
+
+        except asyncio.TimeoutError as caught_exc:
+            new_exc = exceptions.TransportError(caught_exc)
+            six.raise_from(new_exc, caught_exc)
```
tseaver, line 131: Likewise here:

```python
        except asyncio.TimeoutError as caught_exc:
            raise exceptions.TransportError(caught_exc) from caught_exc
```

tests_async/transport/async_compliance.py
```diff
@@ -0,0 +1,136 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import time
+
+import flask
+import pytest
+from pytest_localserver.http import WSGIServer
+from six.moves import http_client
```
tseaver, line 20: Given Python3-only, this can be just:

```python
import http.client
```

and then use `http.client.OK`, etc., below.

PR #573, opened 2020-07-28
feat: all asynchronous credential types required for all authentication types (service accounts, default... etc)

anibadde:
PR # 2 for AsyncIO Auth Library:

- Implemented all asynchronous credentials types
    - oauth2 
         - client
         - credentials (inherited from credentials_async)
         - service account
    - JSON Web Token
    - default credentials


google/auth/_default_async.py
```diff
@@ -0,0 +1,333 @@
...
+import six
+
+from google.auth import environment_vars
+from google.auth import exceptions
+import google.auth.transport._http_client
+
+_LOGGER = logging.getLogger(__name__)
+
+# Valid types accepted for file-based credentials.
+_AUTHORIZED_USER_TYPE = "authorized_user"
```
crwilcox, line 35: similar to previous PR, please go over all of the repeated constants, refactoring so we don't have duplicates that can get out of sync.

google/auth/_default_async.py
```diff
@@ -0,0 +1,307 @@
...
+
+    try:
+        credentials = app_engine.Credentials()
+        project_id = app_engine.get_project_id()
+        return credentials, project_id
+    except EnvironmentError:
+        return None, None
+
+
+def _get_gce_credentials(request=None):
```
crwilcox, line 175: this and the method above don't appear to be async at all. Is there a reason to have duplication from `_default.py`?

google/auth/_default_async.py
anibadde: Resolved, removed the repeated constants between the credentials files and test files in the latest commits.

google/auth/_default_async.py
anibadde: The reason for this is that when we initialize default credentials we initialize of the form default_async = google.auth._default_async, and then call default_async._get_gce_credentials and default_async._get_gae_credentials. So within this file when these two methods are called we can pull them from the sync file, but if these were to be called in other libraries then I wanted to be careful of having sync default and async default stepping on each other.

google/auth/_default_async.py
```diff
@@ -0,0 +1,280 @@
...
+import warnings
+
+import six
+
+from google.auth import _default
+from google.auth import environment_vars
+from google.auth import exceptions
+
+
+def _warn_about_problematic_credentials(credentials):
```
crwilcox, line 32: extract to a helper to share?

google/auth/credentials_async.py
```diff
@@ -62,7 +63,10 @@ async def before_request(self, request, method, url, headers):
         # the http request.)
 
         if not self.valid:
-            self.refresh(request)
+            if inspect.iscoroutinefunction(self.refresh):
```
crwilcox, line 66: what is the scenario where this wouldn't be a coroutine?

google/auth/jwt_async.py
```diff
@@ -0,0 +1,197 @@
...
+
+.. _rfc7519: https://tools.ietf.org/html/rfc7519
+
+"""
+
+import google.auth
+from google.auth import jwt
+
+
+def encode(signer, payload, header=None, key_id=None):
```
crwilcox, line 47: The top level methods of this don't appear to be specific to async. If that is so, you can get them here with `from google.auth.jwt import encode, _unverified_decode...`

google/oauth2/_client_async.py
```diff
@@ -0,0 +1,270 @@
...
+    # retry to fetch token for maximum of two times if any internal failure
+    # occurs.
+    while True:
+
+        response = await request(
+            method="POST", url=token_uri, headers=headers, body=body
+        )
+
+        """
+        except exceptions.TransportError as caught_exc:
```
crwilcox, line 108: commented out code

google/oauth2/_client_async.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+
+        response_body1 = await response.data.read()
+
+        response_body = (
+            response_body1.decode("utf-8")
+            if hasattr(response_body1, "decode")
+            else response_body1
+        )
+        # CHANGE TO READ TO END OF STREAM
```
crwilcox, line 120: TODO?

google/oauth2/_client_async.py
```diff
@@ -0,0 +1,270 @@
...
+        response_data = json.loads(response_body)
+
+        if response.status == http_client.OK:
+            break
+        else:
+            error_desc = response_data.get("error_description") or ""
+            error_code = response_data.get("error") or ""
+            if (
+                any(e == "internal_failure" for e in (error_code, error_desc))
+                and retry < 1
```
crwilcox, line 131: the condition for controlling the number of loops is pretty buried. I also think you may be able to use an existing retry helper for this?

tests_async/test__oauth2client.py
```diff
@@ -0,0 +1,170 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import datetime
+import sys
+
+import mock
+import oauth2client.client
```
crwilcox, line 19: This doesn't appear to be testing an async module, at least according to the naming. Is this a direct copy of what is in tests?

tests_async/test_credentials.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google LLC
+# Copyright 2020 Google LLC
```
crwilcox, line 1: similar, is this a test that is unique to async? If not we shouldn't duplicate it.

google/auth/credentials_async.py
anibadde: There are a couple of async credentials scenarios where self.refresh for those class instances return a None Type object (when it is not an async function). We run into this in an auth/jwt credentials object, and therefore add the check on whether refresh for that specific class instance is a coroutine or not.

google/auth/_default_async.py
anibadde: Changed to extract from auth/_default

google/auth/jwt_async.py
anibadde: Made this change.

google/oauth2/_client_async.py
anibadde: Made this change as the approach to reading in the http response was modified, comment deleted.

google/oauth2/_client_async.py
anibadde: I thought about changing this logic but am wary of diverging from the sync implementation of the error handling as we would want both to behave the same for consistency (especially if a user does use both async and sync clients). I would like to further understand the reasoning for using an existing retry helper so we could clarify this in our next call.

tests_async/test__oauth2client.py
anibadde: These do test the async _oauth2client internal to the auth library implementation. The other library that is imported is from https://github.com/googleapis/oauth2client, and the tests in this file test conversion between credentials from the oauth2client to the types of async credentials that we define in the auth library and that are added to this PR. Therefore, though these tests parallel those of the sync oauth2client, they use the async version and are as a result not a direct copy.

tests_async/test_credentials.py
anibadde: These tests are unique to async, I changed the names of the test files so that this is clear. The import of credentials_async is renamed to credentials to make writing the tests shorter but we need to test the new async constructions.

google/auth/jwt_async.py
anibadde: Turns out that the jwt_async test file defines our async version of the jwt token, and then calls encode/decode etc with the token object, so the direct import of that form wouldn't account for a token.decode/token.encode statement. Is there another way I could do the import to make that type of logic work? The reasoning is that there could be external locations where these encode/decode etc methods are called on the async tokens, so I'm leaning towards the current approach to account for those cases.

google/auth/_default_async.py
```diff
@@ -36,10 +35,7 @@ def _warn_about_problematic_credentials(credentials):
     are problematic because they may not have APIs enabled and have limited
     quota. If this is the case, warn about it.
     """
-    from google.auth import _cloud_sdk
-
-    if credentials.client_id == _cloud_sdk.CLOUD_SDK_CLIENT_ID:
-        warnings.warn(_default._CLOUD_SDK_CREDENTIALS_WARNING)
+    return _default._warn_about_problematic_credentials(credentials)
```
crwilcox, line 38: 99% sure you can just import _warn_about_problematic credentials and not wrap it here, further reducing duplication.

anibadde: Makes sense, made the modification to directly import and remove this method.

google/auth/_oauth2client_async.py
```diff
@@ -0,0 +1,171 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helpers for transitioning from oauth2client to google-auth.
+
+.. warning::
+    This module is private as it is intended to assist first-party downstream
+    clients with the transition from oauth2client to google-auth.
+"""
```
busunkim96, line 20: Hmm I'd kind of be alright with us *not* having this in the async version, given that oauth2client has been deprecated for a few years now. @crwilcox @anibadde thoughts?

anibadde: Interesting, wasn't aware that oauth2client had been depricated - As long as this isn't used in any of the other internal libraries like resumable media it may make sense to remove it if it is not going to be used.

crwilcox: sgtm :)

noxfile.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2020 Google LLC
```
tritone, line 1: This should be reverted to 2019 since it's not a new file.

anibadde: Changed.

PR #574, opened 2020-07-28
feat: added system tests for the asyncIO auth changes and async id_token credentials

anibadde:
PR # 3 for AsyncIO Auth Class changes

- Added in system tests paralleling existing regular system tests for the async implementation of the auth library and credentials
- Fixed Timeout Context Manager bug that enables connection to asynchronous resumable media implementation.
- Added in async id token credentials logic along with defaul, service account systems. 
- Required for credential depending on existence/validity of service account file


anibadde:
@googlebot I signed it!

anibadde:
@googlebot I signed it!

anibadde:
@googlebot I signed it!

google/oauth2/id_token_async.py
```diff
@@ -0,0 +1,267 @@
...
+library like `CacheControl`_ to create a cache-aware
+:class:`google.auth.transport.Request`::
+
+    import cachecontrol
+    import google.auth.transport.requests
+    import requests
+
+    session = requests.session()
+    cached_session = cachecontrol.CacheControl(session)
+    request = google.auth.transport.requests.Request(session=cached_session)
```
tritone, line 54: Would this be the same for async?

google/oauth2/id_token_async.py
```diff
@@ -0,0 +1,267 @@
...
+            data.
+    """
+    response = await request(certs_url, method="GET")
+
+    if response.status != http_client.OK:
+        raise exceptions.TransportError(
+            "Could not fetch certificates at {}".format(certs_url)
+        )
+
+    data = await response.data.read()
```
tritone, line 96: Is this a network op?

google/oauth2/id_token_async.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+async def verify_token(
+    id_token, request, audience=None, certs_url=sync_id_token._GOOGLE_OAUTH2_CERTS_URL
+):
+    """Verifies an ID token and returns the decoded token.
+
+    Args:
+        id_token (Union[str, bytes]): The encoded token.
+        request (google.auth.transport.Request): The object used to make
```
tritone, line 108: I'm guessing this only works for an aiohttp request?

system_tests_async/conftest.py
```diff
@@ -0,0 +1,125 @@
...
+import pytest
+import requests
+import urllib3
+
+import aiohttp
+import google.auth.transport.aiohttp_requests
+
+
+HERE = os.path.dirname(__file__)
+DATA_DIR = os.path.join(HERE, "data")
```
tritone, line 30: So it seems like this will require the data to be re-copied into system_tests_async/data as well? Could we avoid that by using the same data dir for both sets of system tests?

system_tests_async/conftest.py
```diff
@@ -0,0 +1,125 @@
...
+REQUESTS_SESSION = requests.Session()
+
+ASYNC_REQUESTS_SESSION = aiohttp.ClientSession()
+
+REQUESTS_SESSION.verify = False
+TOKEN_INFO_URL = "https://www.googleapis.com/oauth2/v3/tokeninfo"
+
+
+@pytest.fixture
+def service_account_file():
```
tritone, line 46: If these are copied from system_tests/conftest.py, could they be moved to the root conftest.py to de-duplicate?

system_tests_async/conftest.py
```diff
@@ -0,0 +1,125 @@
...
+    yield IMPERSONATED_SERVICE_ACCOUNT_FILE
+
+
+@pytest.fixture
+def authorized_user_file():
+    """The full path to a valid authorized user file."""
+    yield AUTHORIZED_USER_FILE
+
+
+#@pytest.fixture(params=["urllib3", "requests", "aiohttp"])
```
tritone, line 63: Remove commented-out code

system_tests_async/conftest.py
```diff
@@ -0,0 +1,125 @@
...
+@pytest.fixture(params=["aiohttp"])
+async def http_request(request):
+    """A transport.request object."""
+    if request.param == "urllib3":
```
tritone, line 67: Are the cases for urllib3 and requests necessary for the async tests?

system_tests_async/noxfile.py
```diff
@@ -0,0 +1,233 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: Update copyright to 2020 for new file.

system_tests_async/conftest.py
```diff
@@ -0,0 +1,125 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: Update copyright to 2020 for new file.

system_tests_async/noxfile.py
```diff
@@ -0,0 +1,233 @@
...
+
+import os
+import subprocess
+
+from nox.command import which
+import nox
+import py.path
+
+
+HERE = os.path.abspath(os.path.dirname(__file__))
```
tritone, line 33: I'm not familiar enough with nox to know, but, assuming this is mostly copied from system_tests/noxfile, is it possible to re-factor and deduplicate some of this?

system_tests_async/test_default.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: Update copyright

system_tests_async/test_service_account.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2016 Google LLC
```
tritone, line 1: Update copyright

tests_async/oauth2/test_id_token.py
```diff
@@ -0,0 +1,205 @@
+# Copyright 2014 Google Inc.
```
tritone, line 1: Update copyright

system_tests_async/test_service_account.py
```diff
@@ -0,0 +1,53 @@
...
+    info = await token_info(credentials.token)
+
+    assert info["email"] == credentials.service_account_email
+    info_scopes = _helpers.string_to_scopes(info["scope"])
+    assert set(info_scopes) == set(
+        [
+            "https://www.googleapis.com/auth/userinfo.email",
+            "https://www.googleapis.com/auth/userinfo.profile",
+        ]
+    )
```
tritone, line 53: So I see 2 system tests in this file, and 1 for id token. Does the sync version have additional system tests that you did not copy? If so, why?

google/oauth2/id_token_async.py
anibadde: I made this specific request call sync as an async version of this would result in a substantial re-working of the compute_engine credentials classes (as the current design is not compatible with asynchronous http requests). My reasoning for this was that this authentication trace isn't used in storage, but we could open up an issue for it as a future modification to make for the auth library.

google/oauth2/id_token_async.py
anibadde: Yes, response.data returns a stream coroutine, and therefore needs to be awaited in order to read the content of the response.

google/oauth2/id_token_async.py
anibadde: Yes, the input would have to be an aiohttp request if the async id token version is being used. Modified the documentation to account for this.

system_tests_async/conftest.py
anibadde: I thought about this, since the data files aren't in the repository and are there for the user to run locally, I thought that it may make sense to leave this logic as is (without a dependency on the sync system tests) in case someone wanted to import the data files just for the async use case, similar to what I did when I was working on this and running tests locally. I think that regardless of whether we change the directory name, we would have to follow the steps in https://github.com/googleapis/google-auth-library-python/blob/master/CONTRIBUTING.rst to be able to run these tests.

I'm open to changing the directory if that would be better, but I thought it would be good to avoid the dependency in this case?

system_tests_async/conftest.py
anibadde: Same inter-dependency reasoning as above, open to changing if we are okay with having the dependency.

system_tests_async/conftest.py
anibadde: Removed the code.

system_tests_async/conftest.py
anibadde: Not required, removed the first two conditionals.

system_tests_async/noxfile.py
anibadde: Done.

system_tests_async/conftest.py
anibadde: Done.

system_tests_async/test_default.py
anibadde: Done.

system_tests_async/test_service_account.py
anibadde: Done.

tests_async/oauth2/test_id_token.py
anibadde: Done

system_tests_async/test_service_account.py
anibadde: The sync version also has 1 system test for id_token and 2 for service accounts, so a parallel code trace (same corresponding tests) are used here as in the sync system tests.

google/oauth2/id_token_async.py
crwilcox: This seems incomplete. line 108 is still referencing `Request`

google/oauth2/id_token_async.py
```diff
@@ -0,0 +1,267 @@
...
+
+    Returns:
+        str: The ID token.
+
+    Raises:
+        ~google.auth.exceptions.DefaultCredentialsError:
+            If metadata server doesn't exist and no valid service account
+            credentials are found.
+    """
+    # 1. First try to fetch ID token from metada server if it exists. The code
```
crwilcox, line 217: metada -> metadata

system_tests_async/noxfile.py
crwilcox: We shouldn't have two noxfiles if at all possible

google/oauth2/id_token_async.py
crwilcox: Please open a tracking bug

system_tests_async/noxfile.py
anibadde: Resolved this to have 1 noxfile for both the sync and async system tests - essentially split up the python 2/3 dependencies and run both separately (similar to the sync approach)

google/oauth2/id_token_async.py
anibadde: Ok will do.

PR #584, opened 2020-08-07
Feat: gzip/decompressed response content handling 

anibadde:


google/oauth2/id_token_async.py
```diff
@@ -0,0 +1,267 @@
+# Copyright 2020 Google LLC
```
crwilcox, line 1: @anibadde what is this addition from?

google/auth/transport/aiohttp_requests.py
```diff
@@ -31,6 +32,51 @@
 _DEFAULT_TIMEOUT = 180  # in seconds
 
 
+class _CombinedResponse(transport.Response):
```
crwilcox, line 35: TODO: we are seeing if this can't be pushed into resumable

google/auth/transport/aiohttp_requests.py
```diff
@@ -230,8 +280,14 @@ async def request(
                 transmitted. The timout error will be raised after such
                 request completes.
         """
-
-        async with aiohttp.ClientSession() as self._auth_request_session:
+        if headers:
+            for key in headers.keys():
```
crwilcox, line 284: TODO: is this still an issue?

google/auth/transport/aiohttp_requests.py
crwilcox: made issue to track this.

google/auth/transport/aiohttp_requests.py
```diff
@@ -31,6 +32,51 @@
 _DEFAULT_TIMEOUT = 180  # in seconds
 
 
+class _CombinedResponse(transport.Response):
+    '''
+    In order to more closely resemble the `requests` interface, where a raw
+    and deflated content could be accessed at once, this class lazily reads the 
+    stream in `transport.Response` so both return forms can be used
+    '''
```
tritone, line 40: Use double quotes for docstring. Does this need to be fleshed out more?

google/auth/transport/aiohttp_requests.py
```diff
@@ -31,6 +32,51 @@
...
+    def __init__(self, response):
+        self._response = response
+        self._raw_content = None
+
+    def _is_compressed(self):
+        # The gzip and deflate transfer-encodings are automatically decoded for you.
```
tritone, line 46: This comment is a bit unclear, maybe rephrase.

google/auth/transport/aiohttp_requests.py
```diff
@@ -230,8 +280,16 @@ async def request(
                 transmitted. The timout error will be raised after such
                 request completes.
         """
+        # Headers come in as bytes which isn't expected behavior (resumable)
```
tritone, line 283: What do you mean by "isn't expected behavior"? Can you clarify to explain why the lines below are necessary?

google/auth/transport/aiohttp_requests.py
```diff
@@ -264,7 +322,7 @@ async def request(
                     data=data,
                     headers=request_headers,
                     timeout=timeout,
-                    **kwargs
+                    **kwargs,
```
tritone, line 325: Get rid of stray comma here and below. In general, it's good to review your diff to make sure you're only including changes that you really want to include.

google/auth/transport/aiohttp_requests.py
crwilcox: This isn't expected as most headers are of a str type. It is odd that some come through as bytes. This may be a bandaid, or may be entirely necessary :(

google/auth/transport/aiohttp_requests.py
crwilcox: I think this may be the autoformatter at work.  Might be worth running `nox -s blacken` across code quick actually, might be a differing version?

crwilcox: confirmed this is autoformatter (black)

google/oauth2/_client_async.py
```diff
@@ -104,6 +104,7 @@ async def _token_endpoint_request(request, token_uri, body):
             method="POST", url=token_uri, headers=headers, body=body
         )
 
+        # Using data.read() causing zlib decompression error
```
crwilcox, line 107:
```suggestion
        # Using data.read() resulted in zlib decompression errors. This may require future investigation.
```

google/auth/transport/aiohttp_requests.py
crwilcox: possibly move this into the comment up top as this is the point of this added class.

google/auth/transport/aiohttp_requests.py
anibadde: Resolved, changed to double quotes with blacken.

google/auth/transport/aiohttp_requests.py
anibadde: Fleshed out this comment in the latest version to explain reasoning.

google/auth/transport/aiohttp_requests.py
anibadde: Added more explanation to why we are doing this. Inconsistencies in the header value types for some cases in the resumable library, which is why we have this check.

PR #591, opened 2020-08-24
feat: add GOOGLE_API_USE_CLIENT_CERTIFICATE support

arithmetic1728:
Implement the GOOGLE_API_USE_CLIENT_CERTIFICATE variable from https://google.aip.dev/auth/4114. Unless users explicitly set this environment to True, we don't use the default client certificate from device (from the context_aware_metadata.json) and the user provided client certificate (via callback).

### DIFF
--- a/docs/reference/google.auth.transport.mtls.rst
+++ b/docs/reference/google.auth.transport.mtls.rst
@@ -1,0 +1,7 @@
+google.auth.transport.mtls module
+=================================
+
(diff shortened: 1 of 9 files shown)


arithmetic1728:
> Can the env var be moved into https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/environment_vars.py?

Done.

Issue #595, opened 2020-08-31
Setuptools as dependency is problematic w/ pip-tools


akx:
https://github.com/googleapis/google-auth-library-python/commit/908da752d01fef728bd5cb3eb5b13f2b5c335e51 (#322) added `setuptools` as a dependency in this package. However, the [pip-tools](https://github.com/jazzband/pip-tools) package that's commonly used for pinning dependencies considers `setuptools` an unsafe dependency to have in a project at all (as discussed in #492), and as such doesn't save it in the pinned requirements file at all.

Since `google-auth` depends on Setuptools but a version couldn't have been pinned in the requirements, we're seeing

```
Collecting setuptools>=40.3.0 (from google-auth==1.19.1->our-proprietary-package==0.31.1)
  Downloading https://files.pythonhosted.org/packages/b0/8b/379494d7dbd3854aa7b85b216cb0af54edcb7fce7d086ba3e35522a713cf/setuptools-50.0.0-py3-none-any.whl (783kB)
```

which wreaks havoc on Ubuntu 16.04 + Python 3.5 machines due to https://github.com/pypa/setuptools/issues/2352 / https://github.com/pypa/setuptools/issues/2350 / https://github.com/pypa/setuptools/issues/2356 ...

The workaround is to add `--allow-unsafe` or manually pin `setuptools`, but is the requirement _actually_ necessary in this package? No other package in the 48-line `requirements.txt` for this particular project would have required a version of `setuptools`.

#### Environment details

  - OS: Ubuntu 16.04
  - Python version: 3.5
  - pip version: irrelevant
  - `google-auth` version: 1.19.1

#### Steps to reproduce

  1. Install `google-auth` on an Ubuntu 16.04 machine
  2. It installs `setuptools==50.0.0`
  3. https://github.com/pypa/setuptools/issues/2352 and friends

busunkim96:
Hi @akx,

It looks like the pin was added to avoid an issue related to namespaces not being handled correctly in older versions of setuptools. 

https://github.com/googleapis/google-auth-library-python/pull/322
> Gotcha.
> 
> Grep'ing https://setuptools.readthedocs.io/en/latest/history.html for "namespace" shows `40.3.0` solved a problem with `pkg_resource`-style namespaces ([pypa/setuptools#1321](https://github.com/pypa/setuptools/issues/1321)).
> 
> That said, `40.3.0` was released Sept 16, 2018 which is fairly recent...
> 
[...]

PR #596, opened 2020-08-31
fix: remove checks for ancient versions of Cryptography

akx:
Refs https://github.com/googleapis/google-auth-library-python/issues/595#issuecomment-683903062 

I see no point in checking whether someone is running a version of https://github.com/pyca/cryptography/ from 2014 that doesn't even compile against modern versions of OpenSSL anymore.


busunkim96 (approved):
Thank you @akx!

PR #597, opened 2020-09-03
fix: trigger an auto release

arithmetic1728:
trigger an auto release for #561


busunkim96:
Thanks @arithmetic1728!

PR #598, opened 2020-09-03
feat: implements the OAuth token exchange spec based on rfc8693

bojeil-google:
Implements an internal utility for exchanging OAuth tokens using the rfc/8693 spec.


google/oauth2/sts.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def exchange_token(
+        self,
+        request,
+        grant_type,
+        subject_token,
+        subject_token_type,
+        resource=None,
+        audience=None,
+        scope=None,
```
busunkim96, line 69: Can this be renamed to `scopes` for consistency with other methods in this library? https://github.com/googleapis/google-auth-library-python/blob/41599ae932db31cf692d2c7bec15d8bb1778ab50/google/oauth2/credentials.py#L67

google/oauth2/sts.py
```diff
@@ -0,0 +1,155 @@
...
+            subject_token_type (str): The OAuth 2.0 token exchange subject token type.
+            resource (Optional[str]): The optional OAuth 2.0 token exchange resource field.
+            audience (Optional[str]): The optional OAuth 2.0 token exchange audience field.
+            scope (Optional[Sequence[str]]): The optional list of scopes to use.
+            requested_token_type (Optional[str]): The optional OAuth 2.0 token exchange requested
+                token type.
+            actor_token (Optional[str]): The optional OAuth 2.0 token exchange actor token.
+            actor_token_type (Optional[str]): The optional OAuth 2.0 token exchange actor token type.
+            additional_options (Optional[Mapping[str, str]]): The optional additional
+                non-standard GCP specific options.
```
busunkim96, line 93: Same comment as above
```suggestion
                non-standard Google specific options.
```

google/oauth2/sts.py
```diff
@@ -0,0 +1,155 @@
...
+"""OAuth 2.0 Token Exchange Spec.
+
+This module defines a token exchange utility based on the `OAuth 2.0 Token
+Exchange`_ spec. This will be mainly used to exchange external credentials
+for GCP access tokens in workload identity pools to access Google APIs.
+
+The implementation will support various types of client authentication as
+allowed in the spec.
+
+A deviation on the spec will be for additional GCP specific options that
```
busunkim96, line 24: Unless these options are only available with Cloud APIs, I think it's better to be more general (Google instead of GCP).
```suggestion
A deviation on the spec will be for additional Google specific options that
```

busunkim96 (approved):

google/oauth2/sts.py
bojeil-google: The spec defines this as `scope` (the request parameter) but since that is the space separated string, and this is an array which will be joined into a string, I have renamed it to `scopes`. Note that the base credential class will align with the rest of the library in using `scopes`.

google/oauth2/sts.py
bojeil-google: Done

google/oauth2/sts.py
bojeil-google: Done

PR #600, opened 2020-09-08
fix: "Revert "fix: migrate signBlob to iamcredentials.googleapis.com""

busunkim96:
Reverts the revert in googleapis/google-auth-library-python#563.  (original PR is #553) This change is safe to take now that changes have been made internally.

(Copied from #553)

> Migrate signBlob from iam.googleapis.com to iamcredentials.googleapis.com.
> 
> This API is deprecated and will be shutdown in one year.
> 
> This is used google.auth.iam.Signer.
> Added a system_test to sanity check the implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ pylintrc.test
 pytype_output/
 
 .python-version
+.DS_Store
+cert_path
+key_path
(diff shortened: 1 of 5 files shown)


arithmetic1728 (approved):

busunkim96:
@bojeil-google Could you also take another look (since you were the original author)?

bojeil-google (approved):
Thanks for taking care of this!

PR #603, opened 2020-09-10
feat: defines `google.auth.external_account.Credentials` abstract class for external_account credentials

bojeil-google:
Base external account credentials are used to instantiate Credentials for exchanging external account credentials for Google access tokens and authorizing requests to Google APIs.

The base class implements common logic for exchanging various type of external credentials via GCP STS token exchange endpoint for Google access tokens. The logic of determining and retrieving the external credential based on the environment and `credential_source` in the credentials JSON file will be left for the subclasses.


busunkim96 (approved):
LGTM.

Heads up, at some point the docs will need to be updated. (It doesn't need to happen in this PR).

These are the files that should probably be modified:
* https://github.com/googleapis/google-auth-library-python/tree/master/docs/reference (add new RST files)
* https://github.com/googleapis/google-auth-library-python/blob/master/docs/user-guide.rst
* https://github.com/googleapis/google-auth-library-python/blob/master/docs/index.rst

You can build the docs locally by doing `nox -s docs` and opening `docs/html`. It should look like https://googleapis.dev/python/google-auth/latest/index.html

bojeil-google:
Thanks @busunkim96 for the review and docs generation instructions. I absolutely agree on the references/docs generation. That is a launch blocker (we can't launch without that). We will dedicate a lot of effort to that along with snippets and other developer related materials. We have been delaying it as the Python implementation is still in early stages.

PR #617, opened 2020-09-29
feat: defines `google.auth.identity_pool.Credentials` used for K8s and Azure workloads

bojeil-google:
This will subclass the abstract class `google.auth.external_account.Credentials` and will retrieve subject tokens from URLs (eg Azure Instance Metadata Service) or a file location (K8s KSA tokens).

The current change only covers file-sourced credentials. URL-sourced credentials will be covered in a follow up.

The spec has been extended to support generic formats for `google.auth.identity_pool.Credentials` (json and text).
```
  credential_source: {
    // ...
    format?: {
      type: 'text' | 'json';
      subject_token_field_name?: string;
    };
  };
```
When `text` format is specified, the raw file content is used as the subject token.
When `json` format is specified, the file content is parsed as json and the field with key `subject_token_field_name` is used to retrieve the subject token.

The current implementation deviates from the original design doc in that it separates IdentityPoolCredentials into its separate module `google.auth.identity_pool.Credentials` instead of `google.auth.external_account.IdentityPoolCredentials`. This is a better match for the current pattern in the repo.


google/auth/identity_pool.py
```diff
@@ -0,0 +1,206 @@
...
+            ),
+            client_id=info.get("client_id"),
+            client_secret=info.get("client_secret"),
+            credential_source=info.get("credential_source"),
+            quota_project_id=info.get("quota_project_id"),
+            **kwargs
+        )
+
+    @classmethod
+    def from_file(cls, filename, **kwargs):
```
busunkim96, line 193: Does it make sense to allow this type of credential to be loaded via `google.auth.load_credentials_from_file`? Currently authorized user (gcloud) and service account are supported.

I can see an argument for not adding it as well - these are external credentials and it looks like it might be hard to detect the file type.

https://github.com/googleapis/google-auth-library-python/blob/ecd88d4f0efc5c619ebd3e3fa7e2472f11c63452/google/auth/_default.py#L72

bojeil-google: That is the eventual plan and the recommended flow for developers.

We definitely want to support this new type of credential as part of ADC but we also want to provide the ability to explicitly initialize this similar to the ability to initialize service account credentials explicitly from a file or dictionary:
https://github.com/googleapis/google-auth-library-python/blob/ee5617ccfb7caa6ba652c9cc07f716af31eb6ac0/google/oauth2/service_account.py#L216

and

https://github.com/googleapis/google-auth-library-python/blob/ee5617ccfb7caa6ba652c9cc07f716af31eb6ac0/google/oauth2/service_account.py#L195

busunkim96 (approved):

busunkim96:
CLA was raising error on commits already in master with a signed CLA.

PR #619, opened 2020-10-05
fix: move aiohttp to extra as it is currently internal surface

### LINKED ISSUE #618: Use extra for asyncio dependencies
Hello!  The latest release for this library pulls in aiohttp and its dependencies unconditionally, which adds non-trivial burden to projects that don’t need it.  Would you consider using a packaging extra so that people can opt-in?

crwilcox:
Fix #618. Removes aiohttp from required dependencies to lessen dependency tree for google-auth.

This will need to be looked at again as more folks use aiohttp and once the surfaces goes to public visibility.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,9 +27,11 @@ DEPENDENCIES = (
     'rsa>=3.1.4,<5; python_version >= "3.5"',
     "setuptools>=40.3.0",
     "six>=1.9.0",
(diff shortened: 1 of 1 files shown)


busunkim96 (approved):

crwilcox:
Looks like i may need to adjust install for test scenarios to include extras.

PR #624, opened 2020-10-11
Update example in oauth2.id_token docs

matthewhughes934:
Since c05b8b5 oauth2.id_token.verify_oauth2_token handles the issuer
check itself, so remove this redundant check from the docs.

### DIFF
--- a/google/oauth2/id_token.py
+++ b/google/oauth2/id_token.py
@@ -33,9 +33,6 @@ Example::
     id_info = id_token.verify_oauth2_token(
         token, request, 'my-client-id.example.com')
 
-    if id_info['iss'] != 'https://accounts.google.com':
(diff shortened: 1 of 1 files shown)


google/oauth2/id_token.py
```diff
@@ -33,9 +33,6 @@
     id_info = id_token.verify_oauth2_token(
         token, request, 'my-client-id.example.com')
 
-    if id_info['iss'] != 'https://accounts.google.com':
```
matthewhughes934, line 36: This change does make it less clear that an exception is raised (from `verify_oauth2_token`)

tseaver: LGTM.  We don't normally track exceptions raised by other functions, anyway.

tseaver (approved):

PR #625, opened 2020-10-14
feat: defines `google.auth.aws.Credentials` used for AWS workloads

bojeil-google:
This will subclass the abstract class `google.auth.external_account.Credentials` and will compute subject tokens as follows:

- Retrieve AWS region from either `AWS_REGION` envvar or AWS metadata server `availability-zone`.
- Check AWS credentials in environment variables:
  - `AWS_ACCESS_KEY_ID`
  - `AWS_SECRET_ACCESS_KEY`
  - `AWS_SESSION_TOKEN`.

  If not found, get from AWS metadata server `security-credentials` endpoint.
- Get AWS credentials from AWS metadata server `security-credentials` endpoint.
  In order to retrieve this, the AWS role needs to be determined by calling
  `security-credentials` endpoint without any argument. Then the
  credentials can be retrieved via: `security-credentials/role_name`
- Generate the signed request to AWS STS `GetCallerIdentity` action.
- Inject `x-goog-cloud-target-resource` into reformatted header and serialize the
  signed request. This will be the subject-token to pass to GCP STS.


google/auth/aws.py
```diff
@@ -302,3 +322,375 @@ def _generate_authentication_header_map(
...
+                "aws version '{}' is not supported in the current build.".format(
+                    env_version
+                )
+            )
+
+    def retrieve_subject_token(self, request):
+        """Retrieves the subject token using the credential_source object.
+        In this case, this uses a serialized AWS signed request to the AWS STS
+        GetCallerIdentity endpoint.
+        The logic is summarized as:
```
busunkim96, line 410: Is there public documentation somewhere on how this should work? If so, can a link be added?

busunkim96 (approved):

google/auth/aws.py
bojeil-google: Absolutely. I added the link.

busunkim96: Thank you! 😄

PR #627, opened 2020-10-20
test: add asynctest to async dependencies

busunkim96:
`aioresponses==0.7.0` seems to have a hard dependency on asynctest for python < 3.8. See https://github.com/pnuckowski/aioresponses/issues/172

This adds `asynctest` to the test dependencies so the 3.6 and 3.7 unit test sessions pass.


tseaver:
Superseded by #630.

PR #631, opened 2020-10-23
feat: implement `get_project_id()` for google.auth.external_account.Credentials

bojeil-google:
This is introduced to support the current pattern of using the Auth library:

`credentials, project_id = google.auth.default()`

This will be added to the project ID determination logic:

https://github.com/googleapis/google-auth-library-python/blob/3b3172ef94c110c81a49bc160123e8ff55141e65/google/auth/_default.py#L338

This will first determine the project number from the STS audience:
`//iam.googleapis.com/projects/$PROJECT_NUMBER/locations/...`

It will then call cloud resource manager to determine the project information:
https://cloudresourcemanager.googleapis.com/v1/projects/$PROJECT_NUMBER

This may fail for the following reasons:

- The resource may not have permission (resourcemanager.projects.get) to call this API
- Required scopes may not be selected:
  https://cloud.google.com/resource-manager/reference/rest/v1/projects/get#authorization-scopes

To reduce cost of this API, we will fail quickly and not retry and we will use scopes provided as is.


tseaver:
@bojeil-google Please rebase your branch against the current `master`, to remove the clutter which is causing the CLA bot to stumble.  E.g.:

```bash
$ git checkout master
$ git pull upstream master   # assuming this repo is the `upstream` remote
$ git checkout external_account_project_id
$ git rebase -i master  # edit out any commits except for the last
```

busunkim96 (approved):

Issue #579, opened 2020-08-05
Default credentials on App Engine standard (Py3) do not support custom scopes


davidwtbuxton:
Using the `google.auth.default()` helper to access credentials for the default service account on App Engine standard works great, but does not allow one to use auth scopes beyond the default scopes for the Python 3.7 and 3.8 runtimes. This is different to how the old 2.7 runtime works.

This app works on 2.7, fails with an exception on 3.8:

```
# main.py, requires flask, google-api-python-client
import flask
from googleapiclient import discovery

app = flask.Flask(__name__)
@app.route('/')
def home():
    service = discovery.build('drive', 'v3')

    request = service.files().list()
    response = request.execute()

    return response
```

On 3.8 the error is

```
googleapiclient.errors.HttpError: <HttpError 403 when requesting https://www.googleapis.com/drive/v3/files?alt=json returned "Insufficient Permission: Request had insufficient authentication scopes.">" 
```

The problem is that `google.auth.default()` returns an instance of the Compute Engine credentials for Python 3, and that class does not allow changing scopes.

The credentials class [uses the metadata service to get an access token][1]. On App Engine standard the metadata service also allows one to request an access token with additional scopes (I haven't checked if this is also possible on Compute Engine or Flex). I have an implementation  of `ServiceAccountCredentials` that supports requesting a token with additional scopes, which works on App Engine standard Python 3.

https://gist.github.com/davidwtbuxton/525924b7f06f56b8530947d55bad1c21

With that code, the service discovery can request the required scopes:

```
    credentials = ServiceAccountCredentials()
    service = discovery.build('drive', 'v3', credentials=credentials)
```

It would be cool if we could get this supported using `google.auth.default()` for Python 3 on App Engine standard. In particular it simplifies a lot of code that may mess around loading credentials from a JSON file or similar.

Would a PR for this feature be accepted?

Thanks,

David

[1]: https://github.com/googleapis/google-auth-library-python/blob/20f82e22b7e8c6c7fdd29e08eaf7b4cf2abdcf37/google/auth/compute_engine/credentials.py#L100

busunkim96:
Hi @davidwtbuxton,

I believe Compute Engine is the only one (that uses the metadata server) and prevents additional scopes.  IIRC, NodeJS and possibly other language's auth libraries already allow scope changes for this type of credential. I think it would be appropriate to make that change here as well. If you are able to make a PR we would be happy to review. 😄

[...]

Issue #632, opened 2020-10-24
TODO: undo pin of 'aiohttp' once 'aioresponses' releases a fix


davidwtbuxton:
Environment details

  - OS: $ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.14.6
BuildVersion:   18G6020

  - Python version: 3.6, 3.7, 3.8
  - pip version: pip 20.2.4
  - `google-auth` version: 5906c8583ca351b5385a079a30521a9a8a0c7c59

#### Steps to reproduce

  1. nox -s unit

There are 9 tests that fail, all with the same error:

`TypeError: __init__() missing 1 required positional argument: 'limit'`

```
====================================================== short test summary info =======================================================
FAILED tests_async/transport/test_aiohttp_requests.py::TestCombinedResponse::test_content_compressed - TypeError: __init__() missin...
FAILED tests_async/transport/test_aiohttp_requests.py::TestResponse::test_headers_prop - TypeError: __init__() missing 1 required p...
FAILED tests_async/transport/test_aiohttp_requests.py::TestResponse::test_status_prop - TypeError: __init__() missing 1 required po...
FAILED tests_async/transport/test_aiohttp_requests.py::TestAuthorizedSession::test_request - TypeError: __init__() missing 1 requir...
FAILED tests_async/transport/test_aiohttp_requests.py::TestAuthorizedSession::test_ctx - TypeError: __init__() missing 1 required p...
FAILED tests_async/transport/test_aiohttp_requests.py::TestAuthorizedSession::test_http_headers - TypeError: __init__() missing 1 r...
FAILED tests_async/transport/test_aiohttp_requests.py::TestAuthorizedSession::test_regexp_example - TypeError: __init__() missing 1...
FAILED tests_async/transport/test_aiohttp_requests.py::TestAuthorizedSession::test_request_no_refresh - TypeError: __init__() missi...
FAILED tests_async/transport/test_aiohttp_requests.py::TestAuthorizedSession::test_request_refresh - TypeError: __init__() missing ...
============================================ 9 failed, 609 passed, 12 warnings in 33.41s =============================================
```

Here is the traceback for one of the failing tests:

```
____________________________________________ TestCombinedResponse.test_content_compressed ____________________________________________

self = <tests_async.transport.test_aiohttp_requests.TestCombinedResponse object at 0x108803160>
urllib3_mock = <function decompress at 0x10880a820>

    @mock.patch(
[...]

PR #633, opened 2020-10-25
feat: add custom scopes for access tokens from the metadata service

davidwtbuxton:
This works for App Engine, Cloud Run and Flex. On Compute Engine you
can request custom scopes, but they are ignored.

Part of #579 

I will create a separate PR to update the user guide, which has been out-of-date for a while.

https://google-auth.readthedocs.io/en/latest/user-guide.html#compute-engine-container-engine-and-the-app-engine-flexible-environment

I've tested this on App Engine standard and Cloud Run, and using `google.auth.default(scopes=..)` works now. I tested on Compute Engine and requesting scopes has no effect, but does not raise an exception (which is expected).

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -274,10 +274,11 @@ def default(scopes=None, request=None, quota_project_id=None):
             gcloud config set project
 
     3. If the application is running in the `App Engine standard environment`_
-       then the credentials and project ID from the `App Identity Service`_
-       are used.
-    4. If the application is running in `Compute Engine`_ or the
-       `App Engine flexible environment`_ then the credentials and project ID
+       (first generation) then the credentials and project ID from the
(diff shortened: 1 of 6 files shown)


tseaver (dismissed):

davidwtbuxton:
Thanks for looking at this PR @tseaver 

Tests are failing because of #632 , which impacts the head of master too. They would start passing again if https://github.com/pnuckowski/aioresponses/pull/174 is accepted and a new release of aioresponses is published on PyPI.

tseaver:
PR #634 fixes the unit tests on `master.  I've merged to see if this PR will then pass.

silvolu (approved):

silvolu:
Auth bits looks good. Nice to see you again @tseaver :)

busunkim96 (approved):

PR #660, opened 2021-01-08
fix: Add basic AWS url validation

bojeil-google:
Recommended by the security reviewer, this change adds basic scheme/hostname validation to the AWS request URL.
This is auto-generated as part of the configuration file for external accounts. In case, the URL is manually modified to an invalid field, this should help catch that.


busunkim96 (approved):

busunkim96:
CLAs were granted when commits above were merged to master.

PR #665, opened 2021-01-12
feat: use self-signed jwt for service account

busunkim96:
https://google.aip.dev/auth/4111
Googlers see [this doc](https://docs.google.com/document/d/1SNCVTmW6Rtr__u-_V7nsT9PhSzjj1z0P9fAD3YUgRoc/edit#heading=h.kvke81ak77az) for detailed design.

Create a self-signed JWT from the service account credentials when the following conditions are met:

1. A service account is being used 
2. A scope has not been set by the user. (Enforced by adding a new `default_scopes` that is separate from `scopes`).
3. A target_audience has not been set by the user (not relevant at the moment since no client option for audience exists).
4. An `api_endpoint` has not been set by the user (taken care of in the GAPIC client library).

Other Changes:
- Add `default_scopes` to all `ScopedCredentials`. For non-service account credentials, the `default_scopes` are used when `scopes` is not set. Otherwise the `default_scopes` are ignored.

**This change is only being made to synchronous credentials and the gRPC transport.** The async credentials and other transports will be handled in follow up PRs.


google/auth/credentials.py
```diff
@@ -277,7 +286,7 @@ class Scoped(ReadOnlyScoped):
     """
 
     @abc.abstractmethod
-    def with_scopes(self, scopes):
+    def with_scopes(self, scopes, default_scopes=None):
```
busunkim96, line 289: Is adding an additional optional parameter to an abstract method a breaking change?

arithmetic1728 (approved):
LGTM

google/auth/credentials.py
arithmetic1728: It seems fine. 
https://stackoverflow.com/questions/42778784/abstract-classes-with-varying-amounts-of-parameters

google/auth/transport/grpc.py
```diff
@@ -51,15 +52,19 @@ class AuthMetadataPlugin(grpc.AuthMetadataPlugin):
             add to requests.
         request (google.auth.transport.Request): A HTTP transport request
             object used to refresh credentials as needed.
+        audience (Optional[str]): An audience like "pubsub.googleapis.com".
```
bshaffer, line 55: Here audience is not prefixed with `https://`. The [AIP](https://google.aip.dev/auth/4111) specifies audience as a full URI with `https://`. Could we make this consistent, or give it a different name, e.g. `service_address`?

The proto refers to it as [`default_host`](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L41)

bshaffer (commented):
One nit, otherwise this is great

google/auth/transport/grpc.py
busunkim96: 👍 Renamed to `default_host`.

bshaffer (approved):

busunkim96:
Looks like I accidentally removed the system tests from the presubmit last year 🤦‍♀️ . I was hoping it would be a small fix but it looks like some issues have piled up.  I'll pull those changes into a separate PR and revisit this one later.

PR #679, opened 2021-02-03
feat: support self-signed jwt in requests and urllib3 transports

busunkim96:
Follow up to #665.

Allow self-signed JWT flow to be used in the `requests` and `urllib3` transports if a `default_host` is provided. I don't think any of our clients wrap the `urllib3` transport, but a good number of handwritten libraries (Storage for example) uses `AuthorizedSession` from `requests`.

### DIFF
--- a/google/auth/transport/requests.py
+++ b/google/auth/transport/requests.py
@@ -44,6 +44,7 @@ import six  # pylint: disable=ungrouped-imports
 from google.auth import environment_vars
 from google.auth import exceptions
 from google.auth import transport
+from google.oauth2 import service_account
 import google.auth.transport._mtls_helper
 
 _LOGGER = logging.getLogger(__name__)
@@ -313,6 +314,9 @@ class AuthorizedSession(requests.Session):
             refreshing credentials. If not passed,
             an instance of :class:`~google.auth.transport.requests.Request`
             is created.
+        default_host (Optional[str]): A host like "pubsub.googleapis.com".
+            This is used when a self-signed JWT is created from service
+            account credentials.
     """
 
     def __init__(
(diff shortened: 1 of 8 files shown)


system_tests/system_tests_sync/test_requests.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+
+    # List Pub/Sub Topics through the REST API
+    # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/list
+    response = session.get("https://pubsub.googleapis.com/v1/projects/{}/topics".format(project_id))
+    response.raise_for_status()
+
+    # Check that self-signed JWT was created and is being used
+    assert credentials._jwt_credentials is not None
+    assert credentials._jwt_credentials.token == credentials.token
```
busunkim96, line 40: I initially tried to write this test with the Storage API, but requests seemed to fail with invalid credentials (a test with the service account passed).

Are there any APIs where the self-signed JWT approach won't work? @bshaffer

google/auth/transport/requests.py
```diff
@@ -313,6 +314,9 @@ def my_cert_callback():
             refreshing credentials. If not passed,
             an instance of :class:`~google.auth.transport.requests.Request`
             is created.
+        default_host (Optional[str]): A host like "pubsub.googleapis.com".
```
arithmetic1728, line 317: I have a little bit concern about setting the `default_host` in ctor. There are 2 default endpoints, for instance, `pubsub.googleapis.com` and `pubsub.mtls.googleapis.com`. mtls has auto switch logic to switch from the regular endpoint to mtls endpoint. It works as follows:

(1) client creates an AuthorizedSession
(2) client calls AuthorizedSession.is_mtls (with other conditions like the GOOGLE_API_USE_CLIENT_CERTIFICATE env var) to decide which endpoint to use

So this means in the AuthorizedSession ctor, we may not know exactly which default endpoint will be used. So if both endpoints use the regular endpoint as the audience, then this code works fine. Otherwise, we need to find a way. 

@bshaffer

arithmetic1728: mtls only works with user credentials, so this should be fine since service account credentials will be rejected by the server anyway.

arithmetic1728 (approved):
LGTM

PR #684, opened 2021-02-05
docs: update user guide/references for workload identity federation

bojeil-google:
Updates user guide to document using external account credentials with ADC or explicitly in various platforms.
Regenerates references for all classes and utilities related to this feature.


docs/reference/google.auth.credentials_async.rst
```diff
@@ -1,7 +0,0 @@
-google.auth.credentials\_async module
```
busunkim96, line 1: @crwilcox Should we keep the RSTs for the async files or is it alright to delete them for now?

crwilcox: Not sure I understand the 'for now' here? I would assume we still want them?

busunkim96: The way `docgen` is set up it seems to try to wipe these. Opened https://github.com/googleapis/google-auth-library-python/issues/685 to track.

busunkim96 (approved):

busunkim96:
CLAs were granted when changes were merged to the main branch.

PR #686, opened 2021-02-05
feat: workload identity federation support

bojeil-google:
Using workload identity federation, applications can access Google Cloud resources from Amazon Web Services (AWS), Microsoft Azure or any identity provider that supports OpenID Connect (OIDC). Workload identity federation is recommended for non-Google Cloud environments as it avoids the need to download, manage and store service account private keys locally.


busunkim96 (approved):

busunkim96:
CLAs were granted when the commits were merged to the master branch.

PR #694, opened 2021-02-11
fix: Fixes `google.auth.default()` passing scopes to authorized user creds

bojeil-google:
Reverted changes in `google.auth.default()` logic introduced in external account creds.
`_get_explicit_environ_credentials` is called with no arguments as previously done.
Instead after the credential is returned via `with_scopes_if_required()`, we try to
get project ID if the credential has a `get_project_id` method and if the project ID
is not yet been determined. This is a safer change while still keeping the logic to
determine project ID for external credentials.


busunkim96 (approved):

busunkim96:
CLAs were granted when commits were originally merged to `master`.

PR #695, opened 2021-02-12
setup: ban lower than rsa 4.7 where possible

RhinosF1:
help prevent https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-25658

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ DEPENDENCIES = (
     # rsa==4.5 is the last version to support 2.7
     # https://github.com/sybrenstuvel/python-rsa/issues/152#issuecomment-643470233
     'rsa<4.6; python_version < "3.6"',
-    'rsa>=3.1.4,<5; python_version >= "3.6"',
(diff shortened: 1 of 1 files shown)


busunkim96:
Hi @RhinosF1,

Thank you for opening a PR!

We try to keep pins in this library as wide as possible. This is minimize the chance folks run into diamond dependencies when they are using multiple Google libraries.

We recommend that end application developers explicitly pin every dependency (including transitive dependencies) and update them regularly, perhaps relying on something like https://github.com/renovatebot/renovate or dependabot https://docs.github.com/en/github/administering-a-repository/keeping-your-dependencies-updated-automatically. 

Additional note: If you are able to install `cryptography`, the library will prefer it over `python-rsa`.

https://github.com/googleapis/google-auth-library-python/blob/d4d7f3815e0cea3c9f39a5204a4f001de99568e9/google/auth/crypt/rsa.py#L18-L30

PR #697, opened 2021-02-12
fix: add pyopenssl as extra dependency

### LINKED ISSUE #551: Missing pyOpenSSL Dependency
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/google-auth-library-python/issues

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS:
  - Python version:
  - pip version:
  - `google-auth` version:

#### Steps to reproduce

  1. Missing pyOpenSSL dependency in setup.py

For the tests there is a requirement in https://github.com/googleapis/google-auth-library-python/blob/master/noxfile.py against pyOpenSSL. But there are imports for pyOpenSSL in multiple modules in the code. Should pyOpenSSL be added to the requirements in setup.py?

I created https://github.com/googleapis/google-auth-library-python/pull/550 with the proposal but wanted to get feedback from an issue first as I don't see this showing up in previous issues or pull requests.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

arithmetic1728:
fix #551 

`pyopenssl` is only needed if mTLS feature is enabled (disabled by default). To enable the feature, `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable has to be set to "true". More details see https://google.aip.dev/auth/4114.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ DEPENDENCIES = (
 
 extras = {
     "aiohttp": "aiohttp >= 3.6.2, < 4.0.0dev; python_version>='3.6'",
-    "pyopenssl": "pyopenssl",
+    "pyopenssl": "pyopenssl>=20.0.0",
 }
 
 with io.open("README.rst", "r") as fh:
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,10 @@ DEPENDENCIES = (
     "six>=1.9.0",
 )
 
-extras = {"aiohttp": "aiohttp >= 3.6.2, < 4.0.0dev; python_version>='3.6'"}
+extras = {
+    "aiohttp": "aiohttp >= 3.6.2, < 4.0.0dev; python_version>='3.6'",
+    "pyopenssl": "pyopenssl",
+}
 
 with io.open("README.rst", "r") as fh:
     long_description = fh.read()


setup.py
```diff
@@ -29,7 +29,10 @@
     "six>=1.9.0",
 )
 
-extras = {"aiohttp": "aiohttp >= 3.6.2, < 4.0.0dev; python_version>='3.6'"}
+extras = {
+    "aiohttp": "aiohttp >= 3.6.2, < 4.0.0dev; python_version>='3.6'",
+    "pyopenssl": "pyopenssl",
```
busunkim96, line 34: Please add a lower bound (`pyopenssl >= x.x.x`). See https://www.pyopenssl.org/en/stable/changelog.html 

Otherwise LGTM.

busunkim96 (approved):

setup.py
arithmetic1728: Yep, now made it `pyopenssl>=20.0.0`. In the change log, they mentioned they fixed a bug about `OpenSSL.crypto.load_privatekey` function (which we uses for mtls) in 20.0.0, so this version would be a good lower bound.

eviljeff:
Does this address #551?  (If not, can some further detail be added/linked?)

arithmetic1728:
> Does this address #551? (If not, can some further detail be added/linked?)

Yep.

Issue #700, opened 2021-02-16
Session argument for aiohttp request object is ignored


greshilov:
#### Environment details

  - OS: Linux
  - Python version: 3.9
  - pip version: 20.2.3
  - `google-auth` version: 1.26.1

#### Steps to reproduce
Consider the following example:
```python
import asyncio

import aiohttp
from google.auth.transport._aiohttp_requests import Request


async def main():
    async with aiohttp.ClientSession() as session:
        req = Request(session)
        assert req.session == session

if __name__ == '__main__':
    asyncio.run(main())

```
This code will raise `AssertionError`, because the session argument is ignored in the `Request` object.
https://github.com/googleapis/google-auth-library-python/blob/aeab5d07c5538f3d8cce817df24199534572b97d/google/auth/transport/_aiohttp_requests.py#L140-L141

This may also cause runtime warnings about unclosed sessions when `AuthorizedSession` object is used.

PR #701, opened 2021-02-16
fix: session object was never used in aiohttp request (#700)

greshilov:
Fix for #700


greshilov:
@googlebot I signed it!

arithmetic1728 (approved):

google/auth/transport/_aiohttp_requests.py
```diff
@@ -138,7 +138,12 @@ class Request(transport.Request):
     """
 
     def __init__(self, session=None):
-        self.session = None
+        # TODO: Use auto_decompress property for aiohttp 3.7+
+        if session is not None and session._auto_decompress:
+            raise ValueError(
+                "Client sessions with auto_decompress=True are not supported."
```
arithmetic1728, line 144: why raise the error here?

greshilov: Modifying the `session` object provided by the user is inappropriate in this place, so we can't disable `auto_decompress` ourselves. Remaining options are:

- Ignore the `session` object provided by the user and leave the `self.session = None`. Log error in console.
- Raise the error to stop execution and warn user, that's something went wrong with one's `session`.

I chose the latter option because it's better to **fail early** in my opinion. A user may not notice that one's `session` object is ignored.

arithmetic1728: Sorry I meant why auto_decompress=True are not supported. I am not familiar with aiohttp.

I think in this PR you also need to pass `auto_decompress=False` to https://github.com/googleapis/google-auth-library-python/blob/master/system_tests/system_tests_async/conftest.py#L29 to make the system test work.

greshilov: > Sorry I meant why auto_decompress=True are not supported. I am not familiar with aiohttp.
> 
> I think in this PR you also need to pass `auto_decompress=False` to https://github.com/googleapis/google-auth-library-python/blob/master/system_tests/system_tests_async/conftest.py#L29 to make the system test work.

Ah, `auto_decompress=True` is supported by `aiohttp`, but this library prefers to do decompression on its own:
https://github.com/googleapis/google-auth-library-python/blob/04c870a5bcc2c25433774a27280505d3413550e3/google/auth/transport/_aiohttp_requests.py#L80-L88

I've made required changes in `conftest.py`.

arithmetic1728: Ah I see. Thank you! The system test is still failing. I will take a look at why it failed.

greshilov:
@arithmetic1728 I hope I found the problem. `ASYNC_REQUESTS_SESSION` is created outside the `pytest-asyncio` context and thus assigned to the wrong event loop. 
```
ASYNC_REQUESTS_SESSION = aiohttp.ClientSession(auto_decompress=False)
```
I moved the session creation process to a separate fixture. Hope it helps.

arithmetic1728 (approved):

arithmetic1728:
@greshilov Great! Thank you!

PR #714, opened 2021-03-06
fix: fix unit tests so they can work in g3

arithmetic1728:
(1) It seems class method mock doesn't work in g3 if `autospec=True` is used.
(2) `request.call_args.kwargs` doesn't work in g3, should use `request.call_args[1]` instead

### DIFF
--- a/tests/oauth2/test_service_account.py
+++ b/tests/oauth2/test_service_account.py
@@ -203,7 +203,7 @@ class TestCredentials(object):
         assert "x-goog-user-project" not in headers
         assert "token" in headers["authorization"]
 
-    @mock.patch("google.auth.jwt.Credentials.from_signing_credentials", autospec=True)
+    @mock.patch("google.auth.jwt.Credentials.from_signing_credentials")
     def test__create_self_signed_jwt(self, from_signing_credentials):
         credentials = service_account.Credentials(
             SIGNER, self.SERVICE_ACCOUNT_EMAIL, self.TOKEN_URI
@@ -213,7 +213,7 @@ class TestCredentials(object):
         credentials._create_self_signed_jwt(audience)
         from_signing_credentials.assert_called_once_with(credentials, audience)
 
-    @mock.patch("google.auth.jwt.Credentials.from_signing_credentials", autospec=True)
+    @mock.patch("google.auth.jwt.Credentials.from_signing_credentials")
     def test__create_self_signed_jwt_with_user_scopes(self, from_signing_credentials):
         credentials = service_account.Credentials(
             SIGNER, self.SERVICE_ACCOUNT_EMAIL, self.TOKEN_URI, scopes=["foo"]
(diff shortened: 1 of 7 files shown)


tests/oauth2/test_service_account.py
```diff
@@ -203,7 +203,7 @@ def test_apply_with_no_quota_project_id(self):
         assert "x-goog-user-project" not in headers
         assert "token" in headers["authorization"]
 
-    @mock.patch("google.auth.jwt.Credentials.from_signing_credentials", autospec=True)
+    @mock.patch("google.auth.jwt.Credentials.from_signing_credentials")
```
busunkim96, line 206: Is there some other way to get an autospec'd object in g3?

arithmetic1728: I think it only affects class method, I am not familiar with autospec, not sure if it is needed. To autospec an object, I think we need to add `instance=True`.

busunkim96: I see. I'd prefer using `autospec` if possible. https://docs.python.org/3/library/unittest.mock.html#auto-speccing I _wasn't_ autospeccing in a set of tests elsewhere and it ended up hiding some issues.

arithmetic1728: Using `autospec=True` in these 2 tests causes the following error in g3 for python3 test. I think the reason might be g3 uses the python bundled mock lib but in github we use the standalone mock lib, so the behavior is different. I tried adding `instance=True`, it doesn't work either. Do you have any ideas of this error?

```
  File "/build/work/612862d9a9fce3bb1b5934686a005dc597bf/google3/runfiles/google3/third_party/py/google/auth/tests/oauth2/test_service_account.py", line 213, in test__create_self_signed_jwt
    credentials._create_self_signed_jwt(audience)
  File "/build/work/612862d9a9fce3bb1b5934686a005dc597bf/google3/runfiles/google3/third_party/py/google/oauth2/service_account.py", line 392, in _create_self_signed_jwt
    self, audience
TypeError: 'NonCallableMagicMock' object is not callable
```

busunkim96: Hmm maybe mocks can only be made on classes? Could you try this instead?

```py
    @mock.patch("google.auth.jwt.Credentials", instance=True, autospec=True)
    def test__create_self_signed_jwt(self, jwt):
        credentials = service_account.Credentials(
            SIGNER, self.SERVICE_ACCOUNT_EMAIL, self.TOKEN_URI
        )

        audience = "https://pubsub.googleapis.com"
        credentials._create_self_signed_jwt(audience)
        jwt.from_signing_credentials.assert_called_once_with(credentials, audience)
```

arithmetic1728: Yep, it works. Just updated the code. Thanks!

busunkim96 (approved):

PR #721, opened 2021-03-11
feat: allow the AWS_DEFAULT_REGION environment variable

ScruffyProdigy:
Amazon has this variable documented, and apparently people are trying to 
use it, so we should support it


tests/test_aws.py
```diff
@@ -1043,6 +1043,27 @@ def test_retrieve_subject_token_success_environment_vars(self, utcnow, monkeypat
             }
         )
 
+    @mock.patch("google.auth._helpers.utcnow")
+    def test_retrieve_subject_token_success_environment_vars(self, utcnow, monkeypatch):
```
bojeil-google, line 1047: This test has the same name as the above test.

tests/test_aws.py
```diff
@@ -1043,6 +1043,27 @@ def test_retrieve_subject_token_success_environment_vars(self, utcnow, monkeypat
...
+        monkeypatch.setenv(environment_vars.AWS_ACCESS_KEY_ID, ACCESS_KEY_ID)
+        monkeypatch.setenv(environment_vars.AWS_SECRET_ACCESS_KEY, SECRET_ACCESS_KEY)
+        monkeypatch.setenv(environment_vars.AWS_SESSION_TOKEN, TOKEN)
+        monkeypatch.setenv(environment_vars.AWS_DEFAULT_REGION, self.AWS_REGION)
```
bojeil-google, line 1051: Can you add one more test with both `AWS_DEFAULT_REGION` and `AWS_REGION` (give them different values). you can then confirm that `AWS_REGION` is used instead of the other one.

bojeil-google (changes requested):
I believe you need to run `blacken` to fix the linter errors.
You also need to update the functions descriptions, e.g. `_get_region` to mention that we also read `AWS_DEFAULT_REGION`.

tests/test_aws.py
ScruffyProdigy: done

tests/test_aws.py
ScruffyProdigy: done

busunkim96 (approved):

bojeil-google (approved):
Just one more thing. Can you update the descriptions to mention that we also support `AWS_DEFAULT_REGION`?
These places:
- https://github.com/googleapis/google-auth-library-python/blob/d80c85f285ae1a44ddc5a5d94a66e065a79f6d19/google/auth/aws.py#L427
- https://github.com/googleapis/google-auth-library-python/blob/d80c85f285ae1a44ddc5a5d94a66e065a79f6d19/google/auth/aws.py#L507

bojeil-google (approved):
Thanks for making all the changes!

PR #727, opened 2021-03-30
feat: add reauth feature to user credentials

arithmetic1728:
googlers see: go/guac-python-reauth

This PR does 2 things:

#### 1. Integrated [google-reauth-python](https://github.com/google/google-reauth-python) into google-auth library
Main APIs:
```
# API for getting rapt token
google.oauth2.reauth.get_rapt_token(request, client_id, client_secret, refresh_token, token_uri, scopes=None)

# API for refresh grant
google.oauth2.reauth.refresh_grant(request, token_uri, refresh_token, client_id, client_secret, scopes=None, rapt_token=None)
```

Supported challenge types: PASSWORD, SECURITY_KEY. To use SECURITY_KEY type, `pyu2f` dependency is required (either run `pip install pyu2f` or `pip install google-auth[reauth]`.

#### 2. Added reauth feature to user credentials (google.oauth2.credentials.Credentials)

If users opts in reauth (outside the scope of this lib), current google-auth version's refresh method will fail. With the added reauth feature, reauth flow will be triggered and the refresh method will work.
if users don't opt in reauth, then there is no difference w/o reauth feature because reauth won't be triggered.

Future PR will add reauth to async version of the credentials.

### DIFF
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -17,6 +17,8 @@
 import base64
 import calendar
 import datetime
+import getpass
+import sys
 
 import six
 from six.moves import urllib
@@ -230,3 +232,35 @@ def unpadded_urlsafe_b64encode(value):
         Union[str|bytes]: The encoded value
     """
     return base64.urlsafe_b64encode(value).rstrip(b"=")
+
+
+def get_user_password(text):
+    """Get password from user.
+
+    Override this function with a different logic if you are using this library
+    outside a CLI.
+
+    Args:
+        text (str): message for the password prompt.
+
+    Returns:
+        str: password string.
+    """
+    return getpass.getpass(text)
+
+
+def is_interactive():
+    """Check if we are in an interractive environment.
+
+    Override this function with a different logic if you are using this library
+    outside a CLI.
+
+    If the rapt token needs refreshing, the user needs to answer the challenges.
+    If the user is not in an interractive environment, the challenges can not
+    be answered and we just wait for timeout for no reason.
+
+    Returns:
+        bool: True if is interactive environment, False otherwise.
+    """
+
+    return sys.stdin.isatty()
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -17,8 +17,6 @@
 import base64
 import calendar
 import datetime
-import getpass
-import sys
 
 import six
 from six.moves import urllib
@@ -232,35 +230,3 @@ def unpadded_urlsafe_b64encode(value):
         Union[str|bytes]: The encoded value
     """
     return base64.urlsafe_b64encode(value).rstrip(b"=")
-
-
-def get_user_password(text):
-    """Get password from user.
-
-    Override this function with a different logic if you are using this library
-    outside a CLI.
-
-    Args:
-        text (str): message for the password prompt.
-
-    Returns:
-        str: password string.
-    """
-    return getpass.getpass(text)
-
-
-def is_interactive():
-    """Check if we are in an interractive environment.
-
-    Override this function with a different logic if you are using this library
-    outside a CLI.
-
-    If the rapt token needs refreshing, the user needs to answer the challenges.
-    If the user is not in an interractive environment, the challenges can not
-    be answered and we just wait for timeout for no reason.
-
-    Returns:
-        bool: True if is interactive environment, False otherwise.
-    """
-
-    return sys.stdin.isatty()
--- a/google/oauth2/_client.py
+++ b/google/oauth2/_client.py
@@ -35,29 +35,29 @@ from google.auth import exceptions
 from google.auth import jwt
 
 _URLENCODED_CONTENT_TYPE = "application/x-www-form-urlencoded"
+_JSON_CONTENT_TYPE = "application/json"
 _JWT_GRANT_TYPE = "urn:ietf:params:oauth:grant-type:jwt-bearer"
 _REFRESH_GRANT_TYPE = "refresh_token"
 
 
-def _handle_error_response(response_body):
-    """"Translates an error response into an exception.
+def _handle_error_response(response_data):
+    """Translates an error response into an exception.
 
     Args:
-        response_body (str): The decoded response data.
+        response_data (Mapping): The decoded response data.
 
     Raises:
-        google.auth.exceptions.RefreshError
+        google.auth.exceptions.RefreshError: The errors contained in response_data.
     """
     try:
-        error_data = json.loads(response_body)
         error_details = "{}: {}".format(
-            error_data["error"], error_data.get("error_description")
+            response_data["error"], response_data.get("error_description")
         )
     # If no details could be extracted, use the response data.
     except (KeyError, ValueError):
-        error_details = response_body
+        error_details = json.dumps(response_data)
 
-    raise exceptions.RefreshError(error_details, response_body)
+    raise exceptions.RefreshError(error_details, response_data)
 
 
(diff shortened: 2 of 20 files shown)


google/auth/_helpers.py
```diff
@@ -230,3 +232,35 @@ def unpadded_urlsafe_b64encode(value):
         Union[str|bytes]: The encoded value
     """
     return base64.urlsafe_b64encode(value).rstrip(b"=")
+
+
+def get_user_password(text):
+    """Get password from user.
+
+    Override this function with a different logic if you are using this library
```
busunkim96, line 240: Are users expected to override this function? (If so, it seems odd that it is in the private `_helpers.py`)

google/oauth2/challenges.py
```diff
@@ -0,0 +1,139 @@
...
+    @abc.abstractmethod
+    def name(self):  # pragma: NO COVER
+        """Returns the name of the challenge."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def is_locally_eligible(self):  # pragma: NO COVER
+        """Returns true if a challenge is supported locally on this machine."""
+        pass
```
busunkim96, line 45: Can the abstract methods raise `NotImplementedError` for consistency with the existing abstract methods in this library?

https://github.com/googleapis/google-auth-library-python/blob/5832fc1626cf16f6862b4dc5f7ad97dc4b590e3d/google/auth/credentials.py#L84-L99

google/oauth2/_client.py
```diff
@@ -78,25 +78,35 @@ def _parse_expiry(response_data):
...
+        Tuple(bool, Mapping[str, str]): A boolean indicating if the request is
+            successful, and a mapping for the JSON-decoded response data.
     """
-    body = urllib.parse.urlencode(body).encode("utf-8")
-    headers = {"content-type": _URLENCODED_CONTENT_TYPE}
+    if use_json:
+        headers = {"Content-Type": _JSON_CONTENT_TYPE}
+        body = json.dumps(body).encode("utf-8")
+    else:
+        headers = {"content-type": _URLENCODED_CONTENT_TYPE}
```
busunkim96, line 105: Style nit, could you use `content-type` or `Content-Type` in both places?

google/oauth2/challenges.py
```diff
@@ -0,0 +1,139 @@
...
+        Args:
+            metadata: challenge metadata returned in the 'challenges' field in
+                the initial reauth request. Includes the 'challengeType' field
+                and other challenge-specific fields.
+
+        Returns:
+            response that will be send to the reauth service as the content of
+            the 'proposalResponse' field in the request body. Usually a dict
+            with the keys specific to the challenge. For example,
+            {'credential': password} for password challenge.
```
busunkim96, line 60:
```suggestion
            ``{'credential': password}`` for password challenge.
```

google/oauth2/challenges.py
```diff
@@ -0,0 +1,139 @@
...
+    def is_locally_eligible(self):  # pragma: NO COVER
+        """Returns true if a challenge is supported locally on this machine."""
+        pass
+
+    @abc.abstractmethod
+    def obtain_challenge_input(self, metadata):  # pragma: NO COVER
+        """Performs logic required to obtain credentials and returns it.
+
+        Args:
+            metadata: challenge metadata returned in the 'challenges' field in
```
busunkim96, line 52: Please add a type in the docstring

google/oauth2/challenges.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @property
+    def name(self):
+        return "PASSWORD"
+
+    @property
+    def is_locally_eligible(self):
+        return True
+
+    def obtain_challenge_input(self, unused_metadata):
```
busunkim96, line 76: There's a helper to copy docstrings. Could you use that for these methods?     https://github.com/googleapis/google-auth-library-python/blob/f1fee1f4c3d511d9e6ecbc1c0397e743bf2583db/google/auth/app_engine.py#L56

google/oauth2/credentials.py
```diff
@@ -205,23 +219,31 @@ def refresh(self, request):
...
         )
 
         self.token = access_token
         self.expiry = expiry
         self._refresh_token = refresh_token
         self._id_token = grant_response.get("id_token")
+        self._rapt_token = rapt_token
 
-        if scopes and "scopes" in grant_response:
+        if scopes and "scope" in grant_response:
```
busunkim96, line 244: Was this wrong earlier? (wondering about the `scopes` -> `scope` change)

google/oauth2/reauth.py
```diff
@@ -0,0 +1,325 @@
...
+                " in an interactive session."
+            )
+
+        msg = _run_next_challenge(msg, request, access_token)
+
+        if msg["status"] == _AUTHENTICATED:
+            return msg["encodedProofOfReauthToken"]
+
+    # If we got here it means we didn't get authenticated.
+    raise exceptions.ReauthFailError()
```
busunkim96, line 216: Could a message be added here?

busunkim96 (changes requested):
Is pyu2f owned by a team? It looks like it hasn't been touched since last October. https://github.com/google/pyu2f, https://github.com/google/pyu2f

arithmetic1728:
> Is pyu2f owned by a team? It looks like it hasn't been touched since last October. https://github.com/google/pyu2f, https://github.com/google/pyu2f

Yes gnubby team owns it.

google/oauth2/reauth.py
arithmetic1728: done

google/oauth2/credentials.py
arithmetic1728: Yes it is a bug. Response returns `scope` not `scopes`, see step 5 in [google oauth2 doc](https://developers.google.com/identity/protocols/oauth2/web-server) or [oauth2 spec section 5.1](https://tools.ietf.org/html/rfc6749#section-5.1)

The following is the example from the doc:
```
{
  "access_token": "1/fFAGRNJru1FTz70BzhT3Zg",
  "expires_in": 3920,
  "token_type": "Bearer",
  "scope": "https://www.googleapis.com/auth/drive.metadata.readonly",
  "refresh_token": "1//xEoDL4iW3cxlI7yDbSRFYNG01kVKM2C-259HOF2aQbI"
}
```

google/oauth2/challenges.py
arithmetic1728: done

google/oauth2/challenges.py
arithmetic1728: done

google/oauth2/challenges.py
arithmetic1728: done

google/oauth2/_client.py
arithmetic1728: Changed to `Content-Type`.

google/oauth2/challenges.py
arithmetic1728: done

google/auth/_helpers.py
arithmetic1728: Moved the two methods to where they are used, i.e., `get_user_password` to `challenges.py`, and `is_interactive` to `reauth.py`.

busunkim96 (approved):

jay0lee:
Note that it doesn't seem like it's currently possible for GCP projects to enable the reauth API in the cloud console making it impossible to use reauth. See:

https://github.com/googleapis/google-auth-library-python/issues/742

PR #736, opened 2021-04-14
chore: prevent normalization of semver versioning

dandhlee:
When there is a patch version added to semver versioning, setuptools.setup(version) will normalize the versioning from `-patch` to `.patch` which is not correct SEMVER versioning. The added feature with setuptools.sic(version) will prevent this from happening.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ version = version["__version__"]
 
 setup(
     name="google-auth",
-    version=version,
+    version=setuptools.sic(version),
     author="Google Cloud Platform",
     author_email="googleapis-packages@google.com",
     description="Google Authentication Library",
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,22 @@ import os
 
 import setuptools
 
+# Disable version normalization performed by setuptools.setup()
+# Adding this in even though it works for Python3.x, but does not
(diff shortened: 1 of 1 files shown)


parthea (dismissed):

setup.py
```diff
@@ -15,9 +15,24 @@
...
+except ImportError:
+    # Try the approach of replacing packaging.version.Version
+    sic = lambda v: v
+    try:
+        # setuptools >=39.0.0 uses packaging from setuptools.extern
+        from setuptools.extern import packaging
+    except ImportError:
+        # setuptools <39.0.0 uses packaging from pkg_resources.extern
+        from pkg_resources.extern import packaging
+    packaging.version.Version = packaging.version.LegacyVersion
```
parthea, line 35: Instead of the workaround, let's bump the `setuptools` version since it's a dependency.
https://github.com/googleapis/google-auth-library-python/blob/master/setup.py#L29

dandhlee: I tried doing that, it works for Python3 but does not work for Python2 :(

busunkim96: Ah it must be from setuptools removing support for Python 2. This LGTM.

busunkim96 (approved):

PR #751, opened 2021-05-18
fix: fix function name in signing error message

akx:
The name of the API is `signBlob`, not `signBytes`.

See https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signBlob


busunkim96 (approved):

busunkim96:
Thank you for the patch @akx!

PR #760, opened 2021-05-25
fix: enforce constraints during unit tests

### LINKED ISSUE #759: 'testing/constraints-*.txt' not used
From #757.

In [`noxfile.py`](https://github.com/googleapis/google-auth-library-python/blob/a9234423cb2b69068fc0d30a5a0ee86a599ab8b7/noxfile.py#L87-L89), the constraints are not being used, which is why the one populated file, [`testing/constraints-3.6.txt](https://github.com/googleapis/google-auth-library-python/blob/master/testing/constraints-3.6.txt) is in an invalid state (multiple entries for `rsa`, maybe more issues).

tseaver:
Drop explicit pin / constraint on `urllib3`:  specific `requests` versions have very narrow pins, and ours is only likely to create conflicts.

Bump the `requests` lower bound to `2.20.0`, the lowest version for which our tests pass once constraints are being checked.

Closes #759

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -12,9 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import shutil
 import os
+import pathlib
+import shutil
+
 import nox
+
+CURRENT_DIRECTORY = pathlib.Path(__file__).parent.absolute()
 
 TEST_DEPENDENCIES = [
     "flask",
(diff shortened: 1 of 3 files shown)


busunkim96 (approved):
Thank you @tseaver!

noxfile.py
```diff
@@ -123,7 +132,7 @@ def cover(session):
         "--cov=google.oauth2",
         "--cov=tests",
         "--cov=tests_async",
-        "--cov-report=",
+        "--cov-report=term-missing",
```
jimfulton, line 135: Why?

jimfulton (approved):

noxfile.py
tseaver: If lines aren't being covered, then the report should include them, so that the hapless developer can figure out why.  I've never understood why that isn't the default behavior for the report.

PR #770, opened 2021-06-09
feat: define useful properties on `google.auth.external_account.Credentials`

bojeil-google:
This includes the following properties:

- `info`: This is the reverse of `from_info` defined on subclasses and useful to
  serialize external account credentials.
- `service_account_email`: This is the corresponding service account email if impersonation is used.
- `is_user`: This is `False` for workload identity pools and `True` for workforce pools (not yet supported).
  This can be mainly determined from the STS audience.

While the properties will primarily facilitate integration with gcloud, they are publicly useful for other contexts.

### DIFF
--- a/google/auth/external_account.py
+++ b/google/auth/external_account.py
@@ -28,8 +28,10 @@ token exchange endpoint following the `OAuth 2.0 Token Exchange`_ spec.
 """
 
 import abc
+import copy
 import datetime
 import json
+import re
 
 import six
 
@@ -40,6 +42,8 @@ from google.auth import impersonated_credentials
 from google.oauth2 import sts
 from google.oauth2 import utils
 
+# External account JSON type identifier.
+_EXTERNAL_ACCOUNT_JSON_TYPE = "external_account"
 # The token exchange grant_type used for exchanging credentials.
 _STS_GRANT_TYPE = "urn:ietf:params:oauth:grant-type:token-exchange"
 # The token exchange requested_token_type. This is always an access_token.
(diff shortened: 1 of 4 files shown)


busunkim96 (approved):

google/auth/external_account.py
```diff
@@ -117,6 +121,76 @@ def __init__(
...
+            "service_account_impersonation_url": self._service_account_impersonation_url,
+            "credential_source": copy.deepcopy(self._credential_source),
+            "quota_project_id": self._quota_project_id,
+            "client_id": self._client_id,
+            "client_secret": self._client_secret,
+        }
+        # Remove None fields in the info dictionary.
+        for k, v in dict(config_info).items():
+            if v is None:
+                del config_info[k]
```
tseaver, line 148: Hmm, deleting an item from a dictionary while iterating it is unsafe.  It would be better to create a new one, e.g.:

```python
return {key: value for key value in config_info.items() if value is not None}
```

google/auth/external_account.py
```diff
@@ -117,6 +121,76 @@ def __init__(
...
+        if self._service_account_impersonation_url:
+            # Parse email from URL. The formal looks as follows:
+            # https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/name@project-id.iam.gserviceaccount.com:generateAccessToken
+            url = self._service_account_impersonation_url
+            start_index = url.rfind("/")
+            end_index = url.find(":generateAccessToken")
+            if start_index != -1 and end_index != -1 and start_index < end_index:
+                start_index = start_index + 1
+                return url[start_index:end_index]
+        return None
```
tseaver, line 169: Although it is unusual that I would say so, this code would be clearer using a regex:

```python
service_account_regex = re.compile(r".*/(?P<email>.*):generateAccessToken")
...
        if self._service_account_impersonation_url:
            match = service_account_regex.match(self._service_account_impersonation_url)
            if match is not None:
                return match.group("email")
```

bojeil-google: I would disagree on this one. It is better/more efficient to use string methods when possible over regex. It does require a couple more steps to do so here but I provided an example for the format used. It should make it easier to understand and follow the code.

tseaver: Note that you used a regex in a very similar case [below](https://github.com/googleapis/google-auth-library-python/pull/770/files#diff-002ea4a611abe1254cc4abc68bbf57f139a0ca17ff5389ab287e9a732bf44267R189), except without pulling the value from the group.  Efficiency here is likely not germane, and debatable: it would need measuring, if we thought it important;  clarity (to the reader) is far more crucial.

google/auth/external_account.py
```diff
@@ -117,6 +121,76 @@ def __init__(
...
+            bool: True if the credentials represent a user. False if they represent a
+                workload.
+        """
+        # If service account impersonation is used, the credentials will always represent a
+        # service account.
+        if self._service_account_impersonation_url:
+            return False
+        # Workforce pools representing users have the following audience format:
+        # //iam.googleapis.com/locations/$location/workforcePools/$poolId/providers/$providerId
+        p = re.compile(r"//iam\.googleapis\.com/locations/[^/]+/workforcePools/")
```
tseaver, line 189: This regex should likely be compiled at module scope, rather than on each call (yes, I know the `re` module interns some number of recent compilations).

google/auth/external_account.py
bojeil-google: I appreciate your input but I think this is a very subjective situation. I would change it but I think this is a coin toss.
The other pattern I am checking for below would be much harder to achieve without a regex. In this specific case, the implementation is simpler to read and follow, especially when paired with an example.

PR #776, opened 2021-06-14
feat: allow scopes for self signed jwt

arithmetic1728:
This doc implements https://google.aip.dev/auth/4111. Internal doc: go/yoshi-self-signed-jwt-phase-2.

The main feature here is now we can use `scope` claim in self signed JWT.

This PR does the following 2 things:
(1) Add `always_use_jwt_access` property to service account credentials to allow opt-in for the feature.
(2) If `always_use_jwt_access` is True, then apply the following logic. If `alwaysUseJwtAccess` is False, the logic is the same as before; if True, then we can always use self signed jwt with scopes or audience.
```
if (alwaysUseJwtAccess):
    if (scope):
        // create a self signed JWT with "scope" set to the scope
    else if (audience):
        // create a self signed JWT with "aud" set to the audience
    else if (defaultScope):
        // create a self signed JWT with "scope" set to the defaultScope
else:
    if (scope):
        // call OAuth token endpoint
    else if (audience):
        // create a self signed JWT with audience
    else if (defaultScope):
        // call OAuth token endpoint
```

This PR has been tested with python-kms, https://github.com/googleapis/python-kms/pull/122

The follow up PR in python microgenerator is: https://github.com/googleapis/gapic-generator-python/pull/920

### DIFF
--- a/google/auth/jwt.py
+++ b/google/auth/jwt.py
@@ -525,8 +525,9 @@ class Credentials(
             "sub": self._subject,
             "iat": _helpers.datetime_to_secs(now),
             "exp": _helpers.datetime_to_secs(expiry),
-            "aud": self._audience,
         }
+        if self._audience:
+            payload.update({"aud": self._audience})
 
         payload.update(self._additional_claims)
 
--- a/google/auth/jwt.py
+++ b/google/auth/jwt.py
@@ -527,7 +527,7 @@ class Credentials(
             "exp": _helpers.datetime_to_secs(expiry),
         }
         if self._audience:
-            payload.update({"aud": self._audience})
+            payload["aud"] = self._audience
 
         payload.update(self._additional_claims)
 
--- a/google/oauth2/service_account.py
+++ b/google/oauth2/service_account.py
@@ -432,7 +432,7 @@ class Credentials(
                     None,
                     additional_claims={"scope": " ".join(self._default_scopes)},
                 )
-        elif not self._scopes and not self._default_scopes and audience:
+        elif not self._scopes and audience:
             self._jwt_credentials = jwt.Credentials.from_signing_credentials(
                 self, audience
             )
--- a/google/oauth2/service_account.py
+++ b/google/oauth2/service_account.py
@@ -131,6 +131,7 @@ class Credentials(
         project_id=None,
         quota_project_id=None,
         additional_claims=None,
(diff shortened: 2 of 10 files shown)


google/auth/jwt.py
```diff
@@ -525,8 +525,9 @@ def _make_jwt(self):
             "sub": self._subject,
             "iat": _helpers.datetime_to_secs(now),
             "exp": _helpers.datetime_to_secs(expiry),
-            "aud": self._audience,
         }
+        if self._audience:
+            payload.update({"aud": self._audience})
```
tseaver, line 530: Simpler:
```suggestion
            payload["aud"] = self._audience]
```

google/oauth2/service_account.py
```diff
@@ -386,8 +417,22 @@ def _create_self_signed_jwt(self, audience):
...
-        if not self.scopes:
+        if self._always_use_jwt_access:
+            if self._scopes:
+                self._jwt_credentials = jwt.Credentials.from_signing_credentials(
+                    self, None, additional_claims={"scope": " ".join(self._scopes)}
+                )
+            elif audience:
+                self._jwt_credentials = jwt.Credentials.from_signing_credentials(
+                    self, audience
+                )
```
busunkim96, line 428: I think the way `google-api-core` is written it will always pass through an `audience`. https://github.com/googleapis/python-api-core/blob/155da5e18cc2fdcfa57de6f956b7d078e79cd4b7/google/api_core/grpc_helpers.py#L249-L251 and `elif self._default_scopes` will never be reached.  

Is it necessary to distinguish between the "default" audience and a user defined audience? It looks like Cody asked a question about this in the doc as well.

arithmetic1728: I think it is fine that `self._default_scopes` is never reached. We can just leave it here for logic completeness.

We don't need to distinguish a default audience and a user defined audience. If the user doesn't provide the scope but provides the audience, then they are responsible that the audience is correct.

google/auth/jwt.py
arithmetic1728, line 530:
```suggestion
            payload["aud"] = self._audience
```

arithmetic1728: done

busunkim96 (approved):

PR #784, opened 2021-06-25
feat: service account is able to use a private token endpoint

liuchaoren:
In  [Private Service Connect](https://cloud.google.com/vpc/docs/private-service-connect), users can use an endpoint which is private to their VPC network. The request is eventually routed to the oauth2.googleapis.com/token so the "aud" in the assertion still should be oauth2.googleapis.com/token. 

After this change, service account can send requests to the private endpoint (if configured) and still use the oauth2.googleapis.com/token in the assertion.


busunkim96:
@silvolu @arithmetic1728 PTAL. The corresponding bug is 190841374.

arithmetic1728 (approved):

silvolu:
Please wait for me on this.

tseaver:
Tagged as `do not merge` per request of @silvolu.

silvolu:
The fix should be simpler here: when exchanging a JWT for an access token, [the value of the 'aud' field must be `https://oauth2.googleapis.com/token`](https://developers.google.com/identity/protocols/oauth2/service-account#required-claims), so in [_make_authorization_grant_assertion](https://github.com/googleapis/google-auth-library-python/blob/master/google/oauth2/service_account.py#L385) the value should be read from a constant or hardcoded string instead of set to the token uri.

arithmetic1728:
> The fix should be simpler here: when exchanging a JWT for an access token, [the value of the 'aud' field must be `https://oauth2.googleapis.com/token`](https://developers.google.com/identity/protocols/oauth2/service-account#required-claims), so in [_make_authorization_grant_assertion](https://github.com/googleapis/google-auth-library-python/blob/master/google/oauth2/service_account.py#L385) the value should be read from a constant or hardcoded string instead of set to the token uri.

The token uri is not hardcoded because

(1) google-auth allows users to provide their own token uri (even though it should be always `https://oauth2.googleapis.com/token` for googleapis)

https://github.com/googleapis/google-auth-library-python/blob/22affe15b96071abd73faa1baec4f6eb0b6556cd/google/oauth2/service_account.py#L123-L127

(2) if the creds is created from service_account.json file, then google-auth reads token uri from the json file

https://github.com/googleapis/google-auth-library-python/blob/22affe15b96071abd73faa1baec4f6eb0b6556cd/google/oauth2/service_account.py#L237-L240

silvolu:
I'm not saying that the token_uri should be hardcoded, I'm saying that the audience field should be hardcoded. This class is not a generic implementation, the description clearly says "This module implements the JWT Profile for OAuth 2.0 Authorization Grants as defined by `RFC 7523`_ with particular support for how this RFC is implemented in Google's infrastructure."

In Google's infrastructure, when exchanging a JWT for a token, the audience is always `https://oauth2.googleapis.com/token`. Once that is done, you can actually use different token uris supported by the google infrastructure, e.g. a Private Secure Connect endpoint, because with the audience set to the token_uri, the audience check at the token exchange endpoint fails.

arithmetic1728:
> I'm not saying that the token_uri should be hardcoded, I'm saying that the audience field should be hardcoded. This class is not a generic implementation, the description clearly says "This module implements the JWT Profile for OAuth 2.0 Authorization Grants as defined by `RFC 7523`_ with particular support for how this RFC is implemented in Google's infrastructure."
> 
> In Google's infrastructure, when exchanging a JWT for a token, the audience is always `https://oauth2.googleapis.com/token`. Once that is done, you can actually use different token uris supported by the google infrastructure, e.g. a Private Secure Connect endpoint, because with the audience set to the token_uri, the audience check at the token exchange endpoint fails.

Sorry I misunderstood your comment. Yea, this way is much simpler.

liuchaoren:
I updated the PR to hard-code the aud which is much simpler. Please take another look.

arithmetic1728:
looks good to me

silvolu (approved):

liuchaoren:
Hi @tseaver, is it okay to remove the "do not merge" tag given that @silvolu approved it? I confirmed with @silvolu yesterday.

busunkim96:
@liuchaoren Yes I believe Tres added that to prevent an early merge. I've removed the label.

busunkim96:
@liuchaoren It looks like some unit tests are failing, could you take a look?

To run the unit tests locally,  `pip install nox` and run one of the unit test sessions: `nox -s unit-3.6`

PR #785, opened 2021-06-27
fix: comment and uncompress data before loads

b3g00d:
source code:

```
from google.oauth2 import _id_token_async as id_token
from google.auth.transport import _aiohttp_requests as requests


class GoogleClient:
    def __init__(self, client_id):
        self.req = requests.Request()
        self.client_id = client_id

    async def verify_token(self, token):
        id_info = await id_token.verify_token(token, self.req, self.client_id)
        return id_info

import unittest
from unittest import IsolatedAsyncioTestCase


class Test(IsolatedAsyncioTestCase):

    async def test_error_on_fetch_cert(self):
        g_c = GoogleClient("test")
        await g_c.verify_token("test_token")


if __name__ == "__main__":
    unittest.main()

```

It will raise error at function `_fetch_certs` because of the response return as a bytes array, but `json.dumps(data)` is trying to dumps the bytes array

```
ERROR: test_error_on_fetch_cert (__main__.Test)     
----------------------------------------------------------------------
  File "/usr/lib/python3.8/unittest/async_case.py", line 65, in _callTestMethod                          
    self._callMaybeAsync(method)
  File "/usr/lib/python3.8/unittest/async_case.py", line 88, in _callMaybeAsync                          
    return self._asyncioTestLoop.run_until_complete(fut)                                              
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()                                                                               
  File "/usr/lib/python3.8/unittest/async_case.py", line 102, in _asyncioLoopRunner
    ret = await awaitable
  File "core/connections/google_client.py", line 22, in test_error_on_fetch_cert
    await g_c.verify_token("test_token")
  File "core/connections/google_client.py", line 11, in verify_token
    id_info = await id_token.verify_token(token, self.req, self.client_id)
  File "/home/begood/.local/share/virtualenvs/BE-fy1aetm7/lib/python3.8/site-packages/google/oauth2/_id_token_async.py", line 119, in verify_token
    certs = await _fetch_certs(request, certs_url)
  File "/home/begood/.local/share/virtualenvs/BE-fy1aetm7/lib/python3.8/site-packages/google/oauth2/_id_token_async.py", line 98, in _fetch_certs
    return json.loads(json.dumps(data)) 
  File "/usr/lib/python3.8/json/__init__.py", line 231, in dumps    
    return _default_encoder.encode(obj)                                                                  
  File "/usr/lib/python3.8/json/encoder.py", line 199, in encode                                                                                                                                                   
    chunks = self.iterencode(o, _one_shot=True)   
  File "/usr/lib/python3.8/json/encoder.py", line 257, in iterencode                                                                                                                                               
    return _iterencode(o, 0)                                                                             
  File "/usr/lib/python3.8/json/encoder.py", line 179, in default                                                                                                                                                  
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```

I looked into the source code, and the `response.content()` work as expected.


b3g00d:
@googlebot I signed it!

tseaver (commented):
The docstring changes look fine.

For the change to `_fetch_certs`, two things:
- Is there a bug report open for which this change is a fix?  If not, can you please add one.
- We almost certainly need to make correspondng changes to tests.

b3g00d:
> The docstring changes look fine.
> 
> For the change to `_fetch_certs`, two things:
> 
> * Is there a bug report open for which this change is a fix?  If not, can you please add one.
> * We almost certainly need to make correspondng changes to tests.

@tseaver updated

tseaver:
@b3g00d 
>  updated

I don't see any further commits here, and I don't see a related issue.

b3g00d:
> @b3g00d
> 
> > updated
> 
> I don't see any further commits here, and I don't see a related issue.

Oh, I mean I updated the first comment, should I put it to commit message.

google/oauth2/_id_token_async.py
```diff
@@ -93,9 +93,9 @@ async def _fetch_certs(request, certs_url):
             "Could not fetch certificates at {}".format(certs_url)
         )
 
-    data = await response.data.read()
+    data = await response.content()
 
-    return json.loads(json.dumps(data))
+    return json.loads(data)
```
tseaver, line 98: @crwilcox Can you comment on this implementation?  The non-async version just does:

```python
return json.loads(response.data.decode("utf-8"))
```

I'm surprised that the synchronous version, at least, does not use `response.json`.

UlisseMini:
I had the same problem, changing
```py
    data = await response.data.read()

    return json.loads(json.dumps(data))
```
to
```py
    data = await response.content()

    return json.loads(data.decode('utf-8'))
```
[Here](https://github.com/googleapis/google-auth-library-python/blob/main/google/oauth2/_id_token_async.py#L96) fixes it, this is because `auth/transport/_aiohttp_requests.py`'s `_CombinedResponse` [has](https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/transport/_aiohttp_requests.py#L71)
```py
    @property
    def data(self):
        return self._response.content
```
Which never uses `_is_compressed` to check for compression (unlike `content`).

I don't know why `json.loads(json.dumps(data))` was there, it seems like a bad form of decoding.

The only test for `.data` is [here](https://github.com/googleapis/google-auth-library-python/blob/main/tests_async/transport/test_aiohttp_requests.py#L106) which doesn't specify if it should decompress or not, the other tests are for `.content()` and `.raw_content()`.

google/oauth2/_id_token_async.py
gpalasti: I run into the same error.
Async library does not perform automatic decompression.
response.data.read() returns compressed bytes.
response.content() returns decompressed data as string

chalmerlowe:
Hello! I requested a review of this PR and the current state of the main branch by Jules (Gemini). Closing, based on the following assessment.

---------------------------
It appears this PR is no longer needed because the underlying issue has been resolved in the codebase:

**Issue Resolved**: The reported TypeError was caused by json.dumps(data) receiving bytes. The current version of `google/oauth2/_id_token_async.py` in `main` no longer calls `json.dumps(data)`; it uses `json.loads(data)` directly.

**Verification**: I (Gemini) verified with a reproduction script that the current implementation works correctly. `response.content()` returns decompressed bytes (handled by `_CombinedResponse` in `google/auth/transport/_aiohttp_requests.py`), and `json.loads(data)` successfully parses those bytes without needing the explicit .decode('utf-8') proposed in this PR.
Given that the problematic code has been removed and the current implementation is verified to work, this PR appears to be obsolete.
```

PR #793, opened 2021-07-02
feat: define `CredentialAccessBoundary` classes

bojeil-google:
Defines the following classes:
- `google.auth.downscoped.CredentialAccessBoundary`
- `google.auth.downscoped.AccessBoundaryRule`
- `google.auth.downscoped.AvailabilityCondition`

This is based on [Downscoping with Credential Access Boundaries](https://cloud.google.com/iam/docs/downscoping-short-lived-credentials).
These classes help define the list of access boundary rules,
each of which contains information on the resource that the rule
applies to, the upper bound of the permissions that are available
on that resource and an optional condition to further restrict
permissions.


bojeil-google:
Note to reviewers. The failing tests seem to be unrelated to this change.

arithmetic1728:
@busunkim96 Seems the refresh token is expired/revoked again for the system test. Does the user account have token rotations now?

busunkim96:
@arithmetic1728 Nope the process for getting that approved has been rather slow. I'll rotate it again in a separate PR.

arithmetic1728 (approved):

PR #801, opened 2021-07-09
feat: define `google.auth.downscoped.Credentials` class

bojeil-google:
This is based on [Downscoping with Credential Access Boundaries](https://cloud.google.com/iam/docs/downscoping-short-lived-credentials).
The new credentials are initialized mainly using elevated source
credentials and a `google.auth.downscoped.CredentialAccessBoundary`
instance.
The credentials will then get access tokens from the source
credentials and exchange them via the GCP STS token exchange
endpoint using the provided credentials access boundary rules
for downscoped access tokens.

The new credentials will inherit the source credentials' scopes
but the scopes are not exposed as we cannot always determine the
scopes form the source credentials.

### DIFF
--- a/tests/test_downscoped.py
+++ b/tests/test_downscoped.py
@@ -489,13 +489,13 @@ class TestCredentials(object):
     def test_default_state(self):
         credentials = self.make_credentials()
 
-        # Not token acquired yet.
+        # No token acquired yet.
         assert not credentials.token
         assert not credentials.valid
         # Expiration hasn't been set yet.
         assert not credentials.expiry
         assert not credentials.expired
-        # Not quota project ID set.
+        # No quota project ID set.
         assert not credentials.quota_project_id
 
     def test_with_quota_project(self):
--- a/tests/test_downscoped.py
+++ b/tests/test_downscoped.py
@@ -12,9 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pytest
+import datetime
+import json
 
+import mock
+import pytest
+from six.moves import http_client
+from six.moves import urllib
+
+from google.auth import _helpers
+from google.auth import credentials
 from google.auth import downscoped
+from google.auth import exceptions
+from google.auth import transport
 
 
 EXPRESSION = (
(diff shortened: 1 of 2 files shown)


arithmetic1728 (approved):

tests/test_downscoped.py
```diff
@@ -383,3 +410,256 @@ def test_to_json(self):
...
+                    "availableResource": AVAILABLE_RESOURCE,
+                    "availabilityCondition": {
+                        "expression": EXPRESSION,
+                        "title": TITLE,
+                        "description": DESCRIPTION,
+                    },
+                }
+            ]
+        }
+    }
```
tseaver, line 446: Is there a particular reason to have these constants at class scope, rather than at module scope?

tests/test_downscoped.py
```diff
@@ -383,3 +410,256 @@ def test_to_json(self):
...
+
+    @classmethod
```
tseaver, line 448: `@classmethod` is an unusual choice here:  it implies that you expect to override the current class, supplying different class-level attribute values (or methods).  Given that you don't use `cls` here, I think you probably want `@staticmethod` instead.

tests/test_downscoped.py
```diff
@@ -383,3 +410,256 @@ def test_to_json(self):
...
+            AVAILABLE_RESOURCE, AVAILABLE_PERMISSIONS, availability_condition
+        )
+        rules = [access_boundary_rule]
+        credential_access_boundary = make_credential_access_boundary(rules)
+
+        return downscoped.Credentials(
+            source_credentials, credential_access_boundary, quota_project_id
+        )
+
+    @classmethod
```
tseaver, line 465: Likewise here.

tests/test_downscoped.py
```diff
@@ -383,3 +410,256 @@ def test_to_json(self):
...
+        response = mock.create_autospec(transport.Response, instance=True)
+        response.status = status
+        response.data = json.dumps(data).encode("utf-8")
+
+        request = mock.create_autospec(transport.Request)
+        request.return_value = response
+
+        return request
+
+    @classmethod
```
tseaver, line 476: And here.

tseaver (changes requested):

tests/test_downscoped.py
bojeil-google: I use module scope constants when I need them in multiple test classes (such as the ones above). In this case, they are used for a single test class. I can move them to module level but there is no consistent style in the repo.
Example: https://github.com/googleapis/google-auth-library-python/blob/d3944af36737dd7dc6a2a96932de5ae54f4e218a/tests/test_impersonated_credentials.py#L98

tests/test_downscoped.py
bojeil-google: Done.

tests/test_downscoped.py
bojeil-google: Done.

tests/test_downscoped.py
bojeil-google: I do use `cls` in this method so this may need to remain as a `classmethod`.

tseaver: OK.  Of course, moving the constants to module scope would remove that bit. :)

tests/test_downscoped.py
bojeil-google: I moved them to module scope.

tests/test_downscoped.py
bojeil-google: Changed to `staticmethod`.

tseaver (approved):

Issue #795, opened 2021-07-08
Refresh token is expired


tseaver:
See: https://github.com/googleapis/google-auth-library-python/pull/793#issuecomment-875981534

Also the source of the failures for #794.

PR #805, opened 2021-07-19
fix: fallback to source creds expiration in downscoped tokens

bojeil-google:
For downscoping CAB flow, the STS endpoint may not return the expiration
field for certain source credentials. The generated downscoped token
should always have the same expiration time as the source credentials.
When no `expires_in` field is returned in the response, we can just get
the expiration time from the source credentials.


arithmetic1728 (approved):

busunkim96:
Failures are due to an expired user credential (a return of #795).

PR #807, opened 2021-07-20
fix: do not use the GAE APIs on gen2+ runtimes

zevdg:
Currently, this library uses the App Engine credentials in all environments if
the APIs can be imported successfully. This assumption made sense when
the API was only available on gen1, but this is no longer the case.
See https://github.com/GoogleCloudPlatform/appengine-python-standard

In order to comply with AIP-4115, we must treat GAE gen2+ as a "compute
engine equivalent environment" even if the GAE APIs are importable.
In other words, google.auth.default() should not return an
app_engine.Credental on GAE gen2+.

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -230,6 +230,11 @@ def _get_explicit_environ_credentials():
 
 def _get_gae_credentials():
     """Gets Google App Engine App Identity credentials and project ID."""
+    # If not GAE gen1, prefer the metadata service even if the GAE APIs are
+    # available as per https://google.aip.dev/auth/4115.
+    if os.environ.get(environment_vars.LEGACY_APPENGINE_RUNTIME) != "python27":
+        return None, None
+
     # While this library is normally bundled with app_engine, there are
     # some cases where it's not available, so we tolerate ImportError.
     try:
--- a/tests/test__default.py
+++ b/tests/test__default.py
@@ -465,6 +465,7 @@ def test__get_gae_credentials_gen2():
     assert credentials is None
     assert project_id is None
 
+
 @mock.patch.dict(os.environ)
 def test__get_gae_credentials_gen2_backwards_compat():
     # compat helpers may copy GAE_RUNTIME to APPENGINE_RUNTIME
@@ -474,6 +475,7 @@ def test__get_gae_credentials_gen2_backwards_compat():
     credentials, project_id = _default._get_gae_credentials()
     assert credentials is None
     assert project_id is None
+
 
 def test__get_gae_credentials_env_unset():
     assert environment_vars.LEGACY_APPENGINE_RUNTIME not in os.environ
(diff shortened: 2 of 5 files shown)


zevdg:
It's worth noting that this change could be disruptive if a user migrates their app from GAE gen1 to GAE gen2 with a version of google.auth that doesn't include this change.  Then, instead of the ADC changing when they upgrade runtimes as the spec dictates, their ADC would change when they upgrade this library to include this change. As such, any migration guides should encourage users to upgrade this library to the latest py2-compatible version BEFORE migrating to gen2.

Conversely, if we do not make this change, then gen2 users could run into the opposite problem. Their ADC could suddenly start returning an app_engine.Credential instead of a compute_engine.Credential if they (or one of their dependencies) adds appengine-python-standard to their requirements.txt.

The sooner we land this change, the less people will be affected by either of those scenarios.

google/auth/_default.py
```diff
@@ -230,6 +230,11 @@ def _get_explicit_environ_credentials():
 
 def _get_gae_credentials():
     """Gets Google App Engine App Identity credentials and project ID."""
+    # If not GAE gen1, prefer the metadata service even if the GAE APIs are
+    # available as per AIP-4115.
```
busunkim96, line 234:
```suggestion
    # available as per https://google.aip.dev/auth/4115.
```

busunkim96:
CC @wescpy and @engelke since they're listed as DevRel contacts for this project.

Asking folks to update to a minimum version of `google-auth` is captured in a comment in a documentation bug (186268245).

busunkim96 (approved):

silvolu (approved):

arithmetic1728:
@zevdg Kokoro build failed because coverage test failed. It seems we need some additional unit tests to cover google/auth/_default.py line 243-245 and 251-255.

crwilcox (approved):

tests/test__default.py
```diff
@@ -447,7 +447,9 @@ def app_identity(monkeypatch):
     yield app_identity_module
 
 
-def test__get_gae_credentials(app_identity):
+@mock.patch.dict(os.environ)
+def test__get_gae_credentials_gen1(app_identity):
+    os.environ[environment_vars.LEGACY_APPENGINE_RUNTIME] = "python27"
```
tseaver, line 452: Tests below (`test__get_gae_credentials_no_app_engine`, `test__get_gae_credentials_no_apis`) also need to set this envvar:  otherwise, the code they are supposed to cover isn't reached, due to the early exit.

zevdg: Done.  Testing both with and without the env var in those tests for completeness.

FYI in the environments where that env var is normally set, the GAE APIs would always be available.  Someone would have to be explicitly setting APPENGINE_RUNTIME="python27" outside of a gae/devappserver environment to reach that codepath.  That said, they could be doing that, so test added.

PR #808, opened 2021-07-20
revert: "feat: service account is able to use a private token endpoint (#784)" 

arithmetic1728:
revert "feat: service account is able to use a private token endpoint (#784)" until b/194191737 is fixed.

This reverts commit 0e264092e35ac02ad68d5d91424ecba5397daa41.


liuchaoren (approved):

busunkim96 (approved):
Small title nit, `revert:` works as a conventional commit type, could you use that instead?

PR #812, opened 2021-07-21
feat: support refresh callable on google.oauth2.credentials.Credentials

bojeil-google:
This is an optional parameter that can be set via the constructor.
It is used to provide the credentials with new tokens and their
expiration time on `refresh()` call.

```
def refresh_handler(request, scopes):
    # Generate a new token for the requested scopes by calling
    # an external process.
    return (
        "ACCESS_TOKEN",
        _helpers.utcnow() + datetime.timedelta(seconds=3600))

creds = google.oauth2.credentials.Credentials(
    scopes=scopes,
    refresh_handler=refresh_handler)
creds.refresh(request)
```

It is useful in the following cases:
- Useful in general when tokens are obtained by calling some
  external process on demand.
- Useful in particular for retrieving downscoped tokens from a
  token broker.

This should have no impact on existing behavior. Refresh tokens
will still have higher priority over refresh handlers.

A getter and setter is exposed to make it easy to set the callable
on unpickled credentials as the callable cannot be easily serialized.

```
unpickled = pickle.loads(pickle.dumps(oauth_creds))
unpickled.refresh_handler = refresh_handler
```

### DIFF
--- a/google/oauth2/credentials.py
+++ b/google/oauth2/credentials.py
@@ -74,6 +74,7 @@ class Credentials(credentials.ReadOnlyScoped, credentials.CredentialsWithQuotaPr
         quota_project_id=None,
         expiry=None,
         rapt_token=None,
+        refresh_handler=None,
     ):
         """
         Args:
@@ -103,6 +104,13 @@ class Credentials(credentials.ReadOnlyScoped, credentials.CredentialsWithQuotaPr
                 This project may be different from the project used to
                 create the credentials.
             rapt_token (Optional[str]): The reauth Proof Token.
+            refresh_handler (Optional[Callable[[google.auth.transport.Request, Sequence[str]], [str, datetime]]]):
+                A callable which takes in the HTTP request callable and the list of
+                OAuth scopes and when called returns an access token string for the
+                requested scopes and its expiry datetime. This is useful when no
+                refresh tokens are provided and tokens are obtained by calling
+                some external process on demand. It is particularly useful for
+                retrieving downscoped tokens from a token broker.
         """
         super(Credentials, self).__init__()
         self.token = token
@@ -116,13 +124,18 @@ class Credentials(credentials.ReadOnlyScoped, credentials.CredentialsWithQuotaPr
         self._client_secret = client_secret
         self._quota_project_id = quota_project_id
         self._rapt_token = rapt_token
+        self.refresh_handler = refresh_handler
 
     def __getstate__(self):
         """A __getstate__ method must exist for the __setstate__ to be called
         This is identical to the default implementation.
         See https://docs.python.org/3.7/library/pickle.html#object.__setstate__
         """
-        return self.__dict__
+        state_dict = self.__dict__.copy()
+        # Remove _refresh_handler function as functions can't be pickled.
+        # The refresh_handler setter should be used to repopulate this.
+        del state_dict["_refresh_handler"]
+        return state_dict
 
     def __setstate__(self, d):
         """Credentials pickled with older versions of the class do not have
@@ -138,6 +151,8 @@ class Credentials(credentials.ReadOnlyScoped, credentials.CredentialsWithQuotaPr
         self._client_secret = d.get("_client_secret")
         self._quota_project_id = d.get("_quota_project_id")
         self._rapt_token = d.get("_rapt_token")
+        # The refresh_handler setter should be used to repopulate this.
+        self._refresh_handler = None
 
     @property
     def refresh_token(self):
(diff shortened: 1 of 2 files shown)


tests/oauth2/test_credentials.py
```diff
@@ -66,6 +66,50 @@ def test_default_state(self):
...
+            token_uri=None,
+            client_id=None,
+            client_secret=None,
+            rapt_token=None,
+            scopes=scopes,
+            default_scopes=None,
+            refresh_handler=original_refresh_handler,
+        )
+
+        assert creds.refresh_handler == original_refresh_handler
```
tseaver, line 87: ISTM `is` is clearer thatn `==` here (functions compare equality by `id` anyway):
```suggestion
        assert creds.refresh_handler is original_refresh_handler
```
and the same below.

tests/oauth2/test_credentials.py
```diff
@@ -66,6 +66,50 @@ def test_default_state(self):
...
+            credentials.Credentials(
+                token=None,
+                refresh_token=None,
+                token_uri=None,
+                client_id=None,
+                client_secret=None,
+                rapt_token=None,
+                scopes=scopes,
+                default_scopes=None,
+                refresh_handler=[mock.Mock()],
```
tseaver, line 109: That value is very confusing to the eye, because `mock.Mock` instances are explicitly callable:

```python
>>> import mock
>>> callable(mock.Mock())
True
>>> callable(mock.Mock(return_value=True))
True
```

So, I would likely use `object()` here, i.e.:
```suggestion
                refresh_handler=object(),
```

google/oauth2/credentials.py
```diff
@@ -116,13 +124,18 @@ def __init__(
...
         """A __getstate__ method must exist for the __setstate__ to be called
         This is identical to the default implementation.
         See https://docs.python.org/3.7/library/pickle.html#object.__setstate__
         """
-        return self.__dict__
+        state_dict = self.__dict__.copy()
+        # Remove _refresh_handler function as functions can't be pickled.
+        # The refresh_handler setter should be used to repopulate this.
+        del state_dict["_refresh_handler"]
+        return state_dict
```
tseaver, line 138: I'm not sure why you state, "functions can't be pickled":

```python
>>> import pickle
>>> def foo(): pass
... 
>>> bar = {"foo": foo}
>>> dumped = pickle.dumps(bar)
>>> pickle.loads(dumped)
{'foo': <function foo at 0x7fc3a19c6598>}
>>> dumped
b'\x80\x03}q\x00X\x03\x00\x00\x00fooq\x01c__main__\nfoo\nq\x02s.'
```

Some kinds of on-the-fly constructed functions (`lambda`, `functools.partial` instances) cannot be pickled, because they need to be importable.  See [what can be pickled](https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled) in the Python docs.

tseaver (changes requested):

tests/oauth2/test_credentials.py
bojeil-google: Done

tests/oauth2/test_credentials.py
bojeil-google: Done

google/oauth2/credentials.py
bojeil-google: Thanks for the reference. 
I ran into various errors when pickling a mock function and a local function:
```
    def test_pickle_and_unpickle_with_refresh_handler(self):
        expected_expiry = _helpers.utcnow() + datetime.timedelta(seconds=2800)
        # refresh_handler = mock.Mock(return_value=("TOKEN", expected_expiry))
        def refresh_handler():
            return ("TOKEN", expected_expiry)
    
        creds = credentials.Credentials(
            token=None,
            refresh_token=None,
            token_uri=None,
            client_id=None,
            client_secret=None,
            rapt_token=None,
            refresh_handler=refresh_handler,
        )
>       unpickled = pickle.loads(pickle.dumps(creds))
E       AttributeError: Can't pickle local object 'TestCredentials.test_pickle_and_unpickle_with_refresh_handler.<locals>.refresh_handler'
```
As my comment is not accurate, I have updated it to be more accurate.

tseaver (commented):
Your changes for my comments LGTM.  I don't really know the flow being used, so I'll hold off on a final "approve" until those who do can chime in.

bojeil-google:
> Your changes for my comments LGTM. I don't really know the flow being used, so I'll hold off on a final "approve" until those who do can chime in.

Thanks @tseaver for the review. @arithmetic1728 has reviewed the design doc. He should be able to review too.

arithmetic1728 (approved):

silvolu (approved):

tseaver (approved):

PR #819, opened 2021-07-29
fix: add SAML challenge to reauth

arithmetic1728:
This PR:

(1) re-add SAML challenge to reauth. It throws exception to fail early, and the exception is used by gcloud to instruct the users to do `gcloud auth login`. Without it, SAML users have to enter password 5 times then fail.

(2) when creating user credential, read `rapt_token` field from the json file if exists.

(3) disable reauth flow for user credentials and add a new `enable_reauth_refresh` flag to user credentials constructor.


google/auth/exceptions.py
```diff
@@ -57,3 +57,12 @@ def __init__(self, message=None):
...
             "Reauthentication failed. {0}".format(message)
         )
+
+
+class ReauthSamlChallengeFailError(ReauthFailError):
+    """An exception for SAML reauth challenge failures."""
+
+    def __init__(self, message=None):
+        super(ReauthSamlChallengeFailError, self).__init__(
+            "Reauthentication failed. {0}".format(message)
```
liuchaoren, line 67: "Reauthentication failed" appears in the base class too, so we will see this message twice? I seems we don't have to define a __init__ for this class because we can use the __init__ of the base class.

google/oauth2/_reauth_async.py
```diff
@@ -301,6 +304,11 @@ async def refresh_grant(
             == reauth._REAUTH_NEEDED_ERROR_RAPT_REQUIRED
         )
     ):
+        if not enable_reauth_refresh:
+            raise exceptions.RefreshError(
+                "Reauthenticatio is needed. Please run `gcloud auth application-default login` to reauthentciate."
```
liuchaoren, line 309: Reauthentication

liuchaoren: same here

google/oauth2/challenges.py
```diff
@@ -27,6 +27,9 @@
 
 
 REAUTH_ORIGIN = "https://accounts.google.com"
+SAML_CHALLENGE_MESSAGE = (
+    "Please run `gcloud auth login` to complete reauthentication with SAML\n"
```
liuchaoren, line 31: period after saml?

google/oauth2/challenges.py
```diff
@@ -151,7 +154,30 @@ def obtain_challenge_input(self, metadata):
         return None
 
 
+class SamlChallenge(ReauthChallenge):
+    """Challenge that asks the users to browse to their ID Providers.
+
+    Currently SAML challenge is not supported. When obtaining the challenge input,
+    exception will be raised to instruct the users to run `gcloud auth login` for reauthentication.
```
liuchaoren, line 161: breakl long line?

google/oauth2/challenges.py
```diff
@@ -151,7 +154,30 @@ def obtain_challenge_input(self, metadata):
...
+
+    @property
+    def is_locally_eligible(self):
+        return True
+
+    def obtain_challenge_input(self, metadata):
+        # Magic Arch has not fully supported returning a proper dedirect URL
+        # for programmatic SAML users today. So we error our here and request
+        # users to use gcloud to complete a login.
+        sys.stderr.write(SAML_CHALLENGE_MESSAGE)
```
liuchaoren, line 176: raise the exception should be enough?

google/oauth2/credentials.py
```diff
@@ -111,6 +112,8 @@ def __init__(
                 refresh tokens are provided and tokens are obtained by calling
                 some external process on demand. It is particularly useful for
                 retrieving downscoped tokens from a token broker.
+            enable_reauth_refresh (Optional[bool]): Whether reauth refresh flow should
```
liuchaoren, line 115: break long line?

google/oauth2/reauth.py
```diff
@@ -326,6 +329,11 @@ def refresh_grant(
             or response_data.get("error_subtype") == _REAUTH_NEEDED_ERROR_RAPT_REQUIRED
         )
     ):
+        if not enable_reauth_refresh:
+            raise exceptions.RefreshError(
+                "Reauthenticatio is needed. Please run `gcloud auth application-default login` to reauthentciate."
```
liuchaoren, line 334: typo?

liuchaoren: same here

google/auth/exceptions.py
arithmetic1728: removed

google/oauth2/_reauth_async.py
arithmetic1728: done

google/oauth2/challenges.py
arithmetic1728: done

google/oauth2/challenges.py
arithmetic1728: done

google/oauth2/challenges.py
arithmetic1728: removed it

google/oauth2/credentials.py
arithmetic1728: done

google/oauth2/reauth.py
arithmetic1728: fixed

liuchaoren (approved):

google/oauth2/_reauth_async.py
```diff
@@ -299,6 +302,11 @@ async def refresh_grant(
             == reauth._REAUTH_NEEDED_ERROR_RAPT_REQUIRED
         )
     ):
+        if not enable_reauth_refresh:
+            raise exceptions.RefreshError(
+                "Reauthentication is needed. Please run `gcloud auth application-default login` to reauthenticate."
```
silvolu, line 307: The command should be "gcloud auth login --update-adc"

google/oauth2/challenges.py
```diff
@@ -25,6 +25,9 @@
 
 
 REAUTH_ORIGIN = "https://accounts.google.com"
+SAML_CHALLENGE_MESSAGE = (
+    "Please run `gcloud auth login` to complete reauthentication with SAML."
```
silvolu, line 29: "gcloud auth login --update-adc"

google/oauth2/reauth.py
```diff
@@ -292,6 +293,8 @@ def refresh_grant(
             token has a wild card scope (e.g.
             'https://www.googleapis.com/auth/any-api').
         rapt_token (Optional(str)): The rapt token for reauth.
+        enable_reauth_refresh (Optional[bool]): Whether reauth refresh flow
+            should be used. The default value is False.
```
silvolu, line 297: Maybe add something about the fact that if used, the caller should update ADC credentials with the new rapt token to avoid constant re-triggering.

google/oauth2/reauth.py
```diff
@@ -324,6 +327,11 @@ def refresh_grant(
             or response_data.get("error_subtype") == _REAUTH_NEEDED_ERROR_RAPT_REQUIRED
         )
     ):
+        if not enable_reauth_refresh:
+            raise exceptions.RefreshError(
+                "Reauthentication is needed. Please run `gcloud auth application-default login` to reauthenticate."
```
silvolu, line 332: "gcloud auth login --update-adc"

silvolu (changes requested):

google/oauth2/reauth.py
arithmetic1728: added a comment saying this method for gcloud to use only

google/oauth2/_reauth_async.py
arithmetic1728: updated

google/oauth2/challenges.py
arithmetic1728: updated

google/oauth2/reauth.py
arithmetic1728: updated

google/oauth2/reauth.py
```diff
@@ -275,8 +275,10 @@ def refresh_grant(
     client_secret,
     scopes=None,
     rapt_token=None,
+    enable_reauth_refresh=False,
 ):
-    """Implements the reauthentication flow.
+    """Implements the reauthentication flow. This method is intended for gcloud
```
liuchaoren, line 280: This method is used no matter if reauth is enabled, right? enable_reauth_refresh should be used only by gcloud.

liuchaoren (approved):

google/oauth2/challenges.py
```diff
@@ -25,6 +25,7 @@
 
 
 REAUTH_ORIGIN = "https://accounts.google.com"
+SAML_CHALLENGE_MESSAGE = "Please run `gcloud auth login --update-adc` to complete reauthentication with SAML."
```
liuchaoren, line 28: Because only gcloud users will trigger reauth, so this exception message will only be shown to gcloud users. I believe we don't need the "--update-adc" because it is to overwrite the ADC. We only need to ask user to run 'gcloud auth login' to refresh gcloud's credential.

liuchaoren (approved):

google/oauth2/reauth.py
arithmetic1728: moved comments to `enable_reauth_refresh`.

google/oauth2/challenges.py
arithmetic1728: Since user credentials is created from ADC json file. If we don't update the ADC json file reauth will be triggered again, so this becomes a loop to the user.

liuchaoren: This exception is raise when we go through the challenges. If the credential is created from ADC json, enable_reauth_refresh is False and we will not go through challenges. So only CLI will set enable_reauth_refresh to True and may see this exceptions.

arithmetic1728: removed "--update-adc".

silvolu (approved):

Issue #822, opened 2021-08-03
Add 'docfx' session, wire into Kokoro 'docs-presubmit' job.


tseaver:
While working on PR #799 (to fix #797), I discovered that there is no `docfx` session in `noxfile.py`.

I made a quick effort to copy one in, but ran into issues which seemed harder than needed to get the CI job passing, so I'm opening this issue to track.

- [ ] Copy in a `docfx` session from another repo.
- [ ] Normalize `docs/config.py` with other repo to get required changes (path setup, etc.)
- [ ] Get `docs` session building with the same version of Sphinx.
- [ ] Re-add `docfx` to the `.kokoro/docs/docs-presubmit.cfg` configuration.

dandhlee:
docfx sessions are currently only required for Google Cloud client library. While I think it's fine to leave this open for in the future when docfx expands to non-Cloud libraries as well, don't sweat on having this done anytime soon!

parthea:
Closing as obsolete

Issue #797, opened 2021-07-08
Kokoro 'docs-presubmit' jobs fail consistently


tseaver:
E.g., [this failure](https://source.cloud.google.com/results/invocations/480d5f2b-91f0-45d8-9b71-eb4b5be28fad/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fgoogle-auth-library-python%2Fdocs%2Fdocs-presubmit/log):

```
================================================================
2021-07-08T16:53:25Z: Running the tests in a Docker container.
================================================================
KOKORO_KEYSTORE_DIR=/secrets/keystore
KOKORO_GITHUB_PULL_REQUEST_NUMBER=794
KOKORO_GITHUB_PULL_REQUEST_URL=https://github.com/googleapis/google-auth-library-python/pull/794
KOKORO_JOB_NAME=cloud-devrel/client-libraries/python/googleapis/google-auth-library-python/docs/docs-presubmit
KOKORO_GIT_COMMIT=f2e2eae0c46e1e51ac5b1d3669854f9cb1ed757d
KOKORO_GITHUB_PULL_REQUEST_COMMIT=f2e2eae0c46e1e51ac5b1d3669854f9cb1ed757d
KOKORO_BUILD_NUMBER=311
KOKORO_BUILD_ID=480d5f2b-91f0-45d8-9b71-eb4b5be28fad
KOKORO_GFILE_DIR=/secrets/gfile
cat: /secrets/gfile/project-id.txt: No such file or directory
.kokoro/build.sh: line 37: gcloud: command not found
```

Seems like we need to update `.kokoro/docker/docs/Dockerfile` to get the SDK installed?

tseaver:
Unlike similar changes in other repos, PR #803 did **not** fix the `docs-presubmit` job.

tseaver:
@parthea I've re-opened #799, and found at least one change needed to make it run.

PR #831, opened 2021-08-13
chore: omit 'docfx' from docs-presubmit build

tseaver:
We don't have a 'docfx' session in our noxfile (see #822).

Toward #797.


tseaver:
Note that I don't actually expect this PR to fix the `docs-presubmit` build, which has been failing in the `.kokoro/bulid.sh` setup before ever getting to the point of running `nox`.

.kokoro/build.sh
```diff
@@ -31,7 +31,7 @@ env | grep KOKORO
 export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-account.json
 
 # Setup project id.
-export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.txt")
+export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
```
dandhlee, line 34: `cat: /tmpfs/src/gfile/project-id.json: No such file or directory` I think it was fine before when you used `project-id.txt`?

tseaver: @dandhlee That filename works for the main Kokoro build, but *not* for the `docs-presubmit` build (see [this failure](https://source.cloud.google.com/results/invocations/a3937c80-4f2f-48a0-a9cf-d6dfa090d34a/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fgoogle-auth-library-python%2Fdocs%2Fdocs-presubmit/log)).

Maybe there is something weird in the `google3` Kokoro config for this repository?  I honestly can't figure out why `docs-presubmit` is failing here.

owlbot.py
```diff
@@ -7,11 +7,41 @@
...
+# Install gcloud SDK
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | \
+     sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
+  && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
+     sudo apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - \
+  && sudo apt-get update \
+  && sudo apt-get install google-cloud-sdk
+
+CMD ["python3.8"]""",
+    )
```
dandhlee, line 45: Adding the last bit to change `@nox.session(python="3.7")` to `@nox.session(python="3.8")` for `docs(session)` on noxfile.py should get docs session to run successfully now!

tseaver: I fixed that in 0c44e6a by modifying `noxfile.py` directly, because `owlbot.py` skips over it.

.kokoro/docker/docs/Dockerfile
```diff
@@ -64,4 +64,12 @@ RUN wget -O /tmp/get-pip.py 'https://bootstrap.pypa.io/get-pip.py' \
   && python3.8 /tmp/get-pip.py \
   && rm /tmp/get-pip.py
 
+# Install gcloud SDK
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | \
+    tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
+  && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
+    apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - \
+  && apt-get update -y \
+  && apt-get install python2 google-cloud-sdk -y
```
dandhlee, line 73: do we need `python2` if we don't need to run system tests?

tseaver: This PR isn't trying to fix #832 (at least not yet).  To do that, we need either to have a `.kokoro/build-systests.sh` script, which runs only the systests, or else work out some environment variable which we use to turn them on / off in `.kokoro/build.sh`.

Either way will require adding something like `.kokoro/presubmit/system-3.8.cfg`, to either set the environment variable or trigger the alternate script, which will then need a corresponding tweak to the `google3` Kokoro configuration for this repostitory.

dandhlee: Currently it just skips the system tests (missing 3.7 interpreter), and `docs` session runs fine. I think it's fine to move forward as-is.

dandhlee:
Think we're almost there. Blocked by system tests but it'll be fixed soon, docs-presubmit is happy without `docfx` and runs `docs` session fine. Not sure why OwlBot isn't happy though.

tseaver:
@dandhlee Owlbot gets its nose in a twist when the PR is rebased to remove its earlier, unwanted commits.

dandhlee:
Hm, I thought rerunning  it would resolve itself but I guess not.

tseaver:
@dandhlee I think we need to get a refresh-token PR merged to `master` and then here, first.

tseaver:
Blocked on #834.

tseaver:
@dandhlee PTAL.

tseaver:
Added `do not merge` label pending the `google3` change to activate the `Kokoro system-3.7` build.

.kokoro/docker/docs/Dockerfile
tseaver: PR #833 is based on this PR:  once we merge here, I will rebase it against `master`, and then wait to merge until the `google3` changes are in place to enable the `Kokoro system-3.7` build.

tseaver:
Oops, the Owlbot failure is real:  the substitution I added breaks when I run synthool locally.

dandhlee (approved):
LGTM.

PR #836, opened 2021-08-16
chore: separate systest into its own Kokoro entry

dandhlee:
To make the tests more easier for non-Googler folks to read, making system tests into its own entry. Will add internal stuff shortly to be in line with this PR.

Adding a directory of system tests for now in case other versions are to be added, if not we can just leave things as is.


tseaver (commented):
Please see https://github.com/googleapis/python-firestore/pull/412, which does not require a separate subdirectory.  @tswast Can you please confer about the appropriate `google3` change to make the `system-3.7` job work as is?

dandhlee:
Got the new updates, will close this PR after confirming the change.

dandhlee:
`Kokoro system-3.7` is now public and visible, closing this PR.

PR #856, opened 2021-09-03
fix: disable warning if quota project id provided to auth.default()

### LINKED ISSUE #769: google.auth.default(quota_project_id="..") still results in end user credentials warning
`quota_project_id` is not currently passed through to `_get_gcloud_sdk_credentials`. This bit of code still results in the end user credentials warning:

```py
import google.auth

credentials, project_id = google.auth.default(quota_project_id="my-project")
```

https://github.com/googleapis/google-auth-library-python/blob/09e0389db72cc9d6c5dde34864cb54d717dc0b92/google/auth/_default.py#L177-L195

https://github.com/googleapis/google-auth-library-python/blob/09e0389db72cc9d6c5dde34864cb54d717dc0b92/google/auth/_default.py#L73-L75

arithmetic1728:
If quota_project_id is provided via `google.auth.default(quota_project_id="my_id")`, then we pass the quota_project_id all the way down to `load_credentials_from_file`. In this case the created user credentials will have the `quota_project_id` set even if it is not available from the json file, so the users won't see the warning.

Fixes: #769


tests/test__default.py
```diff
@@ -332,11 +332,13 @@ def test__get_explicit_environ_credentials_no_env():
...
 
-    credentials, project_id = _default._get_explicit_environ_credentials()
+    credentials, project_id = _default._get_explicit_environ_credentials(
+        quota_project_id="project-foo"
+    )
 
     assert credentials is MOCK_CREDENTIALS
     assert project_id is mock.sentinel.project_id
-    load.assert_called_with("filename")
+    load.assert_called_with("filename", quota_project_id="project-foo")
```
tseaver, line 341: We should probably clone, or perpahs parmetrize, these tests, so that the implicit "branch" (whether the caller passed `quota_project_id`) is "covered."

tseaver (changes requested):

tests/test__default.py
arithmetic1728: yes, just parametrized the tests.

tseaver: Passing an explicit `None` isn't what I meant -- we need to test that the Right Thing(TM) happens when the user passes no `quota_project_id` at all.

arithmetic1728:
@tseaver added more tests to directly test the expected behavior

tseaver (approved):

PR #862, opened 2021-09-07
fix: reduce the clock skew back to 10 seconds

arithmetic1728:
clock skew has to be less than 30 seconds, see https://github.com/googleapis/google-auth-library-python/pull/581

if it is longer than 30 seconds, it will cause endless token refresh requests and a CPU spike in client libraries.


arithmetic1728:
did the fix in #863 instead

PR #863, opened 2021-09-08
fix: rename CLOCK_SKEW and separate client/server user case

arithmetic1728:
1. Rename `_helper.CLOCK_SKEW` to `_helper.REFRESH_THRESHOLD`. This value is for client to decide when to refresh a token.
2. Add a `clock_skew_in_seconds` keyword parameter (default value is 0) to `jwt.decode`, and remove the hard coded `_helper.CLOCK_SKEW` in the logic. The reason is `jwt.decode` is for server usage to validate iat/exp value. We should leave the clock skew to the server since this is server specific.

### DIFF
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -24,7 +24,7 @@ import urllib
 # token is expiring within 30 seconds, so refresh threshold should not be
 # more than 30 seconds. Otherwise auth lib will send tons of refresh requests
 # until 30 seconds before the expiration, and cause a spike of CPU usage.
-REFRESH_THRESHOLD = datetime.timedelta(seconds=10)
+REFRESH_THRESHOLD = datetime.timedelta(seconds=20)
 
 
 def copy_docstring(source_class):
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -20,8 +20,11 @@ import datetime
 import urllib
 
 
-CLOCK_SKEW_SECS = 60  # 60 seconds
-CLOCK_SKEW = datetime.timedelta(seconds=CLOCK_SKEW_SECS)
+# Token server doesn't provide a new a token when doing refresh unless the
+# token is expiring within 30 seconds, so refresh threshold should not be
+# more than 30 seconds. Otherwise auth lib will send tons of refresh requests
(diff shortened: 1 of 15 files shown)


busunkim96 (approved):

google/auth/_helpers.py
```diff
@@ -20,8 +20,11 @@
 import urllib
 
 
-CLOCK_SKEW_SECS = 60  # 60 seconds
-CLOCK_SKEW = datetime.timedelta(seconds=CLOCK_SKEW_SECS)
+# Token server doesn't provide a new a token when doing refresh unless the
+# token is expiring within 30 seconds, so refresh threshold should not be
+# more than 30 seconds. Otherwise auth lib will send tons of refresh requests
+# until 30 seconds before the expiration, and cause a spike of CPU usage.
+REFRESH_THRESHOLD = datetime.timedelta(seconds=10)
```
busunkim96, line 27: I see the last PR bumped it up to 60 seconds (https://github.com/googleapis/google-auth-library-python/pull/858). Is going back to 10 seconds intended?

The public docs on this https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances?hl=en#applications state "5 minutes".  

> Access tokens expire after a short period of time. The metadata server caches access tokens until they have 5 minutes of remaining time before they expire. You can request new tokens as frequently as you like, but your applications must have a valid access token for their API calls to succeed.

Discussion in b/177490340 from January of this year also seems to suggest it's still 5 minutes, for the metadata server. Is there a shorter window for other servers?

arithmetic1728: Yes going back to 10 seconds is intended. Cloud run metadata server has 30 seconds window from this issue b/161545142. Auth lib used to use 5 minutes, I changed to 10 seconds to solve that issue.

busunkim96: Could we go increase this value to closer to 30 seconds? The 10 second clock skew may be too tight in cases where the customer's clock is slightly out of sync. See last two comments in https://github.com/googleapis/python-api-core/issues/223#issuecomment-871717998.

arithmetic1728: updated to 20 seconds

silvolu (approved):

Issue #590, opened 2020-08-14
Generate id_token from default credentials


dinvlad:
**Is your feature request related to a problem? Please describe.**

Currently, `google.auth.default()` generates `Credentials` that contain `.token` that is an access token.
We'd like to also generate `id_token` with the appropriate audience starting from default credentials.

This would allow us to submit requests that use `id_token` uniformly among different environments:
- Local Service Account key
- Cloud Functions
- Cloud Run
- App Engine
- GKE Workload Identity

 **Describe the solution you'd like**

It would be great to add a method to `google.auth.credentials.Credentials` that allows `id_token` generation, e.g.
```py
credentials.id_token(audience='https://example.org')
```

 **Describe alternatives you've considered**
So far, we had to rely on piece-meal approaches, like this example from https://github.com/apache/airflow/blob/master/airflow/providers/google/common/utils/id_token_credentials.py

This unnecessarily increases the complexity of third-party apps, and we have to re-implement the same logic in each one of them.

 **Additional context**

AndreaGiardini:
@arithmetic1728 I stumbled upon this issue recently. Any plan to add support for user credentials? Or pointer in the right direction?

romanwozniak:
This is already 2 years old, but it is still not clear why it's impossible to generate `id_token` for user credentials. The workaround (i.e. [id_token_credentials.py](https://github.cm/apache/airflow/blob/master/airflow/providers/google/common/utils/id_token_credentials.py)) allows to get the id_token, but not for the right `audience`. I understand that this is not the limitation of the `google-auth` library itself, but rather something, that is not supported by the Google Auth API, but still, I don't understand what is the root cause for this.

In my use case, I would want to build an API server (exposed via Cloud Endpoints with OIDC authentication) and a CLI tool, that interacts with this server. Currently, I can't simply use `google.auth.default()` in the CLI, because this wouldn't give me the ID Token with expected audience. 

Can someone give me some leads to what am I doing wrong? Thank tou!

NicolaSpreafico:
No sure if still useful to someone, but after few searches I did not find an actual solution, but instead a clear indication in the documentation that the given method does NOT work for user credentials, and gcloud need to be used

https://cloud.google.com/run/docs/authenticating/service-to-service#use_the_authentication_libraries
> _This code does not work for obtaining authentication credentials for a user account._

https://cloud.google.com/docs/authentication#user-accounts
[...]

PR #866, opened 2021-09-14
fix: add fetch_id_token_credentials

### LINKED ISSUE #865: In fetch_id_token(), return token's expiry along with the token
Hello,

I have opened a ticket with support, and it was suggested that I open a feature request here.

I’ve been trying to implement [App Entity using OIDC tokens](https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/migrate-app-identity). It works, but I don't have a good way to handle token expiry.

The sending app calls fetch_id_token(). In order to avoid making that call for every outbound HTTP request, I cache the token. The question is: how long should it be cached?

I see that the library currently issues tokens that are valid for an hour, but that may not be true in future versions. So it would be great to access the expiry date along with the token.

Specifically, at https://github.com/googleapis/google-auth-library-python/blob/b79b55407b31933c9a8fe6de01478fa00a33fa2b/google/oauth2/id_token.py#L237 a Credentials object is available, but only `credentials.token` is returned.

Would it be a good idea to return the full object instead, or a dictionary like
```
{
  'token': '...',
  'expiry': ...
}
```
?

I have thought of two workarounds, but neither is satisfying.

The first one would be to call verify_oauth2_token() immediately after fetch_id_token(), and get the expiry date from there. This is not great because getting a new token now takes twice as long.

The second solution would be for the recipient of the request to notify the sender that the token has expired. The sender would then bust its cache and get a new token. This is inefficient because it takes an HTTP round-trip to realize that the token has expired, so it shouldn't be the default way of handling things.

Cheers,

Kevin

arithmetic1728:
fix: googleapis/google-auth-library-python#865 

`fetch_id_token` creates a credentials, refresh it, then return the token. In this PR, we refactor the credentials creation part  in `fetch_id_token` method to a new `fetch_id_token_credentials` method.

Note that we **only support service account credentials and compute engine credentials, not user credentials**.

The usage is:

```
target_audience = "https://pubsub.googleapis.com"

credentials = google.oauth2.id_token.fetch_id_token_credentials(request, target_audience)

# Refresh the credential to obtain an ID token.
credentials.refresh(request)
id_token = credentials.token
id_token_expiry = credentials.expiry
```

### DIFF
--- a/google/oauth2/id_token.py
+++ b/google/oauth2/id_token.py
@@ -64,6 +64,7 @@ from six.moves import http_client
 from google.auth import environment_vars
 from google.auth import exceptions
 from google.auth import jwt
+import google.auth.transport.requests
 
 
 # The URL that provides public certificates for verifying ID tokens issued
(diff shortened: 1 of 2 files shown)


busunkim96 (commented):
LGTM. Could a test be added for the new function?

silvolu:
+1 for test

mik-laj:
https://github.com/googleapis/google-auth-library-python/issues/590 is it related?

google/oauth2/id_token.py
```diff
@@ -174,8 +174,8 @@ def verify_firebase_token(id_token, request, audience=None):
     )
 
 
-def fetch_id_token(request, audience):
-    """Fetch the ID Token from the current environment.
+def fetch_id_token_credentials(request, audience):
```
mik-laj, line 177: Should we make the request parameter optional to keep consistent with `google.auth.default()`?

google/oauth2/id_token.py
```diff
@@ -230,11 +237,9 @@ def fetch_id_token(request, audience):
 
                 info = json.load(f)
                 if info.get("type") == "service_account":
```
mik-laj, line 239: Should we handle other credential types like `authorized_user` to keep consistent with google.autt.default()? See: https://github.com/googleapis/google-auth-library-python/blob/738611bd2914f0fd5fa8b49b65f56ef321829c85/google/auth/_default.py#L36

google/oauth2/id_token.py
```diff
@@ -230,11 +237,9 @@ def fetch_id_token(request, audience):
...
-                    credentials = service_account.IDTokenCredentials.from_service_account_info(
+                    return service_account.IDTokenCredentials.from_service_account_info(
                         info, target_audience=audience
                     )
-                    credentials.refresh(request)
-                    return credentials.token
```
mik-laj, line 237: Should we add else statement to handle unsupported credential type?

google/oauth2/id_token.py
mik-laj: For reference, you can look at Apache Airflow project: https://github.com/apache/airflow/blob/778be7993d4a8d752680d04db4edeb793386fa0c/airflow/providers/google/common/utils/id_token_credentials.py#L83-L104

busunkim96 (approved):

PR #880, opened 2021-10-01
fix: Raise exception on error from iam_sign_endpoint

bgirschig:
Raise an exception when the sign_bytes function of an impersonated credential gets an error from the iam_sign_endpoint.

(currently, it fails with a KeyError exception)

### DIFF
--- a/google/auth/impersonated_credentials.py
+++ b/google/auth/impersonated_credentials.py
@@ -288,7 +288,12 @@ class Credentials(credentials.CredentialsWithQuotaProject, credentials.Signing):
             url=iam_sign_endpoint, headers=headers, json=body
         )
 
-        return base64.b64decode(response.json()["signedBlob"])
+        response_data = response.json()
+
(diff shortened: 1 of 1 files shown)


google/auth/impersonated_credentials.py
```diff
@@ -288,7 +288,12 @@ def sign_bytes(self, message):
             url=iam_sign_endpoint, headers=headers, json=body
         )
 
-        return base64.b64decode(response.json()["signedBlob"])
+        response_data = response.json()
+
+        if "error" in response:
+          raise Exception(response_data["error"])
```
tseaver, line 294: The exception type should be based on the status code from the response (which we ought to be checking before extracting JSON, anyway).

We will also need one or more new tests for the error case.

tseaver (changes requested):

arithmetic1728:
fixed in #905

PR #892, opened 2021-10-18
fix: add back python 2.7 for gcloud usage only

arithmetic1728:
Add back python 2.7 since gcloud still needs it for a while. The support is temporary and it is intended for gcloud only.


setup.py
```diff
@@ -54,10 +63,12 @@
     namespace_packages=("google",),
     install_requires=DEPENDENCIES,
     extras_require=extras,
-    python_requires=">= 3.6",
+    python_requires=">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*",
     license="Apache 2.0",
     keywords="google auth oauth client",
     classifiers=[
+        "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.7",
```
busunkim96, line 71: I believe we wanted to skip the classifiers since the 2.7 compatibility is just for gcloud.

busunkim96 (approved):

setup.py
arithmetic1728: done

PR #913, opened 2021-11-08
chore: drop 'setuptools' dependency

### LINKED ISSUE #595: Setuptools as dependency is problematic w/ pip-tools

tseaver:
Closes #595.


arithmetic1728:
I need the `setuptools` dependency for my mtls work (to add some C/C++ extension). Is there other way to solve the issue without dropping the dependency?

tseaver:
@arithmetic1728

> I need the setuptools dependency for my mtls work (to add some C/C++ extension). Is there other way to solve the issue without dropping the dependency?

`install_requires` (which is what this PR changes) is about runtime dependencies, not build-time.  Given that you are going to script building the C extension and releasing binary wheels using e.g. Github Workflow actions (as in [python-crc32c](https://github.com/googleapis/python-crc32c/blob/main/.github/workflows/python-publish.yml) or some other mechanism, you can guarantee that `setuptools` is present during the build without having to declare it in `setup.py`.

*Users* of `google-auth` shouldn't need to have `setuptools` present in order to install the wheels you build in CI.

arithmetic1728:
@tseaver yep, you are right, thank you for the information!

arithmetic1728 (approved):

busunkim96 (approved):

PR #916, opened 2021-11-10
fix: fix the message format for metadata server exception

haoxins:
```
RefreshError: ("Failed to ... from the Google Compute Enginemetadata service. Status: 403 ...)
```


busunkim96 (approved):

busunkim96:
Thank you @haoxins!

Issue #762, opened 2021-05-27
ADC can load a json of impersonated service account credentials generated by gcloud


liuchaoren:
google-auth-library-java supported ADC impersonation and it is proposed in aip-dev/google.aip.dev#728.

liuchaoren:
@MadJlzz sorry for the late response! Yes, please see https://github.com/googleapis/google-auth-library-python/blob/main/CONTRIBUTING.rst for how to contribute to the repo.

DBCerigo:
Is not supporting `impersonated_service_account` a regression? I'm quite sure that we were using service account impersonation with python clients back in August 2020 (ref: https://stackoverflow.com/questions/69412702/how-to-manage-google-cloud-credentials-for-local-development/69417606)

MadJlzz:
@DBCerigo What I am sure is that you can use it directly from the code (like in my SO thread) but not from the default ADC

MadJlzz:
I made a PR with a first implementation that works where we can discuss around.

sloev:
thanks for doing some work on this @MadJlzz 
coming from AWS, (which has STS for temporary credentials), i must say google is quite inferior on this point.
Its a necessity to be able to run my code locally as a service account in order to develop the strict and narrow permissions my program needs once deployed.

the `gcloud auth application-default login --impersonate-service-account=[SA EMAIL]` is painful (too much friction when context switching between local projects) but it is utterly useless when pythons cloud sdk can't recognize the credentials produced.

crossing my fingers for this to be fixed soon :-)

MadJlzz:
Hey @sloev, thanks for your message. I was out from some time and was stuck with the environment I was working (Windows)
Wasn't able to run all unit tests because spme are using `os` and not `pathlib` if I remind correctly. I will try to move forward on this issue to have a fix asap.

liuchaoren:
Hi @MadJlzz we noticed some some issues which are blocked by this one. Do you think it is possible to fix this issue in one week. I am happy to take it over if it is not feasible. Thanks!

MadJlzz:
Hi @liuchaoren. I don't think I have experience enough with the way your testing this project finish this in one week. Also the PR is waiting your review for the next milestone I have to achieve (which should be covering the new section of the code)
Happy to let you finish if you guys need to be quick on this one.

liuchaoren:
@MadJlzz Thank you for your contribution! I am happy to pick it up from what you have. I will send a PR on the coming Monday.

[...]

PR #918, opened 2021-11-13
feat: ADC can load a json of impersonated service account credentials generated by gcloud

MadJlzz:
Proposition for #762 

It's in a "work in progress" state but I wanted to know if I was going in the right direction. I tried it out with `google-cloud-storage = "1.42.3"` and a simple script like:

```python
from google.cloud.storage import client

conn = client.Client()
print(next(conn.list_buckets()))
```

and it works.

If the `service-account` do not have the permission `storage.buckets.list`, the client returns a `403 Forbidden` as expected.

@liuchaoren what do you think?

edit: I've just signed the cla

### DIFF
--- a/google/auth/impersonated_credentials.py
+++ b/google/auth/impersonated_credentials.py
@@ -27,8 +27,8 @@ service account.
 
 import base64
 import copy
-from datetime import datetime
 import json
+from datetime import datetime
 
 import six
 from six.moves import http_client
@@ -44,18 +44,18 @@ _DEFAULT_TOKEN_LIFETIME_SECS = 3600  # 1 hour in seconds
 _IAM_SCOPE = ["https://www.googleapis.com/auth/iam"]
 
 _IAM_ENDPOINT = (
-    "https://iamcredentials.googleapis.com/v1/projects/-"
-    + "/serviceAccounts/{}:generateAccessToken"
+        "https://iamcredentials.googleapis.com/v1/projects/-"
+        + "/serviceAccounts/{}:generateAccessToken"
 )
 
 _IAM_SIGN_ENDPOINT = (
-    "https://iamcredentials.googleapis.com/v1/projects/-"
-    + "/serviceAccounts/{}:signBlob"
+        "https://iamcredentials.googleapis.com/v1/projects/-"
+        + "/serviceAccounts/{}:signBlob"
 )
 
 _IAM_IDTOKEN_ENDPOINT = (
-    "https://iamcredentials.googleapis.com/v1/"
-    + "projects/-/serviceAccounts/{}:generateIdToken"
+        "https://iamcredentials.googleapis.com/v1/"
+        + "projects/-/serviceAccounts/{}:generateIdToken"
 )
 
 _REFRESH_ERROR = "Unable to acquire impersonated credentials"
@@ -66,7 +66,7 @@ _DEFAULT_TOKEN_URI = "https://oauth2.googleapis.com/token"
 
 
 def _make_iam_token_request(
-    request, principal, headers, body, iam_endpoint_override=None
+        request, principal, headers, body, iam_endpoint_override=None
 ):
     """Makes a request to the Google Cloud IAM service for an access token.
     Args:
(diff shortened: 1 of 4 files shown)


google/auth/impersonated_credentials.py
```diff
@@ -321,6 +321,47 @@ def with_quota_project(self, quota_project_id):
             iam_endpoint_override=self._iam_endpoint_override,
         )
 
+    @classmethod
+    def from_authorized_user_info(cls, source_credentials, info, scopes=None):
```
liuchaoren, line 325: Thank you @MadJlzz for working on it! 

The high level idea of this PR looks good. Here is my thoughts on it:

1. Can we rename this function to "from_impersonated_credentials_info"? authorized_user has a special meaning which is for user account.
2. Why not extend this function to handle source_credentials deserialization, so that you don't need to create the source credentials in the _default? 
3. When deserializing the source credential, can we firstly check its type? It currently assumes the source is user account credentials, but it is not necessarily true. We should at least support user account credentials and service account credentials.

google/auth/impersonated_credentials.py
```diff
@@ -321,6 +321,47 @@ def with_quota_project(self, quota_project_id):
...
+                "fields {}.".format(", ".join(missing))
+            )
+
+        iam_endpoint_parser = re.search(pattern=r"(?P<service_account>[a-zA-Z-0-9]+)@(?P<project_id>[a-zA-Z-0-9]+)",
+                                        string=info.get("service_account_impersonation_url"))
+
+        return cls(
+            source_credentials=source_credentials,
+            target_principal=iam_endpoint_parser.group("service_account"),
+            target_scopes=["https://www.googleapis.com/auth/cloud-platform"],
```
MadJlzz, line 360: When doing impersonation, I found from the doc that `https://www.googleapis.com/auth/cloud-platform` scope is required. 
I was wondering if I should pass the user scope from `scopes` as well here. I was a bit confused with the general concept of `default_scope` and `scope`.

arithmetic1728: `default_scopes` are set up client libraries and `scopes` are set by users. `scopes` are supposed to override `default_scopes`. The difference only makes sense for self signed JWT. If you decide to use include these scopes, you can do `scopes_to_include = scopes if scopes else default_scopes`

google/auth/impersonated_credentials.py
```diff
@@ -321,6 +321,47 @@ def with_quota_project(self, quota_project_id):
...
+            delegates=info.get("delegates"),
+            iam_endpoint_override=info.get("service_account_impersonation_url")
+        ), iam_endpoint_parser.group("project_id")
```
liuchaoren, line 363: do you need to return this? It does not align with the docstring.

google/auth/impersonated_credentials.py
```diff
@@ -321,6 +321,47 @@ def with_quota_project(self, quota_project_id):
...
+        missing = keys_needed.difference(six.iterkeys(info))
+
+        if missing:
+            raise ValueError(
+                "Authorized user info was not in the expected format, missing "
+                "fields {}.".format(", ".join(missing))
+            )
+
+        iam_endpoint_parser = re.search(pattern=r"(?P<service_account>[a-zA-Z-0-9]+)@(?P<project_id>[a-zA-Z-0-9]+)",
+                                        string=info.get("service_account_impersonation_url"))
```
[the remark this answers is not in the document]
liuchaoren: We have https://github.com/googleapis/google-auth-library-python/blob/e756f08dc78616040ab8fbd7db20903137ccf0c7/google/auth/external_account.py#L163 which is used to get the service account email. You can consider to reuse it somehow. I don't think you need the project_id.

google/auth/impersonated_credentials.py
MadJlzz: I returned the `project_id` to follow the same mindset of the other type of auth but I can try if it's actually required.

MadJlzz: I tried, and it worked without. I seemed to be slower though. I guess the `project_id` should be inferred in some way later:
https://github.com/googleapis/google-auth-library-python/blob/e756f08dc78616040ab8fbd7db20903137ccf0c7/google/auth/_default.py#L197-L198

Because I already parse `service_account_impersonation_url` from the ADC, I propose to return it as well.
(rough measure with `time.time()` gave 0.5s when returning the `project_id` vs 1.8s without)

MadJlzz: @liuchaoren any news on that?

MadJlzz:
@liuchaoren I reviewed what I did to follow the general style of the project. If the structure is good, my next step would be to work on the test coverage for this PR.

manmat:
Hi, what is the status of this? It would be great to have this feature in the python library. Is there anything where additional help would move this forward?

liuchaoren:
Hi @manmat, https://github.com/googleapis/google-auth-library-python/issues/762 was fixed. The following release should include this feature.

MadJlzz:
I am closing this PR then. Thanks @liuchaoren for your work!

PR #937, opened 2021-12-30
fix(deps): allow cachetools 5.0 for python 3.7+

### LINKED ISSUE #934: new cachetools version (5.0.0+) can't work with google-auth
`cachetools` has released a new version (5.0.0) which conflicts with google-auth requirements of it being <5, this prevents updates to the `cachetools` package and pose a potential security concern (as updates are no longer possible to it)

```
The conflict is caused by:
    The user requested cachetools==5.0.0
    google-auth 2.3.3 depends on cachetools<5.0 and >=2.0.0
```

issue seems in https://github.com/googleapis/google-auth-library-python/blob/3c3fbf40b07e090f2be7fac5b304dbf438b5cd6c/setup.py#L23 

#### Environment details

  - OS: alpine3.11
  - Python version: python:3.8.6
  - pip version: 20.3.3
  - `google-auth` version: 2.3.3 (latest at time of writing)

#### Steps to reproduce

  1. try pip install using latest `cachetools` with latest `google-auth`
  2. pip fails

busunkim96:
Fixes #934.

Changes to cachetools listed in https://github.com/tkem/cachetools/blob/master/CHANGELOG.rst#v500-2021-12-21


setup.py
```diff
@@ -20,7 +20,8 @@
 
 
 DEPENDENCIES = (
-    "cachetools>=2.0.0,<5.0",
+    'cachetools>=2.0.0,<5.0; python_version < "3.7"',
+    'cachetools>=2.0.0,<6.0; python_version >= "3.7"',
```
cdce8p, line 24:
```suggestion
    'cachetools>=2.0.0,<6.0',
```
That shouldn't be necessary. Pip can automatically resolve the compatible version. I.e. pip will choose `<5.0` for Python 3.6.

cdce8p (commented):
Any updates? Unfortunately this is already causing dependency conflicts. A fix would be much appreciated.

setup.py
busunkim96: Ah sorry, I lost track of this PR. I've condensed this down to one pin as suggested.

arithmetic1728 (approved):

Issue #646, opened 2020-11-30
Used python rsa library brings PKCS#1 decryption vulnerability, is google-auth affected?


attila123:
#### Environment details

  - OS: not relevant for the issue
  - Python version: not relevant for the issue
  - pip version: not relevant for the issue
  - `google-auth` version: 1.23.0, currently latest

#### Steps to reproduce

Security vulnerability scanners pick up an unfixed PKCS#1 decryption code issue with the rsa library 4.6 (see among the dependencies in setup.py in this project): https://github.com/sybrenstuvel/python-rsa/issues/165 => although it is closed, from the comments it seems that it is not fixed, and anyway new version for this library is not released for some time...

One of the proposed solutions would be: https://github.com/sybrenstuvel/python-rsa/issues/165#issuecomment-727929096 :  
"3. modify code you depend on so that it uses libraries that do provide side-channel free behaviour for RSA decryption"

Questions:
1. Could you please confirm that this google-auth library does **not** use the vulnerable PKCS#1 decrypt() method, so that this rsa library issue can be considered a false-positive for google-auth? I tried to check quickly, for me this seems to be the case.
2. Would it be needed / feasible for google-auth to use another crypto library which does not have this issue? Is there a crypto library which would have clear advantages over the python rsa library?

busunkim96:
Hi @attila123, thanks for the report.

> 1. Could you please confirm that this google-auth library does not use the vulnerable PKCS#1 decrypt() method, so that this rsa library issue can be considered a false-positive for google-auth? I tried to check quickly, for me this seems to be the case.

`python-rsa` is used only in https://github.com/googleapis/google-auth-library-python/blob/37141e4dffc2ba3f3f57c5914544fb8b9cf7d017/google/auth/crypt/_python_rsa.py. As you've stated, the `decode()` method is not called.

Full list of calls to `python-rsa`:
https://github.com/googleapis/google-auth-library-python/blob/37141e4dffc2ba3f3f57c5914544fb8b9cf7d017/google/auth/crypt/_python_rsa.py#L77
https://github.com/googleapis/google-auth-library-python/blob/37141e4dffc2ba3f3f57c5914544fb8b9cf7d017/google/auth/crypt/_python_rsa.py#L101
https://github.com/googleapis/google-auth-library-python/blob/37141e4dffc2ba3f3f57c5914544fb8b9cf7d017/google/auth/crypt/_python_rsa.py#L108
[...]

PR #942, opened 2022-01-13
feat: setup.py: remove rsa requirement

### LINKED ISSUE #941: require cryptography in packaging metadata (and remove rsa)
In https://github.com/googleapis/google-auth-library-python/pull/185 , `google-auth` now uses Python's [cryptography](https://pypi.org/project/cryptography/) library if it's available.

The `cryptography` library is faster and better-maintained than [python-rsa](https://pypi.org/project/rsa).

Recently the `rsa` library had a security bug, and it took a long time to ship an update https://github.com/sybrenstuvel/python-rsa/issues/165 . This impacted `google-auth` users, see https://github.com/googleapis/google-auth-library-python/issues/646

Would you please update the `google-auth` packaging metadata to stop installing `rsa` and start installing `cryptography` instead?

ktdreyer:
The `rsa` library is slower and not as well-maintained as the `cryptography` library. Now that we require the `cryptography` library, drop the hard requirement on the `rsa` library.

Fixes: #941

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -372,4 +372,4 @@ intersphinx_mapping = {
 # Autodoc config
 autoclass_content = "both"
 autodoc_member_order = "bysource"
-autodoc_mock_imports = ["grpc"]
+autodoc_mock_imports = ["grpc", "rsa"]
--- a/google/auth/crypt/_cryptography_rsa.py
+++ b/google/auth/crypt/_cryptography_rsa.py
@@ -14,7 +14,7 @@
 
 """RSA verifier and signer that use the ``cryptography`` library.
 
-This is a much faster implementation than the default (in
+This is a much faster implementation than the alternative (in
 ``google.auth.crypt._python_rsa``), which depends on the pure-Python
 ``rsa`` library.
 """
--- a/setup.py
+++ b/setup.py
@@ -22,9 +22,6 @@ from setuptools import setup
 DEPENDENCIES = (
     "cachetools>=2.0.0,<7.0",
     "pyasn1-modules>=0.2.1",
-    # rsa==4.5 is the last version to support 2.7
-    # https://github.com/sybrenstuvel/python-rsa/issues/152#issuecomment-643470233
-    "rsa>=3.1.4,<5",
 )
 
 # TODO(https://github.com/googleapis/google-auth-library-python/issues/1737): Unit test fails with


arithmetic1728:
I think simply removing the rsa dependency will break the library since it is still in use: https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/crypt/_python_rsa.py#L28

ktdreyer:
Please see https://github.com/googleapis/google-auth-library-python/issues/646 , the code does not use python-rsa if python-cryptography is installed.

ktdreyer:
This is still relevant because `pip install google-auth` still pulls in the old `rsa` library.

As I look at rebasing this today, I can remove the entry from `setup.py`, but it's now present in `.kokoro/requirements.txt` too. How do I make the change to that file?

ktdreyer:
I researched kokoro to understand how to remove `rsa` from `.kokoro/requirements.txt`.

`google-auth` requires `rsa`, so we need to remove `rsa` from `setup.py`, then push a new release:

```
google-auth
  └rsa
```

The complicating factor is that `.kokoro/requirements.in` lists two modules that are not in `setup.py`: (`gcp-docuploader` and `gcp-releasetool`). These entries cause `pip-compile` to fetch `google-auth=2.28.0` from PyPI. Because `2.28.0` still requires `rsa`, `pip-compile` still puts `rsa` into `requirements.txt`.

I think the solution here is to release a new `google-auth` version with this PR, then re-run `pip-compile`, and we should see the entry disappear from `.kokoro/requirements.txt`.

prodigysml:
Any chance we can get some eyes on this one? There's a dependabot alert we can't get rid of, that would be fixed with this PR.

miketheman:
This might need to be revived due to https://github.com/sybrenstuvel/python-rsa/issues/245

miketheman:
I _think_ the build failure can be resolved by adding the missing `rsa` package to
https://github.com/googleapis/google-auth-library-python/blob/56b4d3b3bfd6f935411130c8828cdeea90aa69f1/docs/conf.py#L375

Ref: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports

That way, when sphinx autodoc traverses:

`google.auth._jwt_async` -> `google.auth.jwt` -> `google.auth.crypt` -> `google.auth.crypt.rsa` ->  (sphinx follows fallback) `google.auth.crypt._python_rsa` -> `import rsa`

That final import to the missing `rsa` package will be ignored.

ktdreyer:
@parthea would you please merge this? It simply drops `rsa` from the list of mandatory requirements.

daniel-sanche:
Thanks for putting this together and bringing it to our attention. I'll have to look into this a bit closer to make sure I understand the consequences of removing the dependency. Presumably, we'll need to add something else to provide a default implementation

di:
There is a polyfill here that already prefers the `pyca/cryptography` implementation and only falls back to `rsa` if it is unavailable, so in effect there already is a "default" implementation. 

https://github.com/googleapis/google-auth-library-python/blob/56b4d3b3bfd6f935411130c8828cdeea90aa69f1/google/auth/crypt/rsa.py#L18-L30

daniel-sanche:
Yeah, I think we're going to have to make cryptography a default dependency instead of an optional. Closing this in favor of https://github.com/googleapis/google-auth-library-python/pull/1771, which is making that change. Thanks!

PR #958, opened 2022-01-25
fix: Add AWS session token to metadata requests

sai-sunder-s:
AWS instance metadata service v2 introduced a need for session token when making call to the meta data endpoints. This change add the code to fetch the session token and add to our requests to metadata urls.

### DIFF
--- a/google/auth/aws.py
+++ b/google/auth/aws.py
@@ -406,6 +406,7 @@ class Credentials(external_account.Credentials):
         self._cred_verification_url = credential_source.get(
             "regional_cred_verification_url"
         )
+        self._aws_token_url = credential_source.get("aws_token_url")
         self._region = None
         self._request_signer = None
         self._target_resource = audience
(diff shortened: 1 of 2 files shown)


google/auth/aws.py
```diff
@@ -406,6 +406,7 @@ def __init__(
         self._cred_verification_url = credential_source.get(
             "regional_cred_verification_url"
         )
+        self._aws_token_url = credential_source.get("aws_token_url")
```
sai-sunder-s, line 409: Existing config files won't contain aws_token_url. So maybe adding a default value as backup can help?
Or maybe we can skip getting the session token if the aws_token_url is not provided and incase IDMSv2 is used, we can error out

google/auth/aws.py
```diff
@@ -540,7 +562,12 @@ def _get_region(self, request, url):
 
         if not self._region_url:
             raise exceptions.RefreshError("Unable to determine AWS region")
-        response = request(url=self._region_url, method="GET")
+
+        headers = None
+        if session_token is not None:
+            headers = {"X-aws-ec2-metadata-token": session_token}
```
arithmetic1728, line 568: this can be done in one line:
```suggestion
        headers = {"X-aws-ec2-metadata-token": session_token} if session_token else None
```

google/auth/aws.py
```diff
@@ -665,7 +703,14 @@ def _get_metadata_role_name(self, request):
             raise exceptions.RefreshError(
                 "Unable to determine the AWS metadata server security credentials endpoint"
             )
-        response = request(url=self._security_credentials_url, method="GET")
+
+        headers = None
+        if session_token is not None:
+            headers = {"X-aws-ec2-metadata-token": session_token}
```
arithmetic1728, line 709:
```suggestion
        headers = {"X-aws-ec2-metadata-token": session_token} if session_token else None
```

arithmetic1728 (approved):

lsirac:
Hey Sai, can you please update the README too? I think it'll be important to document this

PR #982, opened 2022-02-18
fix: Rename aws imdsv2 url field and update token lifetime

sai-sunder-s:
Rename aws session token url field to "imdsv2_session_token_url" and lower the session token ttl to 5 minutes.

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -261,7 +261,8 @@ following requirements are needed:
   external subject tokens and exchange them for service account access tokens.
 - If you want to use IDMSv2, then below field needs to be added to credential_source
   section of credential configuration.
-  "aws_session_token_url": "http://169.254.169.254/latest/api/token"
+  "_imdsv2_session_token_url": "http://169.254.169.254/latest/api/token"
+  The gcloud create-cred-config command will be updated to support this soon.
 
 Follow the detailed instructions on how to
 `Configure Workload Identity Federation from AWS`_.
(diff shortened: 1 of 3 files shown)


docs/user-guide.rst
```diff
@@ -261,7 +261,8 @@ following requirements are needed:
   external subject tokens and exchange them for service account access tokens.
 - If you want to use IDMSv2, then below field needs to be added to credential_source
   section of credential configuration.
-  "aws_session_token_url": "http://169.254.169.254/latest/api/token"
+  "_imdsv2_session_token_url": "http://169.254.169.254/latest/api/token"
+  The gcloud create-cred-config command will be updated to support this soon.
```
arithmetic1728, line 265: Should it be "imdsv2_session_token_url", no underscore?

I think the `gcloud create-....` sentence can be removed since this doc is for `google-auth` lib.

google/auth/aws.py
```diff
@@ -460,32 +462,35 @@ def retrieve_subject_token(self, request):
             str: The retrieved subject token.
         """
         # Fetch the session token required to make meta data endpoint calls to aws
-        if request is not None and self._aws_session_token_url is not None:
-            headers = {"X-aws-ec2-metadata-token-ttl-seconds": "21600"}
+        if request is not None and self._imdsv2_session_token_url is not None:
+            headers = {"X-aws-ec2-metadata-token-ttl-seconds": "300"}
```
arithmetic1728, line 466: why reduce it from 21600 to 300?

docs/user-guide.rst
sai-sunder-s: Good catch. Yeah there should be no underscore.
Other language libraries actually have the full gcloud command in the documentation. Since python documentation doesn't have the full gcloud command, maybe we can skip this line

arithmetic1728 (approved):

lsirac (approved):

Issue #980, opened 2022-02-15
Update automatic retries and Implement Retryable interface for token endpoint requests


TimurSadykov:
Current implementation of automatic retries to token endpoint does not follow Auth service guidelines. In addition, clients do not have a good way to distinguish retryable errors and implement custom retry strategies. Sometimes this causes infinite retries like here: https://github.com/nwbirnie/gapic-oauth-invalid-hangs

More details in the doc: go/auth-correct-retry

Please prioritize the fix if infinite retry like for Java is repro here as well.

arithmetic1728:
assigning to Timur to prioritize the work

clundin25:
https://github.com/googleapis/google-auth-library-python/pull/1113

PR #984, opened 2022-02-19
fix: retry Get for 500 and 503 error from GCE metadata server

baeminbo:
Currently, [getting metadata](
https://github.com/googleapis/google-auth-library-python/blob/v2.6.0/google/auth/compute_engine/_metadata.py#L148-L188) retries only for a transport error, but doesn't retry for retryable status code.  

[GCE metadata doc](https://cloud.google.com/compute/docs/metadata/querying-metadata#statuscodes) suggests retrying for 503. In addition, GCE metadata server also returns 500 error for intermittent unavailability. 

If this happens in token refresh, an intermittent 500 or 503 error is propagated as `RefreshError`. `RefreshError` is not retryable in `python-api-core` library.  So, just one time of an intermittent retryable error with GCE metadata leads to GCP API call failure. 

To mitigate this, I asked retry of `RefreshError` at [1], but the team suggested adding retry at `auth` layer [2].

[1] https://github.com/googleapis/python-api-core/issues/312
[2] https://github.com/googleapis/python-api-core/pull/313#issuecomment-978006491


arithmetic1728:
link with issue #980

TimurSadykov:
> #980

The #980 is about token endpoint, whilw this change addresses retries to Metadata endpoint

TimurSadykov:
@arithmetic1728 i think we need first to address the #980 and add Retryable interface, then we can leverage that to address Metadata retries.

TimurSadykov:
@baeminbo Hi, could you, please, provide any stats on Metadata service errors that you are trying to mitigate?

chalmerlowe:
Thank you for this contribution. I engaged Jules (Gemini) to review the current codebase and confirmed that the retry logic for 500 and 503 errors in `google.auth.compute_engine._metadata.get` appears to be implemented. The method uses `transport.DEFAULT_RETRYABLE_STATUS_CODES`, which currently includes both `INTERNAL_SERVER_ERROR (500)` and `SERVICE_UNAVAILABLE (503)`. Since the desired functionality is already present and appears to be working, I am closing this pull request.

PR #995, opened 2022-03-17
feat: Pluggable auth support

renkelvin:
Adding pluggable auth support.


google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+        Returns:
+            google.auth.pluggable.Credentials: The constructed
+                credentials.
+        """
+        with io.open(filename, "r", encoding="utf-8") as json_file:
+            data = json.load(json_file)
+            return cls.from_info(data, **kwargs)
+
+    def _parse_subject_token(self, response):
+        if not response["success"]:
```
lsirac, line 305: Let's reduce nesting in this block and remove the elif statements since we're raising errors.

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+            # environment_id is only supported in AWS or dedicated future external
+            # account credentials.
+            if "environment_id" in credential_source:
+                raise ValueError(
+                    "Invalid Pluggable credential_source field 'environment_id'"
+                )
+
+        if not self._credential_source_executable_command:
+            raise ValueError("Missing command. Executable command must be provided.")
+        if not self._credential_source_executable_timeout_millis:
```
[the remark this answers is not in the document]
lsirac: Also validate that it is between 0-120s.

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+        os.environ[
+            "GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE"
+        ] = "0"  # Always set to 0 until interactive mode is implemented.
+        original_service_account_impersonation_url = os.getenv(
+            "GOOGLE_EXTERNAL_ACCOUNT_IMPERSONATED_EMAIL"
+        )
+        if self._service_account_impersonation_url is not None:
+            os.environ[
+                "GOOGLE_EXTERNAL_ACCOUNT_IMPERSONATED_EMAIL"
+            ] = self._service_account_impersonation_url
```
lsirac, line 202: You need to parse the URL for the email, and set the email here not URL.

lsirac: There is a service_account_email() function in the base class that returns it.

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+        # Check output file
+        if self._credential_source_executable_output_file is not None:
+            try:
+                with open(
+                    self._credential_source_executable_output_file
+                ) as output_file:
+                    response = json.load(output_file)
+                    subject_token = self._parse_subject_token(response)
+            except:
+                pass
```
lsirac, line 183: I think ignoring issues when trying to read from the file hides bugs and maybe we should just throw an error. WDYT?

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+            google.auth.pluggable.Credentials: The constructed
+                credentials.
+        """
+        with io.open(filename, "r", encoding="utf-8") as json_file:
+            data = json.load(json_file)
+            return cls.from_info(data, **kwargs)
+
+    def _parse_subject_token(self, response):
+        if not response["success"]:
+            raise exceptions.RefreshError(
```
lsirac, line 306: If success=false, the error will look like this:

{
  "version": 1,
  "success": false,
  "code": "401",
  "message": "Permission denied. Caller not authorized"
}
Both 'code' and 'message' are required in this case. You need to build the error message using these fields.

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+                "Executable returned unsuccessful response: {}.".format(response)
+            )
+        elif response["version"] > EXECUTABLE_SUPPORTED_MAX_VERSION:
```
lsirac, line 309: This is the first thing we should check

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+            raise exceptions.RefreshError(
+                "Executable returned unsupported version {}.".format(
+                    response["version"]
+                )
+            )
+        elif response["expiration_time"] < time.time():
```
lsirac, line 315: Do we have any errors that have a code and a message?

For instance, if expired, we can return {INVALID_RESPONSE, "The executable response is expired.")

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+            )
+
+        # Check output file
+        if self._credential_source_executable_output_file is not None:
+            try:
+                with open(
+                    self._credential_source_executable_output_file
+                ) as output_file:
+                    response = json.load(output_file)
+                    subject_token = self._parse_subject_token(response)
```
lsirac, line 181: Let's add a comment that explains that if the cached response is expired, _parse_subject_token will raise an error which will be ignored and we will call the executable again.

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Pluggable Credentials.
+
+This module provides credentials to access Google Cloud resources from on-prem
```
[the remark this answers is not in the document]
lsirac: Also, you need to mention the executable specification, or point them to the README.

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+                    self._credential_source_executable_output_file
+                ) as output_file:
+                    response = json.load(output_file)
+                    subject_token = self._parse_subject_token(response)
+            except:
+                pass
+            else:
+                return subject_token
+
+        # Inject env vars
```
lsirac, line 187: Nit: A few comments are missing periods :-)

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+            ] = self._service_account_impersonation_url
+        original_credential_source_executable_output_file = os.getenv(
+            "GOOGLE_EXTERNAL_ACCOUNT_OUTPUT_FILE"
+        )
+        if self._credential_source_executable_output_file is not None:
+            os.environ[
+                "GOOGLE_EXTERNAL_ACCOUNT_OUTPUT_FILE"
+            ] = self._credential_source_executable_output_file
+
+        result = subprocess.run(
```
lsirac, line 211: Does this raise an error if the executable times out or is interrupted? We need to throw a custom error if it doesn't run within the specified timeout

lsirac (commented):
First pass. I didn't review the tests. Thanks Chuan!

google/auth/pluggable.py
renkelvin: The error is returned in result and we just raise it afterwards. What does the custom error look like?

google/auth/pluggable.py
renkelvin: I don't see any errors with error code though. They're pure messages.

google/auth/pluggable.py
renkelvin: The output file is more of a cache here. So if the file is broken, we just invalidate the cache and retrieve the token again. Throwing an error requires the developer to manually correct the output file, which might be hard.

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+            self._credential_source_executable_command.split(),
+            timeout=self._credential_source_executable_timeout_millis / 1000,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+        )
+
+        # Reset env vars
```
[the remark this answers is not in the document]
renkelvin: Yeah, it seems to be global in python.

google/auth/pluggable.py
```diff
@@ -0,0 +1,327 @@
...
+        subject_token_type,
+        token_url,
+        credential_source,
+        service_account_impersonation_url=None,
+        client_id=None,
+        client_secret=None,
+        quota_project_id=None,
+        scopes=None,
+        default_scopes=None,
+        workforce_pool_user_project=None,
```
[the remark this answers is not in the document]
renkelvin: Make sense. Added a check to make sure it's none.

google/auth/pluggable.py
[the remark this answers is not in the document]
renkelvin: Yeah, I duplicated the comments from external account credentials and added some comments for pluggable auth.
Make sense, added an example credential_source. Which README are you referring to?

renkelvin:
Hmm, I can't push to the pluggable branch. I may need to create a new pr from my fork.

renkelvin:
Hey @arithmetic1728, mind giving me the temporary write access to this repo? It should make the review much easier.

arithmetic1728:
@renkelvin you need to join https://github.com/googleapis first, then I can give you temporary write access

renkelvin:
> @renkelvin you need to join https://github.com/googleapis first, then I can give you temporary write access

Done joining https://github.com/googleapis

arithmetic1728:
@renkelvin I just gave you the write access, let me know if it doesn't work.

renkelvin:
> @renkelvin I just gave you the write access, let me know if it doesn't work.

It works, thanks!

google/auth/pluggable.py
lsirac: Ok. In Java, I defined a new PluggableAuthException. Maybe we can do something similar here?

google/auth/pluggable.py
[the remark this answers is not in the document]
lsirac: Looks like we document stuff here:

https://github.com/googleapis/google-auth-library-python/blob/main/docs/user-guide.rst

google/auth/pluggable.py
lsirac: If the file is corrupted, or acl'd badly, or something else, then we are hiding this from the developer and it is not behaving as intended. I think throwing an error here instead ensures that their configuration is properly set up and everything is WAI.

google/auth/pluggable.py
lsirac: I defined a new PluggableAuthException that has an error code and error description.

e.g.

```
...
throw new PluggableAuthException(
    "INTERRUPTED", String.format("The executable was interrupted: %s.", e));
```

google/auth/pluggable.py
```diff
@@ -0,0 +1,328 @@
...
+        self._credential_source_executable_output_file = self._credential_source_executable.get(
+            "output_file"
+        )
+
+        if not self._credential_source_executable_command:
+            raise ValueError("Missing command. Executable command must be provided.")
+        if not self._credential_source_executable_timeout_millis:
+            self._credential_source_executable_timeout_millis = 30 * 1000
+        elif (
+            self._credential_source_executable_timeout_millis < 0
```
lsirac, line 155: Need to do this check in milliseconds.

Also I said [0,120] but it is actually [5,120].

google/auth/pluggable.py
```diff
@@ -0,0 +1,328 @@
...
+            return cls.from_info(data, **kwargs)
+
+    def _parse_subject_token(self, response):
+        if response["version"] > EXECUTABLE_SUPPORTED_MAX_VERSION:
+            raise exceptions.RefreshError(
+                "Executable returned unsupported version {}.".format(
+                    response["version"]
+                )
+            )
+        if not response["success"] or not response["success"]:
```
lsirac, line 308: why is this repeated?

google/auth/pluggable.py
```diff
@@ -0,0 +1,328 @@
...
+            service_account_impersonation_url=service_account_impersonation_url,
+            client_id=client_id,
+            client_secret=client_secret,
+            quota_project_id=quota_project_id,
+            scopes=scopes,
+            default_scopes=default_scopes,
+            workforce_pool_user_project=workforce_pool_user_project,
+        )
+        if workforce_pool_user_project is not None:
+            raise ValueError("Pluggable auth doesn't support Workforce poolyet.")
```
lsirac, line 129: nit: missing space
 
and maybe change to

"Workforce Pools are not yet supported with Pluggable Auth."

google/auth/pluggable.py
lsirac: Mentioned in the spec in go/pluggable-auth-3pi

google/auth/pluggable.py
renkelvin: Make sense.

google/auth/pluggable.py
renkelvin: Looks better, let me try add something similar.

google/auth/pluggable.py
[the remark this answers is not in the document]
lsirac: You can actually pass the env to subprocess.run and it won't inherit from the current process:

```
env = os.environ.copy()
// set mappings on the copy
...
subprocess.run(env, ...)
```

Then you don't need to reset.

renkelvin:
@arithmetic1728 Any ideas how can I add `pytest_subprocess` as a dev dependency just for testing purpose? Thanks! https://source.cloud.google.com/results/invocations/e6b81a67-add0-4bfa-9089-fe92b4cc7007/targets/github%2Fgoogle-auth-library-python/tests

arithmetic1728:
@renkelvin I think you can add it here: https://github.com/googleapis/google-auth-library-python/blob/main/testing/requirements.txt

google/auth/pluggable.py
renkelvin: There are multiple cases that we should rerun the executable. For example: the file is not found, the token is expired, etc.. So we can't just throw an exception here.
How about we rerun the executable, but print in the console to inform the developer that the cached output file is not valid?

google/auth/pluggable.py
renkelvin: We throw 2 types of error here: ValueError(https://docs.python.org/3/tutorial/errors.html#raising-exceptions) and google.auth.exceptions(https://google-auth.readthedocs.io/en/stable/reference/google.auth.exceptions.html#google.auth.exceptions.RefreshError). Both of them take only a string to construct.

I tried to define an error class, however I think it makes the error construction more complicated and less clear. It also makes it hard for the system to read the string literal of the error class. WDYT?

google/auth/pluggable.py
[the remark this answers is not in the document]
renkelvin: Done. Good point.

google/auth/pluggable.py
```diff
@@ -0,0 +1,292 @@
...
+
+        Returns:
+            google.auth.pluggable.Credentials: The constructed
+                credentials.
+        """
+        with io.open(filename, "r", encoding="utf-8") as json_file:
+            data = json.load(json_file)
+            return cls.from_info(data, **kwargs)
+
+    def _parse_subject_token(self, response):
```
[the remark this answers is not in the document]
renkelvin: I'm confused here. What's the difference between "validate the response itself" and "check that the response is valid"? Can you redirect me to your java code?

tests/test_pluggable.py
```diff
@@ -0,0 +1,558 @@
...
+QUOTA_PROJECT_ID = "QUOTA_PROJECT_ID"
+SCOPES = ["scope1", "scope2"]
+SUBJECT_TOKEN_FIELD_NAME = "access_token"
+
+TOKEN_URL = "https://sts.googleapis.com/v1/token"
+SUBJECT_TOKEN_TYPE = "urn:ietf:params:oauth:token-type:jwt"
+AUDIENCE = "//iam.googleapis.com/projects/123456/locations/global/workloadIdentityPools/POOL_ID/providers/PROVIDER_ID"
+
+
+class TestCredentials(object):
```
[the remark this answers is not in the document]
renkelvin: 1. Added: unsuccessful response with error code and error message; unsupported token type;
Existing: unsupported response version; token expired;
For 2 & 3, due to the limitation of `pytest_subprocess` package, I didn't find a way to check the env variables of the subprocess, or set a running time for the subprocess. `FILE_NOT_FOUND` exception is covered, where the exception is ignored and we rerun the executable.

google/auth/pluggable.py
renkelvin: Done.

google/auth/pluggable.py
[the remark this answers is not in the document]
lsirac: There are two things: 
1) Validating the executable response format -> is there a "success" field? is there a "version" field? etc. if not, we can throw an error saying this field is missing
2) Validating the response itself -> is the "version" field set to 1?, etc.

google/auth/pluggable.py
lsirac: Let's keep it like this then if this is how it currently works in the lib.

google/auth/pluggable.py
lsirac: Seems like there is missing a change here? Still just ignoring any issues with the output file?

google/auth/pluggable.py
lsirac: If the executable returns a bad response, what does the error look like right now?

google/auth/_default.py
```diff
@@ -317,7 +317,7 @@ def _get_external_account_credentials(
         google.auth.exceptions.DefaultCredentialsError: if the info dictionary
             is in the wrong format or is missing required information.
     """
-    # There are currently 2 types of external_account credentials.
+    # There are currently 3 types of external_account credentials.
```
lsirac, line 320: Seems like you forgot to update ADC?

google/auth/pluggable.py
```diff
@@ -0,0 +1,306 @@
...
+        ] = "0"  # Always set to 0 until interactive mode is implemented.
+        if self._service_account_impersonation_url is not None:
+            env[
+                "GOOGLE_EXTERNAL_ACCOUNT_IMPERSONATED_EMAIL"
+            ] = self.service_account_email()
+        if self._credential_source_executable_output_file is not None:
+            env[
+                "GOOGLE_EXTERNAL_ACCOUNT_OUTPUT_FILE"
+            ] = self._credential_source_executable_output_file
+        result = subprocess.run(
```
lsirac, line 205: I think this will result in this process hanging if the STDOUT buffer is filled. Can you test (go/pluggable-auth-sdk-testing - last test case) and provide a workaround?

renkelvin: Wow, good case. Let me test it out.

google/auth/pluggable.py
renkelvin: See the lines below. The error will be something like:
"Executable exited with non-zero return code 1. Error: DETAILED ERROR MESSAGE FROM STDOUT"

google/auth/pluggable.py
renkelvin: If there is an issue with the file itself (not the content) like file not exists, the error is ignored and we generate the file then.
If the there is an issue with the content, the error is raised.

google/auth/pluggable.py
renkelvin: Confirmed that it won't hang but error out with invalid response.

renkelvin:
Hey @arithmetic1728, it seems the build for Python 3.7 passed but Kokro build failed.
Is the "Kokoro" build for 2.x versions? Received an error of "ModuleNotFoundError: No module named 'pytest_subprocess'". Note that `pytest_subprocess` is only available for Python 3.x. Any ideas? Thanks!

arithmetic1728:
@renkelvin It seems both python 3.6 and 2.7 unit tests failed with this error. `google-auth` supports python 2.7 and >= 3.6. https://github.com/googleapis/google-auth-library-python/blob/main/setup.py#L65

google/auth/pluggable.py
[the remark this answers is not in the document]
lsirac: Looks like you added it. Resolving.

lsirac (approved):
LGTM for implementation details. Thanks Chuan!

renkelvin:
@arithmetic1728 I thought 2.x is no longer supported, according to the README file https://github.com/googleapis/google-auth-library-python/blob/main/README.rst
Is it possible we support this feature somehow only for 3.6+? Since it requires the timeout support of `subprocess` module which is only available in python 3.4+.

arithmetic1728:
@renkelvin We don't need to support new features for 2.7, but the code has to be 2.7 compatible. gcloud CLI uses google-auth and it still supports 2.7, so if we use the 3.x only API in google-auth, gcloud will break.

I guess you can do the following:
```
import sys

def refresh(..):
    if sys.version_info < (3, 0):
        raise RefreshError("pluggable auth is only supported for python 3.6+")
```

If you want to support it for 2.7, then you can implement the timeout on your own.

google/auth/pluggable.py
```diff
@@ -0,0 +1,319 @@
...
+        Returns:
+            google.auth.pluggable.Credentials: The constructed
+                credentials.
+        """
+        with io.open(filename, "r", encoding="utf-8") as json_file:
+            data = json.load(json_file)
+            return cls.from_info(data, **kwargs)
+
+    def _parse_subject_token(self, response):
+        if not "version" in response:
```
sai-sunder-s, line 281: I am confused about the purpose of the version. This is not  a version for the credential_source format, but its a version for the output of the executable?

renkelvin:
> @renkelvin We don't need to support new features for 2.7, but the code has to be 2.7 compatible. gcloud CLI uses google-auth and it still supports 2.7, so if we use the 3.x only API in google-auth, gcloud will break.
> 
> I guess you can do the following:
> 
> ```
> import sys
> 
> def refresh(..):
>     if sys.version_info < (3, 0):
>         raise RefreshError("pluggable auth is only supported for python 3.6+")
> ```

@arithmetic1728 Thanks! After some discussions with my teammates, we decide to support both 2.7 and 3.6. The time timeout support will be added in a separate pr though.

google/auth/pluggable.py
renkelvin: Yes, the output of the executable is expected to comply with the spec(versioned) we defined.

google/auth/pluggable.py
```diff
@@ -0,0 +1,319 @@
...
+                except ValueError:
+                    raise
+                except exceptions.RefreshError:
+                    pass
+                else:
+                    return subject_token
+
+        # Inject env vars.
+        env = os.environ.copy()
+        env["GOOGLE_EXTERNAL_ACCOUNT_AUDIENCE"] = self._audience
```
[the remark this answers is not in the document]
renkelvin: They are used by the customer provided plugged executable.

google/auth/pluggable.py
```diff
@@ -0,0 +1,319 @@
...
+                credentials.
+        """
+        with io.open(filename, "r", encoding="utf-8") as json_file:
+            data = json.load(json_file)
+            return cls.from_info(data, **kwargs)
+
+    def _parse_subject_token(self, response):
+        if not "version" in response:
+            raise ValueError("The executable response is missing the version field.")
+        if response["version"] > EXECUTABLE_SUPPORTED_MAX_VERSION:
```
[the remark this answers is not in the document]
renkelvin: We might stop supporting old versions in the future, while EXECUTABLE_SUPPORTED_MIN_VERSION is not in scope yet and it can be added later I think. @lsirac any thoughts?

google/auth/pluggable.py
```diff
@@ -0,0 +1,319 @@
...
+        if result.returncode != 0:
+            raise exceptions.RefreshError(
+                "Executable exited with non-zero return code {}. Error: {}".format(
+                    result.returncode, result.stdout
+                )
+            )
+        else:
+            try:
+                data = result.stdout.decode("utf-8")
+                response = json.loads(data)
```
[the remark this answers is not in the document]
renkelvin: Yes, the file will be used as cache.

google/auth/pluggable.py
```diff
@@ -0,0 +1,319 @@
...
+            timeout=self._credential_source_executable_timeout_millis / 1000,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            env=env,
+        )
+
+        if result.returncode != 0:
+            raise exceptions.RefreshError(
+                "Executable exited with non-zero return code {}. Error: {}".format(
+                    result.returncode, result.stdout
```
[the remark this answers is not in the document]
renkelvin: See line 213, the output of stderr is redirected to stdout.

google/auth/pluggable.py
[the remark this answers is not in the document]
lsirac: I don't think it makes sense to have a min version. We generally don't make backwards incompatible changes. The use case here is if we update the response specification and try to use an old library. If needed it can be added in the future.

arithmetic1728:
You can run the following easily-to-fail kokoro tests locally before pushing to github. This will save a lot of time. (Don't worry about the kokoro system-3.7 failure - It is caused by expired creds. I will update the creds once your PR is ready to merge).

lint: `python -m nox -s lint`
If lint fails, run `python -m nox -s blacken` to format the files, then re-run lint. If it can pass, don't forget to add and commit the changes.

cover: `python -m nox -s cover`

renkelvin:
> You can run the following easily-to-fail kokoro tests locally before pushing to github. This will save a lot of time. (Don't worry about the kokoro system-3.7 failure - It is caused by expired creds. I will update the creds once your PR is ready to merge).
> 
> lint: `python -m nox -s lint` If lint fails, run `python -m nox -s blacken` to format the files, then re-run lint. If it can pass, don't forget to add and commit the changes.
> 
> cover: `python -m nox -s cover`

Thanks @arithmetic1728! The CI passes now.

arithmetic1728 (approved):

arithmetic1728:
LGTM 

@sai-sunder-s would you like to take another look?
@renkelvin I will fix the system test in a separate PR when we are ready to check in.

sai-sunder-s (approved):

PR #1022, opened 2022-04-14
feat: add experimental GDCH support

arithmetic1728:
internal doc: go/gdch-python-auth-lib

experimental support for GDCH (Google Distributed Cloud Hosted) credentials.

Besides the constructor, a GDCH credential can be created via application default credentials.

To do so, user first creates a JSON file of the following format::

        {
            "type":"gdch_service_account",
            "format_version":"v1",
            "k8s_ca_cert_path":"<k8s ca cert pem file path>",
            "k8s_cert_path":"<k8s cert pem file path>",
            "k8s_key_path":"<k8s key pem file path>",
            "k8s_token_endpoint":"<k8s token endpoint>",
            "ais_ca_cert_path":"<AIS ca cert pem file path>",
            "ais_token_endpoint":"<AIS token endpoint>"
        }

Here "k8s_*" files are used to request a k8s token from k8s token endpoint using mutual TLS connection. The k8s token is then sent to AIS token endpoint to exchange for an AIS token. The AIS token will be used to talk to Google API services.

"k8s_ca_cert_path" field is not needed if the k8s server uses well known CA. "ais_ca_cert_path" field is not needed if the AIS server uses well known CA. These two fields can be used for testing environments.

The "format_version" field stands for the format of the JSON file. For now it is always "v1".

After the JSON file is created, set `GOOGLE_APPLICATION_CREDENTIALS` environment variable to the JSON file path, then use the following code to create the credential::

        import google.auth

        credential, _ = google.auth.default()
        credential = credential.with_audience("<the audience>")

The audience denotes the scope the AIS token is requested, for example, it could be either a k8s cluster or API service.


tests/test__default.py
```diff
@@ -637,6 +640,18 @@ def test__get_gcloud_sdk_credentials_no_project_id(load, unused_isfile, get_proj
     assert get_project_id.called
 
 
+def test__get_gdch_service_account_credentials_no_format_version():
+    with pytest.raises(exceptions.DefaultCredentialsError) as excinfo:
+        _default._get_gdch_service_account_credentials({})
+    assert excinfo.match("format_version is not provided or unsupported")
```
sai-sunder-s, line 646: shouldn't the error be "format_version is not provided or unsupported. Supported version is: v1"? Same below

sai-sunder-s (approved):

sai-sunder-s (approved):

Issue #857, opened 2021-08-26
Json Error


nitishxp:
Lib: google-cloud-bigquery==2.23.1

Getting below errors occasionally when calling  client.delete_table(<>, not_found_ok=True)

```
File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/cloud/bigquery/client.py",
line 1788, in delete_table self._call_api( File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/cloud/bigquery/client.py",
line 747, in _call_api return call() File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/api_core/retry.py",
line 286, in retry_wrapped_func return retry_target( File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/api_core/retry.py",
line 189, in retry_target return target() File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/cloud/_http.py",
line 473, in api_request response = self._make_request( File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/cloud/_http.py",
line 337, in _make_request return self._do_request( File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/cloud/_http.py",
line 375, in _do_request return self.http.request( File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/auth/transport/requests.py",
line 478, in request self.credentials.before_request(auth_request, method, url, request_headers) File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/auth/credentials.py",
line 133, in before_request self.refresh(request) File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/oauth2/service_account.py",
line 376, in refresh access_token, expiry, _ = _client.jwt_grant( File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/oauth2/_client.py",
line 193, in jwt_grant response_data = _token_endpoint_request(request, token_uri, body) File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/oauth2/_client.py",
line 161, in _token_endpoint_request response_status_ok, response_data = _token_endpoint_request_no_throw( File "/layers/google.python.pip/pip/lib/python3.9/site-packages/google/oauth2/_client.py",
line 121, in _token_endpoint_request_no_throw response_data = json.loads(response_body) File "/opt/python3.9/lib/python3.9/json/__init__.py",
[...]

PR #1044, opened 2022-05-26
feat: add experimental GDCH support

arithmetic1728:
Add experimental GDCH service account credential support. Example usage:
 
1.first create a JSON file of the following format
```
{
    "type": "gdch_service_account",
    "format_version": "1",
    "project": "<project name>",
    "private_key_id": "<key id>",
    "private_key": "-----BEGIN EC PRIVATE KEY-----\n<key bytes>\n-----END EC PRIVATE KEY-----\n",
    "name": "<service identity name>",
    "ca_cert_path": "<CA cert path>",
    "token_uri": "https://service-identity.<Domain>/authenticate"
}
```
2.Clone this repo, then switch to `gdch_jwt` branch, and install the dependencies
```
git clone https://github.com/googleapis/google-auth-library-python.git
cd google-auth-library-python
git checkout gdch_jwt
python -m pip install -e .
python -m pip install requests
python -m pip install cryptography
```
3.Create a `sample.py` with the following code, and fill in the api_endpoint. Note that if api endpoint doesn't use well known CA, then you need to provide the CA cert.
For gRPC, you can `export GRPC_DEFAULT_SSL_ROOTS_FILE_PATH=<ca cert path>`.
For HTTP, you can pass `verify=<ca cert path>` to `authed_session.request` (see the sample)
```
import google.auth
from google.auth.transport import requests

api_endpoint = "<fill in the api endpoint>"

creds, project_id = google.auth.default()
creds = creds.with_gdch_audience(api_endpoint)

def test_get_token():
    req = requests.Request()
    creds.refresh(req)
    print(creds.token)

def test_call_api(ca_cert_path=None):
    authed_session = requests.AuthorizedSession(creds)
    response = authed_session.request("GET", api_endpoint, verify=ca_cert_path)
    print(response.status_code)
    print(response.text)

if __name__=="__main__":
    # If you want to test if a token can be obtained, run test_get_token
    test_get_token()

    # if you also want to test if we can call the api endpoint, run test_call_api.
    # provide the CA cert via ca_cert_path parameter if needed.
    # (note that test_get_token is not necessary, since when we call the api
    # endpoint, token will be automatically refreshed)
    test_call_api(ca_cert_path=None)
```
4.First `export GOOGLE_APPLICATION_CREDENTIALS=<the json file path>`, then run the sample with `python sample.py`.


google/oauth2/_client.py
```diff
@@ -112,32 +121,40 @@ def _token_endpoint_request_no_throw(
...
+                error_desc = response_data.get("error_description") or ""
+                error_code = response_data.get("error") or ""
+                if (
+                    any(e == "internal_failure" for e in (error_code, error_desc))
+                    and retry < 1
+                ):
+                    retry += 1
+                    continue
+            except ValueError:
+                response_data = response_body
```
sai-sunder-s, line 150: are we assuming that if the body is not json, it is a valid response string?
Is there a possibility of an error string?

google/oauth2/gdch_credentials.py
```diff
@@ -0,0 +1,242 @@
...
+    We can also create the credential directly::
+
+        from google.oauth import gdch_credentials
+
+        credential = gdch_credentials.ServiceAccountCredentials.from_service_account_file("<the json file path>")
+        credential = credential.with_gdch_audience("<the audience>")
+
+    The token is obtained in the following way. This class first creates a
+    self signed JWT. It uses the `name` value as the `iss` and `sub` claim, and
+    the `token_uri` as the `aud` claim, and signes the JWT with the `private_key`.
```
sai-sunder-s, line 78: nit: signs

google/auth/_default.py
```diff
@@ -421,6 +428,20 @@ def _get_impersonated_service_account_credentials(filename, info, scopes):
...
 
+def _get_gdch_service_account_credentials(filename, info):
+    from google.oauth2 import gdch_credentials
+
+    try:
+        credentials = gdch_credentials.ServiceAccountCredentials.from_service_account_info(
+            info
+        )
+    except ValueError as caught_exc:
+        msg = "Failed to load GDCH service account credentials from {}".format(filename)
```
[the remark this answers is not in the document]
arithmetic1728: The filename is just used for the error message creation at line 439. The file content is already loaded into info.

google/oauth2/_client.py
arithmetic1728: This fixes bug #857

google/oauth2/gdch_credentials.py
arithmetic1728, line 78:
```suggestion
    the `token_uri` as the `aud` claim, and signs the JWT with the `private_key`.
```

arithmetic1728: done

google/oauth2/gdch_credentials.py
```diff
@@ -97,23 +98,27 @@ def __init__(
...
         now = _helpers.utcnow()
         expiry = now + JWT_LIFETIME
+        iss_sub_value = "system:serviceaccount:{}:{}".format(
+            self._project, self._service_identity_name
+        )
 
         payload = {
-            "iss": self._service_identity_name,
-            "sub": self._service_identity_name,
+            "iss": iss_sub_value,
```
sai-sunder-s, line 114: why changing the issuer and subject?

sai-sunder-s (approved):

google/oauth2/gdch_credentials.py
arithmetic1728: GDCH token server requires this new format.

PR #1046, opened 2022-05-31
docs: fix changelog header to consistent size

dandhlee:
There was a minor issue with patch version releases labeled as an H3 header for changelog. Future changelog headers will always be H2 headers but existing entries must manually be fixed.


arithmetic1728 (approved):

arithmetic1728:
system test failure will be fixed in #1048

PR #1072, opened 2022-06-30
fix: Fix IDTokenCredentials update bug

mbmccoy:
The `from_credentials` method should use the specified `target_credentials`.


arithmetic1728:
@mbmccoy could you change the commit message to "fix: Fix IDTokenCredentials update bug" (with a "fix:" prefix) so it can pass the commit message format check? Also could you add a simple unit test for this change?

clundin25:
Thanks @mbmccoy. I have updated your commit to match our conventional commits style guide.

clundin25:
@sai-sunder-s can you take a look as well?

clundin25 (approved):

tests/test_impersonated_credentials.py
```diff
@@ -473,6 +473,7 @@ def test_id_token_from_credential(
 
         assert id_creds.token == ID_TOKEN_DATA
         assert id_creds._include_email is True
+        assert id_creds._target_credentials is credentials
```
sai-sunder-s, line 476: the same creds are used in ln 469 and 471. So there is no point of making the `from_credentials` call in 471 I think?

clundin25: Updated with a fix for that

PR #1073, opened 2022-06-30
feat: add integration tests for pluggable auth

ScruffyProdigy:


system_tests/system_tests_sync/test_external_accounts.py
```diff
@@ -303,3 +305,42 @@ def test_aws_based_external_account(
                 },
             },
         )
+
+def test_pluggable_external_account(
+    oidc_credentials, service_account_info, dns_access
```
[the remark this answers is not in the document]
ScruffyProdigy: Most of these come from functions in this file or the noxfile that have a @pytest.fixture decorator. Somehow it takes the name of the parameter and finds the matching fixture

sai-sunder-s (approved):

system_tests/system_tests_sync/test_external_accounts.py
```diff
@@ -303,3 +305,42 @@ def test_aws_based_external_account(
                 },
             },
         )
+
+def test_pluggable_external_account(
```
TimurSadykov, line 309: please add comments what is happening for readability, like for other tests i.e. "test_before_request_impersonation"

TimurSadykov (changes requested):
so far only comment request, in progress

TimurSadykov (approved):
LGTM, but please add comments to the test )

system_tests/system_tests_sync/test_external_accounts.py
ScruffyProdigy: Done

system_tests/system_tests_sync/test_external_accounts.py
```diff
@@ -303,3 +305,42 @@ def test_aws_based_external_account(
...
+    oidc_credentials, service_account_info, dns_access
+):
+    now = datetime.datetime.now()
+    unix_seconds = time.mktime(now.timetuple())
+    expiration_time = (unix_seconds + 1 * 60 * 60) * 1000
```
clundin25, line 314: nit: I think expiration time could be more descriptive, e.g. `expires_in_one_hour`

system_tests/noxfile.py
```diff
@@ -379,8 +380,9 @@ def mtls_http(session):
     )
 
 
-@nox.session(python=PYTHON_VERSIONS_SYNC)
+@nox.session(python=PYTHON_VERSIONS_ASYNC)
 def external_accounts(session):
+    session.env[ALLOW_PLUGGABLE_ENV] = "1"
     session.install(
         *TEST_DEPENDENCIES_SYNC,
```
clundin25, line 387: Does this need to change as well?

system_tests/noxfile.py
```diff
@@ -379,8 +380,9 @@ def mtls_http(session):
     )
 
 
-@nox.session(python=PYTHON_VERSIONS_SYNC)
+@nox.session(python=PYTHON_VERSIONS_ASYNC)
```
clundin25, line 383: What's the context for this change? Could we use both versions to avoid losing coverage for `PYTHON_VERSIONS_SYNC`?

clundin25:
You'll need a fresh token for the Kokoro system tests. I pushed one [here](https://github.com/clundin25/google-auth-library-python/commit/bdf9829544b5c119d5cc86e074b07e9e3b969e5a) that you can pull in

system_tests/noxfile.py
ScruffyProdigy: The previous version of pluggable auth required async libraries.  I think Chuan updated it to not require that.  I guess we'll find out when the tests run

system_tests/noxfile.py
ScruffyProdigy: Yes, without this environment variable set, pluggable auth tests will fail

clundin25: Sorry, I meant the test dependencies, only `*TEST_DEPENDENCIES_SYNC` was being used and I assumed they would need to be updated to `*TEST_DEPENDENCIES_ASYNC` to reflect the nox session change

ScruffyProdigy: Ah, gotcha.  I updated the nox session to use SYNC instead, because I believe Chuan updated the code to work with 2.7

clundin25: Seems like it [failed](https://source.cloud.google.com/results/invocations/2b166d53-6290-4153-a643-8a353f83a9ca/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fgoogle-auth-library-python%2Fpresubmit%2Fsystem-3.7/log) with 2.7 :( 

I don't have the context for this so I'm okay with moving it back to how you had it

PR #1083, opened 2022-07-17
docs: fix return type of get_service_account_token

jangjunha:
While reading, I found a mismatch between the return type of the code and the doctoring and fixed it.


clundin25 (approved):
Thanks for fixing this!

PR #1085, opened 2022-07-20
bugfix: fix async certificate decoding

### LINKED ISSUE #1050: Async _fetch_certs does not decode certs - TypeError Object of type bytes is not JSON serializable
https://github.com/googleapis/google-auth-library-python/blob/7da6cfcbd1b3b94e63ec85b5a4382d32eadba52d/google/oauth2/_id_token_async.py#L98

There is a bug in the code. The json.dumps returns a `TypeError: Object of type bytes is not JSON serializable` while data instead needs to be decoded `data.decode('utf-8')` before converting to dictionary.

The correct approach is being used in the id_token.py method.

clundin25:
The async _cert_fetch implementation was not properly decoding
certificates into a utf-8 string. This updates the code and tests to
decode the certificates into strings. 

Without this fix, the code will raise a `ValueError` exception, because the data type passed to `json.dumps` is not expected.

This resolves https://github.com/googleapis/google-auth-library-python/issues/1050.

### DIFF
--- a/google/oauth2/_id_token_async.py
+++ b/google/oauth2/_id_token_async.py
@@ -95,7 +95,7 @@ async def _fetch_certs(request, certs_url):
 
     data = await response.data.read()
 
-    return json.loads(json.dumps(data))
+    return json.loads(data.decode("utf-8"))
 
 
 async def verify_token(
--- a/tests_async/oauth2/test_id_token.py
+++ b/tests_async/oauth2/test_id_token.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import os
 
 import mock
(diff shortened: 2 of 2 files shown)


google/oauth2/_id_token_async.py
```diff
@@ -95,7 +95,7 @@ async def _fetch_certs(request, certs_url):
 
     data = await response.data.read()
 
-    return json.loads(json.dumps(data))
+    return json.loads(data.decode("utf-8"))
```
TimurSadykov, line 98: This works )

and a question to  @arithmetic1728: I wonder if there is a way we can decode all the responses at the time we get them... so we don't have to remember this next time

TimurSadykov (approved):

google/oauth2/_id_token_async.py
arithmetic1728: I think this can be achieved by adding a new `decoded_data` property in parallel to the `data` property in the `_Response` class

https://github.com/googleapis/google-auth-library-python/blob/6f49d1fa1dff62c5f2ed0cf198507568f5c4296f/google/auth/transport/requests.py#L73-L75

sai-sunder-s (approved):

PR #1091, opened 2022-07-25
fix: make expiration_time optional in response schema

BigTailWolf:


sai-sunder-s (approved):

google/auth/pluggable.py
```diff
@@ -301,11 +304,7 @@ def _parse_subject_token(self, response):
                     response["code"], response["message"]
                 )
             )
-        if "expiration_time" not in response:
-            raise ValueError(
-                "The executable response is missing the expiration_time field."
-            )
-        if response["expiration_time"] < time.time():
+        if "expiration_time" in response and response["expiration_time"] < time.time():
```
lsirac, line 307: Let's enforce expiration_time if an output_file was specified. So an exception should be thrown even if calling the executable and the expiration_time is not present if the credential configuration specifies an output_file.

tests/test_pluggable.py
```diff
@@ -652,9 +654,35 @@ def test_retrieve_subject_token_missing_expiration_time(self):
                 _ = credentials.retrieve_subject_token(None)
 
             assert excinfo.match(
-                r"The executable response is missing the expiration_time field."
+                r"The output file response must contain the expiration_time field when success=true."
             )
 
+    @mock.patch.dict(os.environ, {"GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES": "1"})
```
lsirac, line 660: Add another test that validates that an error is thrown when reading from the output_file and expiration_time is missing

tests/test_pluggable.py
```diff
@@ -652,9 +654,35 @@ def test_retrieve_subject_token_missing_expiration_time(self):
...
+        with mock.patch(
+            "subprocess.run",
+            return_value=subprocess.CompletedProcess(
+                args=[],
+                stdout=json.dumps(EXECUTABLE_SUCCESSFUL_OIDC_RESPONSE).encode("UTF-8"),
+                returncode=0,
+            ),
+        ):
+            credentials = self.make_pluggable(credential_source=CREDENTIAL_SOURCE)
+            credentials.retrieve_subject_token(None)
```
lsirac, line 684: validate expected value

tests/test_pluggable.py
```diff
@@ -630,7 +630,9 @@ def test_retrieve_subject_token_missing_error_code_message(self):
             )
 
     @mock.patch.dict(os.environ, {"GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES": "1"})
-    def test_retrieve_subject_token_missing_expiration_time(self):
+    def test_retrieve_subject_token_without_expiration_time_should_fail_when_using_output_file(
```
lsirac, line 633: when_using_output_file is a little bit misleading because here we're reading from STDOUT.

maybe 'when_output_file_specified'?

lsirac (approved):

google/auth/pluggable.py
```diff
@@ -301,11 +301,14 @@ def _parse_subject_token(self, response):
...
                 )
             )
-        if "expiration_time" not in response:
+        if (
+            "expiration_time" not in response
+            and self._credential_source_executable_output_file
+        ):
             raise ValueError(
-                "The executable response is missing the expiration_time field."
+                "The output file response must contain the expiration_time field when success=true."
```
lsirac, line 309: This isn't only the output file response, it could also be the response to STDOUT. Need to update this

"The response must contain an expiration_time for successful responses when an output_file has been specified in the configuration."

lsirac:
Update user guide:

https://github.com/googleapis/google-auth-library-python/blob/main/docs/user-guide.rst

docs/user-guide.rst
```diff
@@ -429,24 +429,29 @@ These are all required fields for an error response. The code and
...
+- ``saml_response``: The 3rd party SAML response.
+- ``expiration_time``: The 3rd party subject token expiration time in seconds
+  (unix epoch time).
+- ``code``: The error code string.
+- ``message``: The error message.
+
+All response types must include both the ``version`` and ``success`` fields.
+Successful responses must include the ``token_type``, and one of ``id_token``
+or ``saml_response``.
+``expiration_time`` is only optional when ``output_file`` is not specified.
```
lsirac, line 453: nit: expiration_time is only required when an output_file is specified. 

(the reverse makes it confusing)

PR #1095, opened 2022-08-02
fix: refactor credential subclass parameters

ScruffyProdigy:
Parameters that are unused by the subclasses should be skipped and passed through via kwargs.  This will make the code more readable


google/auth/aws.py
```diff
@@ -349,35 +348,15 @@ class Credentials(external_account.Credentials):
     def __init__(
         self,
         audience,
-        subject_token_type,
```
clundin25, line 352: Is making this change backwards compatible?  I suspect it is not but I haven't done a refactor like this before.

google/auth/aws.py
```diff
@@ -349,35 +348,15 @@ class Credentials(external_account.Credentials):
...
-        client_secret=None,
-        quota_project_id=None,
-        scopes=None,
-        default_scopes=None,
+        **kwargs,
     ):
         """Instantiates an AWS workload external account credentials object.
 
         Args:
-            audience (str): The STS audience field.
```
clundin25, line 366: This class should still take the same args as before, right?

google/auth/aws.py
```diff
@@ -750,23 +721,7 @@ def from_info(cls, info, **kwargs):
...
-                "service_account_impersonation"
-            )
-            or {},
-            client_id=info.get("client_id"),
-            client_secret=info.get("client_secret"),
-            credential_source=info.get("credential_source"),
-            quota_project_id=info.get("quota_project_id"),
-            **kwargs
-        )
+        return super(Credentials, cls).from_info(info, **kwargs)
```
clundin25, line 724: This looks a lot nicer

google/auth/aws.py
ScruffyProdigy: Actually, most of the changes I've made are backwards compatible.  All I've done is instead of passing the parameters through explicitly, they're being passed through via kwargs, but they're still being passed through.  The only difference is with the from_info in aws.py. It will pass a workforce_pool_user_project=None through, but this shouldn't make a difference, because aws.Credentials would initialize external_account.Credentials, which has workforce_pool_user_project=None as the default value

google/auth/aws.py
ScruffyProdigy: The args are still the same as before.  Should I list out all of the parameters here, or should I mention kwargs, and to look at external_account.Credentials to see all of the values that can be set there?

google/auth/aws.py
ScruffyProdigy: Thank you :-)

google/auth/aws.py
clundin25: What if the arguments were positional? Do we need to pass `*args` as well?

clundin25:
Looks like you need to run `nox -s blacken` to get past the linter

google/auth/aws.py
ScruffyProdigy: I guess it's a balance between user convenience and maintainability.  It looks like we haven't done a 100% job at maintaining these as new arguments have gotten added, but I'll fix them up for now

ScruffyProdigy: Added documentation for args and kwargs

clundin25 (approved):

PR #1096, opened 2022-08-02
chore: add aion-sdks to code owners

sai-sunder-s:


clundin25 (approved):

TimurSadykov (changes requested):
we want to revisit this )

Issue #1100, opened 2022-08-06
Async certificate retrieving - problem still exists


tivaliy:
After the latest release the problem of retrieving certificate still exists for async mode. Here is the previous fix - https://github.com/googleapis/google-auth-library-python/issues/1050

#### Environment details
  
- Python version: Python 3.9
  - pip version: 22.1.2
  - `google-auth` version: 2.10.0

#### Steps to reproduce

```python
In [1]: from google.oauth2 import _id_token_async

In [2]: from google.auth.transport import _aiohttp_requests

In [3]: certs_url = "https://www.googleapis.com/oauth2/v1/certs"

In [4]: request = _aiohttp_requests.Request()

In [5]: response = await request(certs_url, method="GET")

In [6]: response.status
Out[6]: 200

In [7]: data = await response.data.read()

In [8]: import json

In [9]: json.loads(data.decode("utf-8"))
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Input In [10], in <cell line: 1>()
----> 1 json.loads(data.decode("utf-8"))

UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte
```

I suppose that the problem is in `'Content-Encoding': 'gzip'` (`response`), which is not supported by `Request` (https://github.com/googleapis/google-auth-library-python/blob/e9b83040c8073a33a0713499a1d0806e02d0aabb/google/auth/transport/_aiohttp_requests.py#L140):

```python
    def __init__(self, session=None):
        # TODO: Use auto_decompress property for aiohttp 3.7+
        if session is not None and session._auto_decompress:
            raise ValueError(
                "Client sessions with auto_decompress=True are not supported."
            )
        self.session = session
```

instead, the following code should be used:

```python
In [12]: response = await request(certs_url, method="GET")

In [13]: data = await response.content()

In [14]: json.loads(data)
Out[14]:
{'fda1066453dc9dc3dd933a41ea57da3ef242b0f7': '-----BEGIN CERTIFICATE-----\nMIIDJzCCAg+gAwIBAgIJAIVR59GV94ZKMA0GCSqGSIb3DQEBBQUAMDYxNDAyBgNV\nBAMMK2ZlZ
...........................
[...]

PR #1101, opened 2022-08-06
Fix: Async certificate retrieving

tivaliy:
This PR fixes bug reported in #1100


tests_async/oauth2/test_id_token.py
```diff
@@ -36,6 +36,9 @@ def make_request(status, data=None):
         response.data.read = mock.AsyncMock(
             spec=["__call__"], return_value=json.dumps(data).encode("utf-8")
         )
```
clundin25, line 38: Can you remove these lines? They aren't needed anymore.

clundin25:
Looks good. I have one minor outstanding comment. For the system tests to pass either merge https://github.com/googleapis/google-auth-library-python/pull/1105 or allow me to merge it, and then rebase it into your changes.

Thanks for this contribution!

tests_async/oauth2/test_id_token.py
tivaliy: Done

clundin25 (approved):

PR #1103, opened 2022-08-08
feat: add integration tests for configurable token lifespan

ScruffyProdigy:


system_tests/system_tests_sync/test_external_accounts.py
```diff
@@ -202,6 +201,47 @@ def test_file_based_external_account_with_configure_token_lifetime(
...
+    def check_impersonation_expiration():
+        # First, get the default credentials.
+        credentials, _ = google.auth.default(
+            scopes=["https://www.googleapis.com/auth/cloud-platform.read-only"],
+            request=http_request,
+        )
+
+        utcmax = _helpers.utcnow() + datetime.timedelta(seconds=TOKEN_LIFETIME_SECONDS)
+        utcmin = utcmax - datetime.timedelta(seconds=BUFFER_SECONDS)
+        assert utcmin < credentials._impersonated_credentials.expiry <= utcmax
```
lsirac, line 217: should just be .token?

lsirac (approved):

system_tests/system_tests_sync/test_external_accounts.py
ScruffyProdigy: I'm just checking the expiry here, to make sure it's reasonably close to the value we set it to.  I'm not checking the value of the token at all

BigTailWolf (approved):

sai-sunder-s (approved):

PR #1113, opened 2022-08-23
feat: Retry behavior

clundin25:
* Introduce `retryable` property to auth library exceptions. This can be
  used to determine if an exception should be retried.
* Introduce `should_retry` parameter to token endpoints. If set to `False`
  the auth library will not retry failed requests. If set to `True` the
  auth library will retry failed requests. The default value is `True`
  to maintain existing behavior.
* Expanded list of HTTP Status codes that will be retried.
* Modified retry behavior to use exponential backoff.
* Increased default retry attempts from 2 to 3.


google/oauth2/_client.py
```diff
@@ -129,32 +176,51 @@ def _token_endpoint_request_no_throw(
...
+    response, response_data = _perform_request()
+
+    if response.status == http_client.OK:
+        return True, response_data, None
+
+    retryable_error = _can_retry(
+        status_code=response.status, response_data=response_data
+    )
+
+    if not retryable_error or not should_retry:
```
sai-sunder-s, line 196: This condition makes me think whether we should rename `should_retry` to something like `can_retry` because even if user says `should_retry` `true`, we will not retry if error code says its not retryable.

google/oauth2/_client.py
```diff
@@ -59,7 +66,41 @@ def _handle_error_response(response_data):
...
+      bool: True if the response is retryable. False otherwise.
+    """
+    try:
+        # For a failed response, response_body could be a string
+        error_desc = response_data.get("error_description") or ""
+        error_code = response_data.get("error") or ""
+
+        # Per Oauth 2.0 RFC https://www.rfc-editor.org/rfc/rfc6749.html#section-4.1.2.1
+        # This is needed because a redirect will not return a 500 status code.
+        retryable_error_descriptions = {"internal_failure", "server_error"}
```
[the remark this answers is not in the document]
clundin25: `internal_failure` was the existing condition. Not sure the context for it, but it is not in the RFC as you said.

`temporarily_unavailable` IMO we should not retry, as the server is trying to load shed or is overwhelmed.

google/oauth2/_client.py
clundin25: That sounds good to me, `should_retry` felt a bit off.

google/oauth2/_client.py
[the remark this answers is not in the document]
sai-sunder-s: `temporarily_unavailable` is retryable I feel because it was of no fault from the user's request and it could succeed by retrying. If the server is receiving too many requests from this specific client/user and wants them to backoff, then the error code will be 429.
If server is overloaded due to cumulative requests from multiple clients, it can still return 429 to some requests to control the load. When 429 is returned we should not retry.
Since `temporarily_unavailable` comes for 503, I think we could retry.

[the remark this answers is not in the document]
clundin25: Ok, that sounds good to me. With the exponential backoff change I think it reasonable to assume that the impact on the servers will be reduced.

I'll make this change.

google/oauth2/_client.py
```diff
@@ -59,7 +66,41 @@ def _handle_error_response(response_data):
...
+      bool: True if the response is retryable. False otherwise.
+    """
+    try:
+        # For a failed response, response_body could be a string
+        error_desc = response_data.get("error_description") or ""
+        error_code = response_data.get("error") or ""
+
+        # Per Oauth 2.0 RFC https://www.rfc-editor.org/rfc/rfc6749.html#section-4.1.2.1
+        # This is needed because a redirect will not return a 500 status code.
+        retryable_error_descriptions = {"internal_failure", "server_error", "temporarily_unavailable"}
```
TimurSadykov, line 92: Why we still have it? we have a status code check below, those are more reliable than error descriptions

clundin25: My understanding from the spec is that these errors _may_ service as a non-500 status code in case the request was redirected.

google/auth/transport/__init__.py
```diff
@@ -29,9 +29,18 @@
 import six
 from six.moves import http_client
 
+DEFAULT_RETRYABLE_STATUS_CODES = (
```
TimurSadykov, line 32: Where is 429?

google/oauth2/_client.py
```diff
@@ -30,27 +30,34 @@
...
 _REFRESH_GRANT_TYPE = "refresh_token"
 
 
-def _handle_error_response(response_data):
+def _handle_error_response(response_data, retryable_error):
     """Translates an error response into an exception.
 
     Args:
         response_data (Mapping | str): The decoded response data.
+        retryable_error Optional[bool]: A boolean indicating if an error is retry able.
```
TimurSadykov, line 50: typo: retry able

google/oauth2/_client.py
```diff
@@ -59,7 +66,41 @@ def _handle_error_response(response_data):
...
+        retryable_error_descriptions = {"internal_failure", "server_error", "temporarily_unavailable"}
+
+        if any(e in retryable_error_descriptions for e in (error_code, error_desc)):
+            return True
+
+    except AttributeError:
+        pass
+
+    return (
+        status_code in transport.DEFAULT_REFRESH_STATUS_CODES
```
TimurSadykov, line 101: could you please show where is the refresh logic in case of refresh failure?

TimurSadykov (changes requested):
still looking )

google/oauth2/_client.py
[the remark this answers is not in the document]
TimurSadykov: This section of the RFC is related to 3LO right? its a redirect for code exchange. It is not supported by the library?

google/oauth2/_client.py
TimurSadykov: + it seems to be related to 3LO that is not supported

TimurSadykov: i would check how this was added, as original implementation or related to a fix

clundin25: @TimurSadykov IIUC this is the original implementation https://github.com/googleapis/google-auth-library-python/pull/368.

Bug report https://bugs.chromium.org/p/chromium/issues/detail?id=1001206

google/auth/transport/__init__.py
clundin25: Python2's six library does not contain a 429 code. I can add one manually as a solution, or we can add it once Python 2 support is dropped

google/oauth2/_client.py
clundin25: I will remove this. This lingered from when I was including the google auth code in this change.

google/oauth2/_client.py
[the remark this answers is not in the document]
clundin25: I don't have the context for why this is included

google/oauth2/_client.py
```diff
@@ -30,27 +30,34 @@
...
     Args:
         response_data (Mapping | str): The decoded response data.
+        retryable_error Optional[bool]: A boolean indicating if an error is retryable.
+            Defaults to False.
 
     Raises:
         google.auth.exceptions.RefreshError: The errors contained in response_data.
     """
+
+    retryable_error = retryable_error if retryable_error else False
```
sai-sunder-s, line 57: why can't we set a default value when declaring the argument?

google/oauth2/_client.py
```diff
@@ -59,7 +66,42 @@ def _handle_error_response(response_data):
...
+            "temporarily_unavailable",
+        }
+
+        if any(e in retryable_error_descriptions for e in (error_code, error_desc)):
+            return True
+
+    except AttributeError:
+        pass
+
+    return status_code in transport.DEFAULT_RETRYABLE_STATUS_CODES
```
sai-sunder-s, line 104: nit: This check seems very light. Why not do this first and if false, then do the description check?

sai-sunder-s (approved):

google/oauth2/_client.py
clundin25: The default parameter is used if the argument is omitted. In this case retryable_error is `None` if the request was successful. This is a defensive check in case the API is used incorrectly

google/oauth2/_client.py
clundin25: Makes sense to me. Added

google/oauth2/_client.py
```diff
@@ -129,32 +180,51 @@ def _token_endpoint_request_no_throw(
...
-            except ValueError:
-                response_data = response_body
-            return False, response_data
-
-    return True, response_data
+        except ValueError:
+            response_data = response_body
+        return response, response_data
+
+    response, response_data = _perform_request()
```
TimurSadykov, line 191: There is some copy-paste that we can combine in a single method? 

Probably a "while True:" loop?

google/oauth2/_client_async.py
```diff
@@ -83,26 +83,44 @@ async def _token_endpoint_request_no_throw(
...
 
-        response_data = json.loads(response_body)
+        try:
+            response_data = json.loads(response_body)
+        except ValueError:
+            response_data = response_body
+
+        return response, response_data
+
+    response, response_data = await _perform_request()
```
TimurSadykov, line 93: similarly, should be a way to refactor so we do perform_request and check retryable_error one

TimurSadykov (changes requested):
Overall looks good, lets just try to remove some dup code

google/oauth2/_client.py
```diff
@@ -214,20 +293,24 @@ def jwt_grant(request, token_uri, assertion):
...
+    response_data = _token_endpoint_request(
+        request, token_uri, body, can_retry=can_retry
+    )
 
     try:
         access_token = response_data["access_token"]
     except KeyError as caught_exc:
-        new_exc = exceptions.RefreshError("No access token in response.", response_data)
+        new_exc = exceptions.RefreshError(
+            "No access token in response.", response_data, retryable=True
```
TimurSadykov, line 304: I understand it is existing behavior.. but i wonder how it works for example when refresh fail due to reauth? 
How this does not retry indefinitely?

google/oauth2/_client.py
clundin25: I'll condense the code some more.

I am avoiding the while loop because I don't think the first request and the retries belong in the same loop. The iterator I have written will sleep on the first request. That would be bad :D

google/oauth2/_client.py
clundin25: Good point, I changed this to False

clundin25:
@parthea do you mind reviewing this for Python readability?

google/oauth2/_client_async.py
```diff
@@ -155,20 +173,24 @@ async def jwt_grant(request, token_uri, assertion):
...
+    response_data = await _token_endpoint_request(
+        request, token_uri, body, can_retry=can_retry
+    )
 
     try:
         access_token = response_data["access_token"]
     except KeyError as caught_exc:
-        new_exc = exceptions.RefreshError("No access token in response.", response_data)
+        new_exc = exceptions.RefreshError(
+            "No access token in response.", response_data, retryable=True
```
TimurSadykov, line 184: why this is by default Retryable?

TimurSadykov: if it is expected why we expect the retry to succeed?

google/oauth2/_client_async.py
```diff
@@ -195,12 +218,16 @@ async def id_token_jwt_grant(request, token_uri, assertion):
...
+    response_data = await _token_endpoint_request(
+        request, token_uri, body, can_retry=can_retry
+    )
 
     try:
         id_token = response_data["id_token"]
     except KeyError as caught_exc:
-        new_exc = exceptions.RefreshError("No ID token in response.", response_data)
+        new_exc = exceptions.RefreshError(
+            "No ID token in response.", response_data, retryable=True
```
TimurSadykov, line 229: same question on a default retryable

TimurSadykov (commented):
couple more questions on default retryables.

google/oauth2/_client_async.py
clundin25: My interpretation was that should this happen we have received a malformed request or error, that ideally should have been handled at the transport layer.

I think the best path is to retry it to see if that resolves the issue. 

If you disagree I am happy to set this to `False`. It is hard to decide what is right :)

TimurSadykov: Let's have them as false since that is the original behavior

We don't want to retry those because - it is not a valid case, we don't expect it often enough to retry just in case

google/oauth2/_client_async.py
TimurSadykov: like above, lets switch to false

TimurSadykov (approved):
LGTM, thanks!

PR #1126, opened 2022-08-30
ci: Stop setting CLOUDSDK_PYTHON for system_tests/noxfile.py

jameslynnwu:
google-cloud-sdk.tar.gz is now compatible with Python 3.


system_tests/noxfile.py
```diff
@@ -70,11 +70,6 @@
 # The full path to the gcloud cli executable.
 GCLOUD = str(CLOUD_SDK_INSTALL_DIR.join("bin", "gcloud"))
 
-# gcloud requires Python 2 and doesn't work on 3, so we need to tell it
-# where to find 2 when we're running in a 3 environment.
-CLOUD_SDK_PYTHON_ENV = "CLOUDSDK_PYTHON"
-CLOUD_SDK_PYTHON = which("python2", None)
```
sai-sunder-s, line 76: Has gcloud stopped supporting 2.7 or is this change because gcloud supports python 3 as well, so we can just use it?

sai-sunder-s (approved):

parthea (approved):

system_tests/noxfile.py
jameslynnwu: gcloud supports python 3 now, and we will stop supporting python 2 soon, so we shouldn't choose py2.

dandhlee: If it's "soon", shouldn't we still support until it is for sure deprecated?

arithmetic1728:
close the PR as this is removed in #1278

PR #1131, opened 2022-08-31
feat: implement pluggable auth interactive mode

BigTailWolf:
For interactive mode:
1. Always using output to read the result.
2. Make "expiration_time" optional for all mode.
3. Implement interactive mode run executable
4. Implement `revoke()` function.
4. Refactor tests


google/auth/pluggable.py
```diff
@@ -132,6 +137,20 @@ def __init__(
...
+            )
+        elif (
+            self._credential_source_executable_interactive_timeout_millis
+            < 5 * 60 * 1000
+            or self._credential_source_executable_interactive_timeout_millis
+            > 30 * 60 * 1000
+        ):
+            raise ValueError("Interactive timeout must be between 5 and 30 minutes.")
+        if self._interactive and not self._credential_source_executable_output_file:
+            raise ValueError("Output file must be specified in interactive mode")
```
lsirac, line 152: An `output_file` must be specified in the credential configuration for interactive mode.

aeitzman (changes requested):
First pass

google/auth/pluggable.py
```diff
@@ -105,6 +106,7 @@ def __init__(
             raise ValueError(
                 "Missing credential_source. The credential_source is not a dict."
             )
+        self._interactive = os.environ.get("GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE") == "1"
```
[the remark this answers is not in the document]
lsirac: +1 GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE is set by the auth library, not the user

google/auth/pluggable.py
```diff
@@ -183,31 +211,61 @@ def retrieve_subject_token(self, request):
...
-                stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
-                env=env,
-            )
-            if result.returncode != 0:
-                raise exceptions.RefreshError(
-                    "Executable exited with non-zero return code {}. Error: {}".format(
-                        result.returncode, result.stdout
-                    )
+        if self._interactive:
```
lsirac, line 214: Let's avoid duplicating code here

google/auth/pluggable.py
```diff
@@ -171,9 +198,10 @@ def retrieve_subject_token(self, request):
         env = os.environ.copy()
         env["GOOGLE_EXTERNAL_ACCOUNT_AUDIENCE"] = self._audience
         env["GOOGLE_EXTERNAL_ACCOUNT_TOKEN_TYPE"] = self._subject_token_type
-        env[
-            "GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE"
-        ] = "0"  # Always set to 0 until interactive mode is implemented.
+        env["GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE"] = (
```
lsirac, line 201: Implementation is missing GOOGLE_EXTERNAL_ACCOUNT_ID and GOOGLE_EXTERNAL_ACCOUNT_REVOKE

google/auth/pluggable.py
```diff
@@ -132,6 +137,20 @@ def __init__(
         ):
             raise ValueError("Timeout must be between 5 and 120 seconds.")
 
+        if not self._credential_source_executable_interactive_timeout_millis:
```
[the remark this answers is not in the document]
BigTailWolf: Hi Leo, do you know where can we figure out whether it's workload?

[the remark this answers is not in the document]
lsirac: You can check the audience

google/auth/pluggable.py
```diff
@@ -183,31 +211,61 @@ def retrieve_subject_token(self, request):
...
                 )
-        except Exception:
-            raise
+                if result.returncode != 0:
+                    raise exceptions.RefreshError(
+                        "Executable exited with non-zero return code {}.".format(
+                            result.returncode
+                        )
+                    )
+            except Exception:
```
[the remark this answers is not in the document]
BigTailWolf: Fixed all kinds of this issue

google/auth/pluggable.py
```diff
@@ -141,12 +157,13 @@ def retrieve_subject_token(self, request):
             raise ValueError(
                 "Executables need to be explicitly allowed (set GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES to '1') to run."
             )
+        interactive_mode = os.environ.get("GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE") == "1"
```
aeitzman, line 160: This still doesn't match with the design, library should be setting this environment var, not reading it.

google/auth/pluggable.py
```diff
@@ -264,10 +291,11 @@ def _parse_subject_token(self, response):
             )
         if (
             "expiration_time" not in response
+            and not os.environ.get("GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE") == "1"
```
aeitzman, line 294: same here, don't read the environment variable.

google/auth/pluggable.py
```diff
@@ -116,6 +117,9 @@ def __init__(
         self._credential_source_executable_timeout_millis = self._credential_source_executable.get(
             "timeout_millis"
         )
+        self._credential_source_executable_interactive_timeout_millis = self._credential_source_executable.get(
+            "interactive_timeout_millis"
```
aeitzman, line 121: should there be a check that an output file has been provided if interactive timeout millis is present? Check could also happen in the retrieve_subject_token method, just think we need a clear error message if someone tries to run interactive mode with no output file.

aeitzman (changes requested):

google/auth/pluggable.py
```diff
@@ -171,9 +192,7 @@ def retrieve_subject_token(self, request):
         env = os.environ.copy()
         env["GOOGLE_EXTERNAL_ACCOUNT_AUDIENCE"] = self._audience
         env["GOOGLE_EXTERNAL_ACCOUNT_TOKEN_TYPE"] = self._subject_token_type
-        env[
-            "GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE"
-        ] = "0"  # Always set to 0 until interactive mode is implemented.
+
```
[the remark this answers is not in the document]
BigTailWolf: GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE is set via `os.environ.copy()` I assume. The `os.environ` won't miss this variable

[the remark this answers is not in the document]
aeitzman: That variable shouldn't be set already in the env, its not supposed to be a user set variable

google/auth/pluggable.py
BigTailWolf: Should be in `retrieve_subject_token` since we can use both interactive mode and non-interactive mode for a config.
Then we can only check it when actually need to use the output file which is the retrieve_subject_token.

google/auth/pluggable.py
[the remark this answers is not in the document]
BigTailWolf: Oh Thanks! I just saw the property method in the parent class.

google/auth/pluggable.py
sai-sunder-s: Why not allow the user to set as well? Similar to `GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES`

google/auth/pluggable.py
```diff
@@ -132,6 +137,20 @@ def __init__(
         ):
             raise ValueError("Timeout must be between 5 and 120 seconds.")
 
+        if not self._credential_source_executable_interactive_timeout_millis:
+            self._credential_source_executable_interactive_timeout_millis = (
+                5 * 60 * 1000
```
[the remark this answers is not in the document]
BigTailWolf: Make sense. Do you think it's good to put it in this file or somewhere else that can be referred more widely?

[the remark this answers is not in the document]
sai-sunder-s: Lets start with this file. If we find this value repeating elsewhere, we can pull out to a common place.

google/auth/pluggable.py
```diff
@@ -155,7 +174,15 @@ def retrieve_subject_token(self, request):
                 try:
                     # If the cached response is expired, _parse_subject_token will raise an error which will be ignored and we will call the executable again.
                     subject_token = self._parse_subject_token(response)
+                    if (
+                        self._interactive and "expiration_time" not in response
```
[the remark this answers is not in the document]
sai-sunder-s: Why non-interactive mode does not treat missing `expiration_time` as expired?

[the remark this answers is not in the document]
BigTailWolf: For non-interactive mode, we already do a sanity check in the `_parse_subject_token` for `expiration_time` if an output file is specified. We treat missing expiration time as malformed and throw an exception in case of an output filed is specified.

google/auth/pluggable.py
```diff
@@ -183,32 +222,87 @@ def retrieve_subject_token(self, request):
...
+
+        result = subprocess.run(
+            self._credential_source_executable_command.split(),
+            timeout=self._credential_source_executable_interactive_timeout_millis
+            / 1000,
+            env=env,
+        )
+
+        if result.returncode != 0:
+            raise exceptions.RefreshError("Auth revoke failed on executable.")
```
aeitzman, line 302: Include return code in the exception

BigTailWolf (commented):
Required changes addressed

google/auth/pluggable.py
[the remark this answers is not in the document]
BigTailWolf: Hi @sai-sunder-s , I changed the check of "expiration_time" and unified the logic between non-interactive and interactive. This will case less confusion to users.
1. Make it as optional regardless of interactive or non-interactive
2. Treat missing expiration time as "expired" when retrieving from file

The reason we want it being optional is:
It doesn't make sense to provide an expiration time when user don't want to cache it regardless interactive or non-interactive.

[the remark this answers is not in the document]
aeitzman: Will this make it less confusing? If we just ignore an invalid output file in non-interactive without throwing an error, the users may not realize that their output file is not being used even if they are creating it. It doesn't make sense with the current implementation for a non-interactive mode output file to not have an expiration time.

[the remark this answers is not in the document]
BigTailWolf: Just need to document a warning "Missing expiration time will cause the file not being used for caching".
This will keep the interactive and non-interactive treat "expiration_time" in the same way.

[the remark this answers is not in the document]
aeitzman: That would also either:
1: create a difference between the non-interactive mode behavior of the python library and the other 3 already released SDK's
2: Make the behavior of the other 3 SDK's not really make sense if we change them to match, then they wouldn't be checking for expiration_time even though they don't support interactive mode.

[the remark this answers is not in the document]
BigTailWolf: OK. Make sense. I forget we have other libraries already made that check.

google/auth/pluggable.py
```diff
@@ -183,32 +224,98 @@ def retrieve_subject_token(self, request):
...
+            raise exceptions.RefreshError(
+                "Pluggable auth is only supported for python 3.6+"
+            )
+
+        # Inject variables
+        env = os.environ.copy()
+        env["GOOGLE_EXTERNAL_ACCOUNT_AUDIENCE"] = self._audience
+        env["GOOGLE_EXTERNAL_ACCOUNT_TOKEN_TYPE"] = self._subject_token_type
+        env["GOOGLE_EXTERNAL_ACCOUNT_ID"] = self.service_account_email
+        env["GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE"] = "1"
```
aeitzman, line 293: Its a little confusing to me that we set the interactive flag here(which makes sense sense since its only available in interactive mode) but we don't actually pipe the stdin/stdout. should make sure to clarify in the docs what we expect the executables to do when both flags are set.

aeitzman (approved):

google/auth/pluggable.py
BigTailWolf: Added stdin/stdout since it might be needed from the executable.

google/auth/pluggable.py
```diff
@@ -212,18 +212,15 @@ def retrieve_subject_token(self, request):
...
-        if self._service_account_impersonation_url is not None:
-            env[
-                "GOOGLE_EXTERNAL_ACCOUNT_IMPERSONATED_EMAIL"
-            ] = self.service_account_email
-        if self._credential_source_executable_output_file is not None:
-            env[
-                "GOOGLE_EXTERNAL_ACCOUNT_OUTPUT_FILE"
-            ] = self._credential_source_executable_output_file
+        env["GOOGLE_EXTERNAL_ACCOUNT_IMPERSONATED_EMAIL"] = (
+            self.service_account_email or ""
```
aeitzman, line 219: why was this switched to filling with "" vs just not setting it at all?

BigTailWolf: Redundant change by an unsophisticated idea. I'll revert.

google/auth/pluggable.py
```diff
@@ -183,32 +226,112 @@ def retrieve_subject_token(self, request):
...
+        )
+
+        if result.returncode != 0:
+            raise exceptions.RefreshError(
+                "Auth revoke failed on executable. Exit with non-zero return code {}".format(
+                    result.returncode
+                )
+            )
+
+        # TODO: clear cache when the in memory cache feature implemented.
```
lsirac, line 320: I think we should remove this comment, as we currently don't have plans to cache the response in memory.

google/auth/pluggable.py
```diff
@@ -183,32 +226,112 @@ def retrieve_subject_token(self, request):
...
+        response = json.loads(result.stdout.decode("utf-8")) if result.stdout else None
+        if not response and self._credential_source_executable_output_file is not None:
+            response = json.load(
+                open(self._credential_source_executable_output_file, encoding="utf-8")
+            )
 
+        subject_token = self._parse_subject_token(response)
         return subject_token
 
+    def revoke(self, request):
```
lsirac, line 262: Let's refactor this to re-use the existing code that handles calling the executable

google/auth/pluggable.py
```diff
@@ -183,32 +226,112 @@ def retrieve_subject_token(self, request):
...
+                "Auth revoke failed on executable. Exit with non-zero return code {}".format(
+                    result.returncode
+                )
+            )
+
+        # TODO: clear cache when the in memory cache feature implemented.
+
+    @property
+    def external_account_id(self):
+        """Get the GOOGLE_EXTERNAL_ACCOUNT_ID which needs to be polulated to executable
```
lsirac, line 324: I think you should reword this here with a lot less detail and maybe explain that this should not be called. 

Something like: 

"Returns the external account identifier. When service account impersonation is used the identifier is the service account email. Without service account impersonation, this returns None, unless it is being used by the Google Cloud CLI which populates this field."

google/auth/pluggable.py
BigTailWolf: In the original design doc, I saw it was supposed to be called in the form of
```
credentials.revoke()
```

By "re-use the existing code" we are calling
```
credentials.retrieve_subject_token()
```
looks like doesn't make sense literally.

lsirac: Sure, but the code can be modified to be shared by the two code paths?

google/auth/pluggable.py
```diff
@@ -183,32 +226,112 @@ def retrieve_subject_token(self, request):
...
+        env["GOOGLE_EXTERNAL_ACCOUNT_REVOKE"] = "1"
+        if self._service_account_impersonation_url is not None:
+            env[
+                "GOOGLE_EXTERNAL_ACCOUNT_IMPERSONATED_EMAIL"
+            ] = self.service_account_email
+        env[
+            "GOOGLE_EXTERNAL_ACCOUNT_OUTPUT_FILE"
+        ] = self._credential_source_executable_output_file
+
+        result = subprocess.run(
```
lsirac, line 304: I think you need to validate the response?

BigTailWolf: Added a validation function

google/auth/pluggable.py
BigTailWolf: You mean `retrieve_subject_token()` and `revoke()` share a same function of calling subprocess?

BigTailWolf: I don't think that's common enough. They have different variables to populate, different stdin/stdout to assign and different post process( `_parse_subject_token` vs `_validate_response`)

BigTailWolf: Adding some function to let `retrieve_subject_token()` and `revoke()` to share.

tests/test_pluggable.py
```diff
@@ -316,6 +380,39 @@ def test_retrieve_subject_token_oidc_jwt(self):
 
             assert subject_token == self.EXECUTABLE_OIDC_TOKEN
 
+    @mock.patch.dict(os.environ, {"GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES": "1"})
+    def test_retrieve_subject_token_oidc_jwt_interactive_mode(self, tmpdir):
```
lsirac, line 384: You can probably just merge this with the test above using a for loop (loop over output file responses).

lsirac, line 384: You can probably just merge these tests into one by looping over the output file responses.

tests/test_pluggable.py
```diff
@@ -334,6 +431,38 @@ def test_retrieve_subject_token_saml(self):
 
             assert subject_token == self.EXECUTABLE_SAML_TOKEN
 
+    @mock.patch.dict(os.environ, {"GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES": "1"})
```
lsirac, line 434: Might also be worth it to add a test that tests changing between interactive/non-interactive modes, but not sure what you're able to validate given the way the code is written

docs/user-guide.rst
```diff
@@ -450,7 +450,8 @@ Response format fields summary:
 All response types must include both the ``version`` and ``success`` fields.
 Successful responses must include the ``token_type``, and one of ``id_token``
 or ``saml_response``.
-If output file is specified, ``expiration_time`` is mandatory.
+``expiration_time`` is optional. Missing ``expiration_time`` during retrieve
+from file will be treat as "expired".
```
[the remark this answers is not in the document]
BigTailWolf: Will do

tests/test_pluggable.py
BigTailWolf: Doing some refactoring

tests/test_pluggable.py
BigTailWolf: Refactoring to have all successful paths together will non-interactive and interactive together with all types of supported token type.

lsirac (approved):

sai-sunder-s (approved):

PR #1136, opened 2022-09-01
docs: Fix secure_authorized_channel argument positions.

clundin25:
This resolves
https://github.com/googleapis/google-cloud-python/issues/15201.


google/auth/transport/grpc.py
```diff
@@ -132,7 +132,7 @@ def secure_authorized_channel(
 
         # Create a channel.
         channel = google.auth.transport.grpc.secure_authorized_channel(
-            credentials, regular_endpoint, request,
+            credentials, request, regular_endpoint,
```
salrashid123, line 135: why use posisional args again here?  woudn't it be clearer using the names itself for clarity   `request=request, regular_endpoint=`...)

clundin25: I don't feel particularly strongly about this, but switching to keyword arguments would go against the style of the rest of the codebase.

salrashid123: sure, but using  incorrect positional led to filing this bug.   apart from styleguide stuff, the better way to think about it is what would be the best for the customer/reader

PR #1145, opened 2022-09-16
Fixed issues on passed scopes as set

zloidemon:


clundin25:
The documentation seems to indicate the scope is a sequence, so it could either be a list or set. I think the general case has been a list. Can you create an issue documenting what issue this is causing for you?

Also, please sign the CLA :)

PR #1153, opened 2022-09-26
Revert "fix: Update token refresh threshold from 20 seconds to 5 minutes (#1146)"

clundin25:
This reverts commit 261a56138fba33ff7d898ab5907a6098125fefef.

### DIFF
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -23,7 +23,11 @@ import six
 from six.moves import urllib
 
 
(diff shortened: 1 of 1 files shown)


TimurSadykov:
Why we do this? Did we get the confirmation that 5 min does not work?

clundin25:
> Why we do this? Did we get the confirmation that 5 min does not work?

I want to cut a release to unblock other work

sai-sunder-s (approved):

PR #1155, opened 2022-09-28
docs(samples): API key auth samples and tests

Sita04:
Samples and tests to authenticate using API keys

Kindly enable the API keys API for testing.


Sita04:
`google.api_core.exceptions.PermissionDenied: 403 API Keys API has not been used in project before or it is disabled. 
reason: "SERVICE_DISABLED"
domain: "googleapis.com"
metadata {
key: "service"
value: "apikeys.googleapis.com"
}`

samples/cloud-client/snippets/authenticate_with_api_key.py
```diff
@@ -0,0 +1,50 @@
...
+
+from google.cloud import language_v1
+
+
+def authenticate_with_api_key(quota_project_id: str, api_key_string: str) -> None:
+    """
+    Demonstrate authenticating with an API key for Google Language service.
+
+    // TODO(Developer):
+    //  1. Make sure you have the necessary permission to use Google Language service.
```
piaxc, line 25: Actually I think that is the point of API keys-- you don't need further permissions. You can remove this TODO

samples/cloud-client/snippets/authenticate_with_api_key.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Args:
+        quota_project_id: Google Cloud project id that should be used for quota and billing purposes.
+        api_key_string: The API key to authenticate to the service.
+    """
+
+    # Initialize the Language Service client and set the API key and the quota project id.
+    client = language_v1.LanguageServiceClient(client_options={"api_key": api_key_string,
+                                                               "quota_project_id": quota_project_id})
```
piaxc, line 34: This is interesting-- there's no way to create a client with an API key and no quota project? I thought that was the point of the API key, that it provided the project to be billed...

samples/cloud-client/snippets/create_api_key.py
```diff
@@ -0,0 +1,66 @@
...
+    # Create the API Keys client.
+    client = api_keys_v2.ApiKeysClient()
+
+    # Restrict the API key usage by specifying the target service and methods.
+    # The API key can only be used to authenticate the specified methods in the service.
+    api_target = api_keys_v2.ApiTarget()
+    api_target.service = "translate.googleapis.com"
+    api_target.methods = ["transate.googleapis.com.TranslateText"]
+
+    # Set the API restriction.
```
piaxc, line 44: Would it be possible to just create the key here, and then create snippets for the different types of restrictions? As it is organized in the doc page? https://cloud.google.com/docs/authentication/api-keys#adding-application-restrictions
Would be good to put in a note saying "you should always restrict your keys" here, but it would help me include the snippets if they followed the organization of the page more closely.

piaxc (changes requested):

samples/cloud-client/snippets/authenticate_with_api_key.py
Sita04: Done. Thanks

samples/cloud-client/snippets/create_api_key.py
Sita04: Pia, I've modified the samples to account for the doc page structure. PTAL.

samples/cloud-client/snippets/restrict_api_key_api.py
```diff
@@ -0,0 +1,65 @@
...
+
+from google.cloud import api_keys_v2
+from google.cloud.api_keys_v2 import Key
+
+
+def restrict_api_key_api(project_id: str, location: str, key_id: str) -> Key:
+    """
+    Restrict an API key. Restrictions specify which APIs can be called using the API key.
+
+    // TODO(Developer): Replace the variables before running the sample.
```
piaxc, line 25: Global comment-- we tell them to replace the variables, but I don't see any variables, I just see sample values. Might it be easier for them to determine what they needed to replace if we listed them out, and used some kind of variable? What is the convention for this?

piaxc (changes requested):
Thanks Sita, for the restructuring! Just one comment, about making it easier for developers to understand what they need to replace.

samples/cloud-client/snippets/restrict_api_key_api.py
Sita04: Hi Pia, the TODO is to replace the values of the variables in the method.
I'll leave this comment open to get the input of the sample reviewers group.

samples/cloud-client/snippets/authenticate_with_api_key.py
Sita04: I've followed up with this in an internal bug. Will update it here post response from the team.

Sita04:
Moved the samples to python-docs-samples repo.
PR: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/8455

PR #1159, opened 2022-10-07
feat: Introduce the functionality to override token_uri in credentials

sai-sunder-s:
There are scenarios where a custom token endpoint needs to be used. Currently the only way to provide a custom token endpoint is if a credential is created directly from the constructor. For example if a service account is loaded from a file, then custom token endpoint cannot be provided. This change introduces a method that overrides token endpoint after credential creation.


clundin25 (approved):

google/oauth2/credentials.py
```diff
@@ -254,6 +254,23 @@ def with_quota_project(self, quota_project_id):
             enable_reauth_refresh=self._enable_reauth_refresh,
         )
 
+    @_helpers.copy_docstring(credentials.CredentialsWithTokenUri)
```
TimurSadykov, line 257: Why this is needed? What credentials types it affects?

TimurSadykov:
Please add description for the change

google/oauth2/credentials.py
sai-sunder-s: Python library uses helper methods to override things like quota project, scopes etc. checkout `with_quota_project` method in many credentials.
This override method is added to all the credentials that have a token uri.

TimurSadykov: That is the question - do we need it for every credential type?

TimurSadykov: Why the auth credetials has it as not implemented exception and here we have a default implementation? It is preferable to have explicit implementation

sai-sunder-s: I am following the existing pattern https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/credentials.py#:~:text=class%20CredentialsWithQuotaProject(,quota%20project.%22)

sai-sunder-s: credentials.py is user credential. It is not a base class for other credentials. Base class is the credentials.py in auth package.

TimurSadykov: figured out that this is in fact user creds

TimurSadykov (approved):
LGTM

PR #1163, opened 2022-10-12
feat: Read Quota Project from Environment Variable

sai-sunder-s:
Introduce a new environment variable "GOOGLE_CLOUD_QUOTA_PROJECT" that can be used to specify a quota project id. This quota project will be used to override the quota project from the credential detected during ADC generation if another quota project was not provided explicitly.
This change also introduces a helper method to set the quota project from environment variable on the credential explicitly.


google/auth/credentials.py
```diff
@@ -149,6 +150,12 @@ def with_quota_project(self, quota_project_id):
         """
         raise NotImplementedError("This credential does not support quota project.")
 
+    def with_quota_project_from_environment(self):
+        quota_from_env = os.environ.get(environment_vars.GOOGLE_CLOUD_QUOTA_PROJECT)
+        if quota_from_env is not None:
+            return self.with_quota_project(quota_from_env)
```
clundin25, line 156:
```suggestion
        if quota_from_env:
            return self.with_quota_project(quota_from_env)
```

clundin25 (approved):

PR #1167, opened 2022-10-19
Raise HTTP error if encountered

joshtemple:
Currently, if the user is not authorized to make a request, there is no token in the response so the user receives a KeyError, which is less informative than the HTTPError returned from the API.

Raising the HTTP error solves this.

### DIFF
--- a/google/auth/impersonated_credentials.py
+++ b/google/auth/impersonated_credentials.py
@@ -433,6 +433,7 @@ class IDTokenCredentials(credentials.CredentialsWithQuotaProject):
             headers=headers,
             data=json.dumps(body).encode("utf-8"),
         )
+        response.raise_for_status()
 
         id_token = response.json()["token"]
         self.token = id_token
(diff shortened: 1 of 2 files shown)


google/auth/impersonated_credentials.py
```diff
@@ -433,6 +433,7 @@ def refresh(self, request):
             headers=headers,
             data=json.dumps(body).encode("utf-8"),
         )
+        response.raise_for_status()
```
clundin25, line 436: Could you write a test for this?

joshtemple: @clundin25, I've attempted to write a test. Let me know if you'd like me to approach it differently.

arithmetic1728: `refresh` method normally throws `RefreshError` (https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/exceptions.py#L35), could you maybe check the status code and raise a `RefreshError` instead?

chalmerlowe:
Closing in deference to a PR #1897 that addresses some minor issues with this approach.

PR #1168, opened 2022-10-24
feat: add token_info_url to external account credentials

ScruffyProdigy:
Also some minor refactoring


google/auth/external_account.py
```diff
@@ -161,13 +164,25 @@ def info(self):
...
+            type=_EXTERNAL_ACCOUNT_JSON_TYPE,
+            service_account_impersonation=config_info.pop(
+                "service_account_impersonation_options"
+            ),
+        )
+        config_info.pop("scopes")
+        config_info.pop("default_scopes")
+        return {key: value for key, value in config_info.items() if value is not None}
+
+    def constructor_args(self):
```
clundin25, line 178: Should this method be private?

google/auth/external_account.py
```diff
@@ -161,13 +164,25 @@ def info(self):
                 useful for serializing the current credentials so it can deserialized
                 later.
         """
-        config_info = {
-            "type": _EXTERNAL_ACCOUNT_JSON_TYPE,
+        config_info = self.constructor_args()
+        config_info.update(
+            type=_EXTERNAL_ACCOUNT_JSON_TYPE,
+            service_account_impersonation=config_info.pop(
```
clundin25, line 170: This pop should have `None` as a default to avoid a KeyError exception?

google/auth/external_account.py
```diff
@@ -161,13 +164,25 @@ def info(self):
...
+                "service_account_impersonation_options"
+            ),
+        )
+        config_info.pop("scopes")
+        config_info.pop("default_scopes")
```
clundin25, line 175: Same comment re: these pop usages

google/auth/external_account.py
```diff
@@ -255,25 +274,17 @@ def project_number(self):
...
-            scopes=scopes,
-            default_scopes=default_scopes,
-            workforce_pool_user_project=self._workforce_pool_user_project,
-        )
-        if not self.is_workforce_pool:
-            d.pop("workforce_pool_user_project")
-        return self.__class__(**d)
+        kwargs = self.constructor_args()
+        kwargs.update(scopes=scopes, default_scopes=default_scopes)
+        return self.__class__(**kwargs)
```
clundin25, line 287: Thanks for this, this looks a lot better!

tests/test_pluggable.py
```diff
@@ -280,9 +288,17 @@ def test_info_with_credential_source(self):
             "audience": AUDIENCE,
             "subject_token_type": SUBJECT_TOKEN_TYPE,
             "token_url": TOKEN_URL,
+            "token_info_url": TOKEN_INFO_URL,
             "credential_source": self.CREDENTIAL_SOURCE,
         }
 
+    def test_token_info_url(self):
```
clundin25, line 295: Can you add a negative case test?

tests/test_identity_pool.py
```diff
@@ -638,6 +651,13 @@ def test_retrieve_subject_token_file_not_found(self):
 
         assert excinfo.match(r"File './not_found.txt' was not found")
 
+    def test_token_info_url(self):
```
clundin25, line 654: Can you add a negative case test?

tests/test_aws.py
```diff
@@ -954,9 +962,17 @@ def test_info(self):
             "audience": AUDIENCE,
             "subject_token_type": SUBJECT_TOKEN_TYPE,
             "token_url": TOKEN_URL,
+            "token_info_url": TOKEN_INFO_URL,
             "credential_source": self.CREDENTIAL_SOURCE,
         }
 
+    def test_token_info_url(self):
```
clundin25, line 969: Can you add a negative test case?

google/auth/external_account.py
ScruffyProdigy: Done

google/auth/external_account.py
ScruffyProdigy: Done

google/auth/external_account.py
ScruffyProdigy: Done

tests/test_pluggable.py
ScruffyProdigy: Done

tests/test_aws.py
ScruffyProdigy: Done

tests/test_identity_pool.py
ScruffyProdigy: Done

google/auth/external_account.py
```diff
@@ -112,6 +114,7 @@ def __init__(
         self._audience = audience
         self._subject_token_type = subject_token_type
         self._token_url = token_url
+        self._token_info_url = token_info_url
```
lsirac, line 117: Not needed for now since we don't use it, but can you add a TODO to validate the token info URL?

lsirac (approved):

google/auth/external_account.py
ScruffyProdigy: Done

lsirac: Actually lets validate this now

google/auth/external_account.py
```diff
@@ -94,6 +95,7 @@ def __init__(
                 impersonation generateAccessToken URL.
             client_id (Optional[str]): The optional client ID.
             client_secret (Optional[str]): The optional client secret.
+            token_info_url (Optional[str]): The optional STS endpoint for token info.
```
aeitzman, line 98: Should this be token Introspection isntead of info?

google/auth/external_account.py
ScruffyProdigy: Done

google/auth/external_account.py
ScruffyProdigy: Done

BigTailWolf (approved):
LGTM. Very helpful changes. Then the pluggable auth won't be done in a dirty way.

google/auth/external_account.py
```diff
@@ -161,13 +166,25 @@ def info(self):
...
+
+    def _constructor_args(self):
+        args = {
             "audience": self._audience,
             "subject_token_type": self._subject_token_type,
             "token_url": self._token_url,
+            "token_info_url": self._token_info_url,
             "service_account_impersonation_url": self._service_account_impersonation_url,
-            "service_account_impersonation": copy.deepcopy(
+            "service_account_impersonation_options": copy.deepcopy(
```
sai-sunder-s, line 187: why is this called service_account_impersonation sometimes and service_account_impersonation_options sometimes?

tests/test_aws.py
```diff
@@ -954,9 +963,24 @@ def test_info(self):
...
```
sai-sunder-s, line 970: add a test for when custom token info url is provided?

tests/test_external_account.py
```diff
@@ -539,7 +534,6 @@ def test_with_scopes_full_options_propagated(self):
             quota_project_id=self.QUOTA_PROJECT_ID,
             scopes=["email"],
             default_scopes=["default2"],
-            workforce_pool_user_project=None,
```
sai-sunder-s, line 542: why is this removed?

tests/test_aws.py
sai-sunder-s: Similarly in other credential tests as well

sai-sunder-s (approved):

google/auth/external_account.py
ScruffyProdigy: I'm not sure.  I don't mind leaving this as service_account_impersonation_options, but I also didn't want to break anybody's code using our API

tests/test_external_account.py
ScruffyProdigy: Because the changes I've made cause it to be called without this parameter at all, and the default value of `None` gets used instead

google/auth/external_account.py
sai-sunder-s: Looks like this is existing behavior. So I am not blocking for this PR. But would be good to follow up with if we can standardize this. I will let AION SDK team decide

ScruffyProdigy: I think it has to do with the fact that this is called "service_account_impersonation" in the JSON file.  But yeah, consistent behavior here would be nice

tests/test_aws.py
ScruffyProdigy: Done.  Also added a test with an invalid token info url

tests/test_external_account.py
```diff
@@ -73,28 +73,16 @@ def __init__(
...
         credential_source,
-        service_account_impersonation_url=None,
-        service_account_impersonation_options={},
-        client_id=None,
-        client_secret=None,
-        quota_project_id=None,
-        scopes=None,
-        default_scopes=None,
-        workforce_pool_user_project=None,
+        *args,
```
lsirac, line 76: Please add tests for token info url validation. You can copy the way they are done in this file.

ScruffyProdigy: Done.  I did the same for token_url and service_account_impersonation_url

PR #1178, opened 2022-11-07
Cleanup dependencies since only Python ≥ 3.6 is supported

pietrodn:
The README states that only Python ≥ 3.6 is supported.
Therefore, we can remove all the dependencies conditioned to older versions of Python from the `setup.py`.

The Python version constraint can also be simplified.


parthea:
I'm not sure if we can drop support for python 2.7 as it was re-introduced in https://github.com/googleapis/google-auth-library-python/pull/892 after being dropped. According to #892 this was meant to be temporary. I've assigned this PR to @sai-sunder-s to take a look.

pietrodn:
How can you be sure it is actually working? Tests don't even run for Python 2.7, it's very easy to introduce a breaking change. 🤔

arithmetic1728:
Kokoro runs unit tests for python 2.7 (https://github.com/googleapis/google-auth-library-python/blob/main/noxfile.py#L107). If gcloud drops 2.7 then I think we can drop 2.7 support as well. I think @sai-sunder-s has more context on gcloud.

clundin25:
Thank you for your contribution! Closed in favor of https://github.com/googleapis/google-auth-library-python/pull/1301

PR #1182, opened 2022-11-09
fix: updated the lower bound of interactive timeout and fix the kwarg…

BigTailWolf:
* Udpated the lower bound of interactive timeout
* Fix the kwargs invalid syntax


clundin25 (approved):

google/auth/pluggable.py
```diff
@@ -132,7 +132,7 @@ def __init__(
         self._credential_source_executable_output_file = self._credential_source_executable.get(
             "output_file"
         )
-        self._tokeninfo_username = kwargs.get("tokeninfo_username", "")  # dummy value
+        self._tokeninfo_username = kwargs.get("_tokeninfo_username", "")  # dummy value
```
clundin25, line 135: Why was this not caught earlier? Is a unit test missing?

BigTailWolf: Unfortunately there wasn't a case to test that and didn't break the coverage rate.

clundin25: Can you add a test case to catch this?

clundin25 (changes requested):
This change needs accompanied unit tests

google/auth/pluggable.py
BigTailWolf: Test added for variable injection. Also I realized it's not a good idea to put it into constructor.

google/auth/pluggable.py
```diff
@@ -132,7 +132,7 @@ def __init__(
         self._credential_source_executable_output_file = self._credential_source_executable.get(
             "output_file"
         )
-        self._tokeninfo_username = kwargs.get("tokeninfo_username", "")  # dummy value
+        self._tokeninfo_username = "" # dummy value. This variable is only used via injection, not exposed to ctor
```
clundin25, line 135: Nit: I think `None` is a more appropriate placeholder

clundin25 (approved):

google/auth/pluggable.py
BigTailWolf: Then we are going to re-write a bunch of tests since we didn't properly mock the behavior.
This will be updated by a full feature of retrieve token info from the proper endpoint.

google/auth/pluggable.py
```diff
@@ -53,7 +53,7 @@
 EXECUTABLE_TIMEOUT_MILLIS_UPPER_BOUND = 120 * 1000  # 2 minutes
 
 EXECUTABLE_INTERACTIVE_TIMEOUT_MILLIS_DEFAULT = 5 * 60 * 1000  # 5 minutes
-EXECUTABLE_INTERACTIVE_TIMEOUT_MILLIS_LOWER_BOUND = 5 * 60 * 1000  # 5 minutes
+EXECUTABLE_INTERACTIVE_TIMEOUT_MILLIS_LOWER_BOUND = 30 * 1000  # 30 seconds
```
lsirac, line 56: Why do we have a default timeout for interactive mode? It looks like we set it whenever it is not provided. This should not be done - please remove

BigTailWolf: You mean in the pluggable Credential we only enable it when we see it explicitly?
I thought "using the interactive timeout as a sign of enabling interactive mode" is a pure `gcloud` logic which just one of our SDK user. Maybe we can keep the capability in SDK for other users may using interactive all the way?

lsirac: Yes, we should not be setting a default value for an optional field. All pluggable auth credentials would then be considered interactive by your gcloud changes. It would break non-interactive pluggable auth. Needs to be removed and you'll have to pull in google-auth again.

I am not sure what you mean by capability - if they want to use interactive mode with the SDK, they can generate a config that includes an interactive timeout.

BigTailWolf: By capability I mean for example if a 3rd party user wanted use our SDK to develop a client application, maybe a long-live session is keeping our `pluggalbe.Credentials` object and able to switch between interactive and non-interactive status. Our `interactive` member variable is just a status that the object maintains.
A 3rd party be able to using default interactive timeout if they don't want to explicitly have one.

Back to the `gcloud` workforce non-interactive mode. The pluggable auth object will have to be created at initial login and we checked the interactive timeout (gcloud cli logic here). Then non-interactive existing flow won't be considered as interactive. For every time we retrieve it from the JSON BLOB. We only rely on the previous state who persisted the credential. So that won't be a breaking change.

lsirac: The use case you're describing is not very likely. Interactive mode will mostly be used through the CLIs, not standalone with the SDK. This is why we are not adding interactive mode to the other auth libraries. That being said, if they want to use it standalone, they can, but they need to specify an interactive mode configuration - consistent behavior with gcloud. 

I don't see any benefit of adding a default value for this optional field. I think adding it results in a worse experience as now the user's timeout is not transparent to them (e.g. they can set interactive to true on the creds but think the executable timeout is the non interactive timeout in the config since we're now mixing both). It complicates an already complicated flow, adds more documentation burden, and may cause issues in the future with the gcloud integration.

I don't remember your exact gcloud changes for this since they were deleted, but if this isn't breaking that doesn't really change my opinion on this. Relying on `interactive_timeout_millis` to be set in the config and then setting a default value here is error prone. 

If you want to provide a default value for them, I think it should be done during config creation. You can have them specify --interactive instead of an interactive timeout or something like that.

BigTailWolf: Thanks for the clarification of usage. I'll update this one.
On the other hand. This won't block the `gcloud` integration change. I might update both side tomorrow.

BigTailWolf: Updated:
1. Remove the default value for `interactive_timeout_millis`
2. Adding check for `interactive_timeout_millis` when running in interactive mode.
3. Adding associated test.

lsirac: Thanks!

lsirac (approved):

PR #1185, opened 2022-11-15
fix: Allow mtls sts endpoint for external account token urls.

clundin25:


arithmetic1728 (approved):

tests/test_external_account.py
```diff
@@ -67,6 +67,7 @@
 
 VALID_TOKEN_URLS = [
     "https://sts.googleapis.com",
+    "https://sts.mtls.googleapis.com",
```
lsirac, line 70: Probably worth it to add some test cases

google/auth/external_account.py
```diff
@@ -439,6 +439,7 @@ def validate_token_url(token_url, url_type="token"):
         _TOKEN_URL_PATTERNS = [
             "^[^\\.\\s\\/\\\\]+\\.sts\\.googleapis\\.com$",
             "^sts\\.googleapis\\.com$",
+            "^sts\\.mtls\\.googleapis\\.com$",
```
lsirac, line 442: Do we need to support regional mtls endpoints?

clundin25: Let me check if that is the case

tests/test_external_account.py
clundin25: This should get included in this loop here https://github.com/googleapis/google-auth-library-python/blob/690b05509b7679b1d32fe71f36760d0f8c739268/tests/test_external_account.py#L342

google/auth/external_account.py
clundin25: I can confirm regional MTLS endpoints should also be supported. Will follow up.

tests/test_external_account.py
lsirac: Let's add some invalid cases too

lsirac (approved):

google/auth/external_account.py
```diff
@@ -437,11 +437,11 @@ def _initialize_impersonated_credentials(self):
     @staticmethod
     def validate_token_url(token_url, url_type="token"):
         _TOKEN_URL_PATTERNS = [
```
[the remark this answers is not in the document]
BigTailWolf: Not all languages apply to raw strings. Here we are keeping all the languages using the same pattern.

BigTailWolf (approved):
LGTM

tests/test_external_account.py
```diff
@@ -67,23 +67,31 @@
...
     "https://us-east-1.sts.googleapis.com",
+    "https://us-east-1.sts.mtls.googleapis.com",
     "https://US-EAST-1.sts.googleapis.com",
     "https://sts.us-east-1.googleapis.com",
     "https://sts.US-WEST-1.googleapis.com",
```
sai-sunder-s, line 75: add a case for "https://sts.mtls.US-WEST-1.googleapis.com" (if that is allowed)?

clundin25: IIUC MTLS must prefix googleapis.com but I will check

PR #1189, opened 2022-11-17
feat: Introduce a way to provide scopes granted by user

sai-sunder-s:
Users have an option to allow only a subset of the requested scopes. Developers do not have an easy way currently to see which of the scopes were granted. This change introduces a way to expose the granted scopes and also provide granted scopes at time of credential creation (This change will be paired with a [change](https://github.com/googleapis/google-auth-library-python-oauthlib/pull/257/files) in google-auth-oauthlib to provide granted scopes at credential creation time)


google/oauth2/credentials.py
```diff
@@ -117,6 +121,9 @@ def __init__(
                 retrieving downscoped tokens from a token broker.
             enable_reauth_refresh (Optional[bool]): Whether reauth refresh flow
                 should be used. This flag is for gcloud to use only.
+            granted_scopes (Sequence[str]): The scopes that were consented/granted by the user.
```
clundin25, line 124:
```suggestion
            granted_scopes (Optional(Sequence[str])): The scopes that were consented/granted by the user.
```

google/oauth2/credentials.py
```diff
@@ -174,6 +183,11 @@ def scopes(self):
         """Optional[str]: The OAuth 2.0 permission scopes."""
         return self._scopes
 
+    @property
+    def granted_scopes(self):
+        """Optional[str]: The OAuth 2.0 permission scopes that were granted by the user."""
```
clundin25, line 188:
```suggestion
        """Optional[Sequence[str]]: The OAuth 2.0 permission scopes that were granted by the user."""
```

google/oauth2/credentials.py
```diff
@@ -335,10 +351,11 @@ def refresh(self, request):
 
         if scopes and "scope" in grant_response:
             requested_scopes = frozenset(scopes)
-            granted_scopes = frozenset(grant_response["scope"].split())
+            self._granted_scopes = grant_response["scope"].split()
+            granted_scopes = frozenset(self._granted_scopes)
             scopes_requested_but_not_granted = requested_scopes - granted_scopes
             if scopes_requested_but_not_granted:
-                raise exceptions.RefreshError(
+                _LOGGER.warning(
```
clundin25, line 358: Can you give some extra context on why changing the behavior is OK?

Seems reasonable, just want to understand the change.

clundin25 (approved):

google/oauth2/credentials.py
```diff
@@ -79,6 +82,7 @@ def __init__(
         rapt_token=None,
         refresh_handler=None,
         enable_reauth_refresh=False,
+        granted_scopes=None,
```
arithmetic1728, line 85: since granted scopes are not given by the user, I guess we should remove it from the constructor? Inside the class we still keep the _granted_scopes and let the user use the property to access the value.

arithmetic1728 (approved):

PR #1191, opened 2022-11-18
feat: wrap all python built-in exceptions into library excpetions

BigTailWolf:
No longer throw raw exception from python.
All the exceptions are throwing through wrapped exception from `google/auth/exceptions.py` so that any client code can catch those exceptions and knowing it's raised by SDK.


google/auth/_exponential_backoff.py
```diff
@@ -86,7 +88,7 @@ def __iter__(self):
 
     def __next__(self):
         if self._backoff_count >= self._total_attempts:
-            raise StopIteration
+            raise exceptions.StopIteration
```
clundin25, line 91: This doesn't seem necessary. 

I think the for loop catches the `StopIteration`, so it should never leave the loop block

BigTailWolf: I didn't quite understand this part.
Is it in the
```
for _ in retries:
   ...
```

clundin25:
@BigTailWolf

This LGTM. Can you also add automation to prevent unwrapped exceptions from being checked in?

arithmetic1728:
@BigTailWolf  This looks like a big change. Since gcloud is a major customer of google-auth library, could you ask James Wu from gcloud team to see if he has any concerns about it? Thanks!

arithmetic1728:
@BigTailWolf Never mind. Just noticed your new error type inherits the original error type, so I think this should be backward compatible. LGTM

arithmetic1728 (approved):

PR #1203, opened 2022-12-14
fix: allow get_project_id to take a request

ScruffyProdigy:


google/auth/external_account_authorized_user.py
```diff
@@ -222,7 +222,7 @@ def can_refresh(self):
             (self._refresh_token, self._token_url, self._client_id, self._client_secret)
         )
 
-    def get_project_id(self):
+    def get_project_id(self, request=None):
```
arithmetic1728, line 225: this parameter is not used, why add it?

Do you mind adding an `Args` and `Returns` comment like the `to_json` method below.

lsirac: gcloud passes it and currently crashes

lsirac (approved):

google/auth/external_account_authorized_user.py
ScruffyProdigy: I've added Args and Returns, and used them to explain why we don't use request

google/auth/external_account_authorized_user.py
```diff
@@ -222,12 +222,20 @@ def can_refresh(self):
...
         the workforce_pool_user_project.
 
         When not determinable, None is returned.
+
+        Args:
+            request (google.auth.transport.requests.Request): Request object.
+                Unused here, but passed from _default.default().
+
+        Return:
+          None — project ID is not determinable for this credential type.
```
arithmetic1728, line 237: I think it should be `str: project ID is not determinable for this credential type so it returns None`, not `None - `. 

This line or next line somehow fails the kokoro test:
```
from google.auth import external_account_authorized_user
E     File "/tmpfs/src/github/google-auth-library-python/google/auth/external_account_authorized_user.py", line 237
E   SyntaxError: Non-ASCII character '\xe2' in file /tmpfs/src/github/google-auth-library-python/google/auth/external_account_authorized_user.py on line 238, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

BigTailWolf (approved):
LGTM

google/auth/external_account_authorized_user.py
ScruffyProdigy: Done

arithmetic1728 (approved):

arithmetic1728:
Will merge after #1210 is submitted

PR #1208, opened 2023-01-05
fix: Make OAUTH2.0 client resistant to string type 'expires_in' responses from non-compliant services

### LINKED ISSUE #1207: expires_in appears to be a string when returned from the token server
When we reach this line of code, we get "TypeError: unsupported type for timedelta seconds component: str"

https://github.com/googleapis/google-auth-library-python/blob/6c1297c4d69ba40a8b9392775c17411253fcd73b/google/oauth2/_client.py#L123

It would appear that the response from the token server has expires_in as a string:

Request:
```
https://securetoken.googleapis.com/v1/token?key=AI... 
{
  'grant_type': 
  'refresh_token', 
  'client_id': '76...', 
  'client_secret': '...', 
  'refresh_token': 'AO...'
}
```

Response:
```
{
  'access_token': 'ey...', 
  **'expires_in': '3600'**, 
  'token_type': 'Bearer', 
  'refresh_token': 'AO...', 
  'id_token': 'ey...', 
  'user_id': 'Xs...', 
  'project_id': '23...'
}
```

Seems this code should expect a string and not an integer.

clundin25:
This fixes https://github.com/googleapis/google-auth-library-python/issues/1207.

### DIFF
--- a/google/oauth2/_client.py
+++ b/google/oauth2/_client.py
@@ -120,6 +120,11 @@ def _parse_expiry(response_data):
     expires_in = response_data.get("expires_in", None)
 
     if expires_in is not None:
+        # Some services do not respect the OAUTH2.0 RFC and send expires_in as a
+        # JSON String.
+        if isinstance(expires_in, str):
+            expires_in = int(expires_in)
+
         return _helpers.utcnow() + datetime.timedelta(seconds=expires_in)
     else:
         return None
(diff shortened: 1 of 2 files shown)


wangyutongg (approved):
LGTM
Just curious: When do we generate the secrets.tar.enc file and how to generate it?

clundin25:
@wangyutongg Thanks!

Will follow up on that offline.

arithmetic1728 (approved):

PR #1217, opened 2023-01-25
print more info

hphucnp:


clundin25:
@hphucnp This does not seem to be a legitimate PR. Please open an issue to discuss what problem you are facing, and we can discuss a potential solution. Thanks!

PR #1220, opened 2023-02-01
fix: Remove 3PI config url validation

BigTailWolf:
Remove the url validation for TPC support.
Failed configuration will be on user's own risk in future.


clundin25:
nit: Using the `$ git revert $COMMIT` could be an easier way to revert this code. This will also result in self-documenting commits.

BigTailWolf:
> 

I considered that. But I found those validations are not added the same time. And very hard to find all these separate changes.

BigTailWolf:
> nit: Using the `$ git revert $COMMIT` could be an easier way to revert this code. This will also result in self-documenting commits.

I considered that. But I found those validations are not added the same time. And very hard to find all these separate changes.

BigTailWolf:
> Please update the README to mention that we do not validate the URLs in the config for Workload/Workforce

Hi @lsirac , and for the `is_workforce_pool`, we also have regex matching there, that would fail TPC user's check.

lsirac:
> 

Yes, please update the regex as well

BigTailWolf:
> > > Please update the README to mention that we do not validate the URLs in the config for Workload/Workforce
> > 
> > 
> > Hi @lsirac , and for the `is_workforce_pool`, we also have regex matching there, that would fail TPC user's check.
> 
> ~Yes, please update the regex as well~
> 
> Let's keep it for now, and circle back on this

Good. Maybe need backlog a "loose the pattern check for is_workforce_pool"

lsirac (dismissed):

docs/user-guide.rst
```diff
@@ -823,6 +823,16 @@ in the target project. Using `ImpersonatedCredentials` will allow the source_cre
 to assume the identity of a target_principal that does have access.
 
 
+Security considerations on configuration External and Impersonated credentials
```
lsirac, line 826: lets add this in the workload and workforce section like Alex did

BigTailWolf:
@arithmetic1728 may I get an approval?

docs/user-guide.rst
```diff
@@ -793,6 +803,13 @@ Cloud resources from an OIDC or SAML provider.
     https://cloud.google.com/iam/docs/workforce-identity-federation#workforce-pools-user-project
 
 
+Note that this library does not perform any validation on the token_url,
```
lsirac, line 806: Missing Security considerations header?

BigTailWolf: No, this section already have header a few lines above

clundin25, line 806: Why duplicate this snippet?

BigTailWolf: To have this in both workload and workforce section since they are already separated.

clundin25 (approved):

PR #1243, opened 2023-03-02
fix: call gcloud config get project to get project for user cred

### LINKED ISSUE #1239: Password prompt opened by `google.auth.default()` when credentials have expired
When you have a reauthentication policy set in Google Cloud and your crendentials have expired, calling `google.auth.default()` will open a `Please enter your password:` password prompt, instead if raising a `DefaultCredentialsError` as I believe would be expected.

The password prompt interrupts the flow of the program, with no way to catch or handle it actively.

If credentials are _revoked_ (instead of expired), you don't get the password prompt, and instead get an exception as expected.

For example, when creating a `bigquery.Client`, its `__init__` calls `super(Client, self).__init__()` which in turn calls `google.auth.default()`, which in turn [calls the external program `gcloud config config-helper --format json` in a `subprocess`](https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/_cloud_sdk.py#L108), which finally opens the prompt.

Full traceback after starting it with `python -m pdb ...` and hitting Ctrl-C from when the prompt is open:

```
  ".../code_under_test.py", line 90, in _bqclient
    client = bigquery.Client(
  File ".../.venv/lib/python3.10/site-packages/google/cloud/bigquery/client.py", line 244, in __init__
    super(Client, self).__init__(
  File ".../.venv/lib/python3.10/site-packages/google/cloud/client/__init__.py", line 321, in __init__
    Client.__init__(
  File ".../.venv/lib/python3.10/site-packages/google/cloud/client/__init__.py", line 178, in __init__
    credentials, _ = google.auth.default(scopes=scopes)
  File ".../.venv/lib/python3.10/site-packages/google/auth/_default.py", line 620, in default
    credentials, project_id = checker()
  File ".../.venv/lib/python3.10/site-packages/google/auth/_default.py", line 614, in <lambda>
    lambda: _get_gcloud_sdk_credentials(quota_project_id=quota_project_id),
  File ".../.venv/lib/python3.10/site-packages/google/auth/_default.py", line 205, in _get_gcloud_sdk_credentials
    project_id = _cloud_sdk.get_project_id()
  File ".../.venv/lib/python3.10/site-packages/google/auth/_cloud_sdk.py", line 108, in get_project_id
    output = _run_subprocess_ignore_stderr((command,) + _CLOUD_SDK_CONFIG_COMMAND)
  File ".../.venv/lib/python3.10/site-packages/google/auth/_cloud_sdk.py", line 90, in _run_subprocess_ignore_stderr
    output = subprocess.check_output(command, stderr=devnull)
  ...
```

#### Environment details

  - OS: macOS 13.2.1
  - Python version: 3.10.8
  - pip version: 23.0.1
  - `google-auth` version: 2.16.1

#### Steps to reproduce

  1. Set a [Reauthentication policy](https://admin.google.com/ac/security/reauth/admin-tools) in Google Cloud.
  2. Authenticate using `gcloud auth login --update-adc`.
  3. Let your access token expire.
  4. Now, `python -c "import google.auth; google.auth.default()"` will prompt you with `Please enter your password:`.

If instead of letting your token expire, you actively revoke it with `gcloud auth application-default revoke` and `gcloud auth revoke`, then you get the following exception raised:

```
  File "<string>", line 1, in <module>
  File ".../.venv/lib/python3.10/site-packages/google/auth/_default.py", line 653, in default
    raise exceptions.DefaultCredentialsError(_HELP_MESSAGE)
google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see https://cloud.google.com/docs/authentication/getting-started
```

Also, for some reason the password prompt is **not** opened if you capture `stdin` like `subprocess.check_output(["gcloud", "config", "config-helper", "--format", "json"], stdin=subprocess.PIPE)`. This has the interesting effect that if you run code in `pytest`(that captures `stdin`), you don't get the password prompt, but if you run it with `pytest -s` you do.

arithmetic1728:
Fix #1239 

`google.auth.default()` returns the ADC cred and project id. If the cred is user cred, currently we use `gcloud config config-helper --format json` to obtain the json of all properties and find the `project` property from the json. The problem is that this command refreshes the credentials after expiry, and if reauth is enabled, it will prompt for user interaction. This is not the right behavior. We should use `gcloud config get project` instead because it simply return the project id and not do any refresh.


google/auth/_cloud_sdk.py
```diff
@@ -105,18 +105,14 @@ def get_project_id():
...
-        return None
+        project = _run_subprocess_ignore_stderr(
+            (command,) + _CLOUD_SDK_CONFIG_GET_PROJECT_COMMAND
+        )
 
-    try:
-        return configuration["configuration"]["properties"]["core"]["project"]
-    except KeyError:
+        # Turn bytes into a string and remove "\n"
+        project = _helpers.from_bytes(project).strip()
```
clundin25, line 113: Do we need to do any parsing? This command outputs this for me:

```
>>> subprocess.run(['gcloud', 'config', 'get', 'project'])
Your active configuration is: [test]
stellar-day-254222
```

Seems that the first line may cause some issues, or is it okay?

clundin25 (approved):

google/auth/_cloud_sdk.py
arithmetic1728: The first line won't show up because it writes to stderr

```
import subprocess

output = subprocess.check_output(('gcloud', 'config', 'get', 'project'))
print(output)
print('---------------')
print(_cloud_sdk.get_project_id())
```

result is:
```
Your active configuration is: [foo]
b'sijun-auth\n'
---------------
sijun-auth
```

clundin25: Missed that! Thanks!

PR #1254, opened 2023-03-20
fix: Python: Remove aws url validation

BigTailWolf:


clundin25 (approved):

lsirac (approved):

arithmetic1728:
@BigTailWolf  There is a lint error: `google/auth/aws.py:50:1: F401 'six.moves.urllib.parse.urlparse' imported but unused`
After removing it you can run `python -m nox -s lint` to verify the problem is fixed.

PR #1255, opened 2023-03-23
feat: add httpx transport; add compute_engine async

bkanuka:
Opening a draft PR to gauge interest in adding an [httpx](https://www.python-httpx.org/) transport.

This is used internally for a project that otherwise relied on httpx (instead of aiohttp) for async http requests.  Using this httpx transport means that the response from an `AuthorisedSession` is an `httpx.Respose` like the other responses in our code.

Frankly, this a copy/paste job from `_aiohttp_requests.py` with relevant parts changed.

---

`compute_engine._credentials_async.py` is also added for async Compute Engine credentials.  Arguably, this should be discussed separately, but was created while developing the httpx transport.

---

There is more work to be done before this could merge, but please discuss and provide feedback.


clundin25:
Super cool! I will give this a thorough read later.

Two high level questions I have for you.

1. What made you decide to use HTTPX instead of aiohttp?
2. Is it feasible to support both with the same code?
    - I will be digging into this later.

Lawouach:
Oh this is exactly what I'd need. httpx is such a brilliant library, it would be awesome of the Google auth client could natively support it.

chalmerlowe:
Added an issue to track this post migration.

See Issue #1908

PR #1258, opened 2023-03-24
feat: experimental service account iam endpoint flow for id token

arithmetic1728:
For service account credentials, implement a new ID token flow which uses iam.generateIdToken endpoint. This feature is currently experimental since iam endpoint doesn't support setAzpToEmail option yet. The ID token generated by this new flow will have a different azp claim (it uses id instead of email)

Design doc: go/googleapis-auth-id-token-iam-for-tpc

Example usage:
```
from google.oauth2 import service_account

cred = service_account.IDTokenCredentials.from_service_account_file(
    '/usr/local/google/home/sijunliu/wks/creds/nondca/srv.json',
    target_audience = "https://pubsub.googleapis.com",
)

# Trigger IAM endpoint flow.
# Note this feature is still experimental since setAzpToEmail is
# not supported yet
cred = cred._with_use_iam_endpoint(True)
req = google.auth.transport.requests.Request()

cred.refresh(req)
print(cred.token)
```


google/oauth2/service_account.py
```diff
@@ -639,6 +640,31 @@ def with_target_audience(self, target_audience):
             quota_project_id=self.quota_project_id,
         )
 
+    def _with_use_iam_endpoint(self, use_iam_endpoint):
```
clundin25, line 643: Is there a scenario where the IAM endpoint needs to be overridden? E.g. mtls, TPC?

google/oauth2/service_account.py
```diff
@@ -692,14 +718,49 @@ def _make_authorization_grant_assertion(self):
...
+        1. First we create a self signed jwt with
+        https://www.googleapis.com/auth/iam being the scope.
+
+        2. Next we use the self signed jwt as the access token, and make a POST
+        request to IAM generateIdToken endpoint. The request body is:
+            {
+                "audience": self._target_audience,
+                "includeEmail": "true"
+            }
+        TODO: add "set_azp_to_email": "true" once it's ready from server side.
```
clundin25, line 735: nit: Would be good to track this TODO in a bug or GitHub issue and link to it here

clundin25 (approved):

google/oauth2/service_account.py
arithmetic1728: done

Issue #927, opened 2021-12-13
Ungraceful error `None could not be converted to unicode`


k0ral:
#### Environment details

- OS: Ubuntu 18.04.5 LTS, running docker image based on Ubuntu 20.04
- Python version: 3.7.1
- pip version: 19.0.3
- `google-auth` version: 1.35.0
- `google-cloud-bigtable` version: 1.7.0
- `google-cloud-storage` version: 1.42.3
- `google-cloud-pubsub` version: 2.8.0

#### Error

```
Traceback (most recent call last):
  File "/opt/venv/lib/python3.7/site-packages/grpc/_plugin_wrapping.py", line 90, in __call__
    context, _AuthMetadataPluginCallback(callback_state, callback))
  File "/opt/venv/lib/python3.7/site-packages/google/auth/transport/grpc.py", line 101, in __call__
    callback(self._get_authorization_headers(context), None)
  File "/opt/venv/lib/python3.7/site-packages/google/auth/transport/grpc.py", line 88, in _get_authorization_headers
    self._request, context.method_name, context.service_url, headers
  File "/opt/venv/lib/python3.7/site-packages/google/auth/credentials.py", line 134, in before_request
    self.apply(headers)
  File "/opt/venv/lib/python3.7/site-packages/google/auth/credentials.py", line 110, in apply
    _helpers.from_bytes(token or self.token)
  File "/opt/venv/lib/python3.7/site-packages/google/auth/_helpers.py", line 130, in from_bytes
    raise ValueError("{0!r} could not be converted to unicode".format(value))
ValueError: None could not be converted to unicode
```

This is essentially the same error as described in [this StackOverflow question](https://stackoverflow.com/questions/70172317/intermittent-authentication-error-when-posting-to-a-pubsub-topic).

#### Steps to reproduce

I'm afraid we haven't iterated enough to isolate the minimal reproducing piece of code, so at the moment we can't provide a detailed steps to reproduce with a minimal, complete, verifiable example, but we know that:
- the error happens at the start of a microservice that interacts with multiple GCP components (namely: PubSub, Storage and Bigtable)
- the error seems to have no consequence whatsoever: the microservice correctly starts and runs after that error 

It would be nice if `google-auth` library could report a more graceful/meaningful error, then we can figure out by ourselves what we did wrong.

syrkuit:
**Reproduction script:**

```#! /usr/bin/python3
import logging
import time
import google.cloud.pubsub_v1 as pubsub

[...]

PR #1267, opened 2023-04-04
fix: Do not create new JWT credentials if they make the same claims as the existing.

clundin25:
This resolves race condition reported in https://github.com/googleapis/google-auth-library-python/issues/927.


arithmetic1728 (approved):

tests/oauth2/test_service_account.py
```diff
@@ -253,6 +253,24 @@ def test__create_self_signed_jwt_always_use_jwt_access_with_audience(self, jwt):
...
+            SIGNER,
+            self.SERVICE_ACCOUNT_EMAIL,
+            self.TOKEN_URI,
+            default_scopes=["bar", "foo"],
+            always_use_jwt_access=True,
+        )
+
+        audience = "https://pubsub.googleapis.com"
+        credentials._create_self_signed_jwt(audience)
+        credentials._jwt_credentials._audience = audience
```
sai-sunder-s, line 270: why do we set the audience here explicitly? User code would not be doing that right?

clundin25: In this case I'm mocking what the real object would do, which is store the audience property

sai-sunder-s (approved):

PR #1282, opened 2023-05-03
feat: implement token introspection flow

BigTailWolf:
Implement token introspection flow that when the cred has a valid client auth and token_info_url.
For any token, the cred can introspect token info.

For external_account_id, we implement it at `external_account.Credentials` layer.
If the service_account_email is not available, we are using token_info_username which retrieved from token introspection.


google/auth/external_account.py
```diff
@@ -425,6 +484,23 @@ def _initialize_impersonated_credentials(self):
...
 
+    def _token_info_username(self):
+        if not self._token_info_introspection:
+            try:
+                self.token_info_introspection(None)
+            except exceptions.GoogleAuthError:
+                # A failure of introspection shouldn't break the flow
+                pass
+
+        # One runnig of token_info_introspection() is not guaranteed we have a
```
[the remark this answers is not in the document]
BigTailWolf: Not necessarily empty, it may remain the value of last successful token introspection

google/auth/external_account.py
```diff
@@ -382,6 +397,50 @@ def with_token_uri(self, token_uri):
...
+        if not self._token_info_url:
+            raise exceptions.InvalidResource("Missing token_info_url")
+
+        if not self.token:
+            raise exceptions.InvalidOperation("Missing token for introspection")
+
+        oauth_introspection_client = utils.TokenIntrospectionClient(
+            self._token_info_introspection, self._client_auth
+        )
+        if not request:
```
clundin25, line 433: I only see this pattern in one other place. Why fallback to requests instead of throwing an invalid argument exception?

google/auth/external_account.py
```diff
@@ -40,6 +40,7 @@
 from google.auth import credentials
 from google.auth import exceptions
 from google.auth import impersonated_credentials
+from google.auth.transport import requests
```
clundin25, line 43: I'm wondering if this breaks the intention of the transport library. The code should be agnostic to the underlying implementation.

google/auth/external_account.py
```diff
@@ -382,6 +397,50 @@ def with_token_uri(self, token_uri):
         kwargs.update(token_url=token_uri)
         return self.__class__(**kwargs)
 
+    def token_info_introspection(self, request):
+        """Introspection of token info.
```
clundin25, line 401: nit: This is redundant with the function name. IMO it would be useful to describe _what_ introspection means, because it is not clear without further digging.

google/auth/external_account.py
```diff
@@ -425,6 +484,23 @@ def _initialize_impersonated_credentials(self):
...
 
+    def _token_info_username(self):
+        if not self._token_info_introspection:
+            try:
+                self.token_info_introspection(None)
+            except exceptions.GoogleAuthError:
+                # A failure of introspection shouldn't break the flow.
+                pass
+
+        # One runnig of token_info_introspection() is not guaranteed we have a
```
clundin25, line 495: nit: Typo s/runnig/running/

google/auth/external_account.py
```diff
@@ -425,6 +484,23 @@ def _initialize_impersonated_credentials(self):
...
+        # valid inspection result cached. A failed introspection may results
+        # the cache unchanged and an empty cache still there.
+        return (
+            None
+            if not self._token_info_introspection
+            else self._token_info_introspection.get("username")
+        )
```
clundin25, line 502:
```suggestion
        return self._token_info_introspection.get("username", None)
```

Assuming this is just a regular dict

BigTailWolf: `self._token_info_introspection` may remain `None` and won't have a `get` method.

clundin25: Gotcha. Why use None instead of an empty dict? Do you need to make any decisions based on it being an optional type?

BigTailWolf: Good raise. Seems reasonable to initialize it as an empty dict.

google/auth/external_account.py
BigTailWolf: Oh the choose here is because I want the default headers of `google.auth.transport.requests.Request`.
For the body we are reassigning it anyway in the `TokenIntrospectionClient`.
Actually I can use a raw request and add all the headers to it if we are going to remove this dependency.

clundin25: @arithmetic1728 thoughts? Should `external_account` code pull in `requests` as a dependency?

google/auth/external_account.py
BigTailWolf: Actually as we have valid client auth (client id + client secret) and a sts issued token, we are OK to introspect the token. I think it doesn't matter that we have a request or not.
For here I just exposed one parameter in `token_info_introspection` method to keep consistency with other public methods.
I'd like to know your idea about this part.

google/auth/external_account.py
```diff
@@ -210,6 +214,17 @@ def service_account_email(self):
                 return url[start_index:end_index]
         return None
 
+    @property
+    def external_account_id(self):
+        """Return the external account id.
+        When service account impersonation is used, it will return the service account email.
+        When no service account impersonation is used, it will return the principal.
```
lsirac, line 221: This is a little misleading as it can only be used by gcloud -- I don't think we should expose it.

BigTailWolf: It's a public method, if it can be used by `gcloud`, it can be called by any other user client code. Since our library is just a dependency of `gcloud`

google/auth/external_account.py
BigTailWolf: Oh, even with a raw `requests.Request` I'm still having `requests` as a dependency.

google/auth/external_account.py
lsirac: Yes, exactly. It shouldn't be public. There is currently no way for someone to use these credentials and call token info. Only gcloud can do that.

google/auth/external_account.py
clundin25: If I understood correctly, IMO I would not introduce the request parameter unless it is absolutely needed. 

You can always add a param without a breaking change, but they can never be removed.

google/auth/external_account.py
BigTailWolf: I'm little bit confused. If there is no public method, then how the library expose this to `gcloud`?

google/auth/external_account.py
BigTailWolf: Now the case is:
In `oauth2.utils.TokenIntrospectionClient`, which will actually make the request there is no dependency on `requests` package. All the requests are passed in through the `introspection(request, token)` method.
If in `external_account.Credentials` we don't have dependency on the requests, then we must let it be provided by the caller.
Current implementation in `gcloud` is to having a `google.auth.request` session. And then that session's request calling the `IntrospectionClient` which implements in-house.

google/auth/external_account.py
BigTailWolf: I know it is possible technically, but seems not a good choice.
"Note that accessing a private method from outside the class is generally not recommended, as it violates the principle of encapsulation and can lead to unexpected behavior. It's better to rely on the public interface of the class to interact with its functionality."

Besides, don't we want to expose a public method for SDK users to have an insight of the token?
What I read from our token introspection proposal is:

```
Today, [GCP Security Token Service ](https://cloud.google.com/iam/docs/reference/sts/rest)(sts.googleapis.com) issues GCP access tokens. The contents of these tokens are opaque to clients. This means that the client can’t know anything about the content or structure of the token itself.
```

lsirac: Token introspection doesn't work without basic auth. SDK users can't call token info when using external account credentials. We can't expose a method that says this returns the principal because that won't work yet.

You can expose this on external account authorized user credentials, which are generated through gcloud auth application-default login and will be able to call token info.

BigTailWolf: By basic auth I think you mean `client_id` and `client_secret`. How about the `external_account` throw an exception if there is no basic auth? And with a valid auth. We are offering this kind of operation only when there is an access token already.

BigTailWolf: Expose `token_info_introspection` to be a public method when used with basic auth can return metadata of a token.
Throw an exception if there is:
1. No `token_info_url` configured
2. No basic auth with the external_account
3. No active token for the credential object

Expose `external_account_id` to be a property method. It will call visit cached value of `token_info_introspection` `username` filed if no `service_account_email` and throwing no error in any cases.

google/auth/external_account.py
arithmetic1728: Yes it's fine. `requests` is the most popular transport that most people, our client lib, and gcloud use. `urllib3` is low level and not convenient (btw `requests` wraps `urllib3` to make it easier to use). I doubt if anyone want to use `urllib3`.

BigTailWolf: In current iteration, I already get rid of the `requests` :)

google/auth/external_account.py
```diff
@@ -389,6 +391,48 @@ def with_token_uri(self, token_uri):
         kwargs.update(token_url=token_uri)
         return self.__class__(**kwargs)
 
+    def introspect_token(self, request):
+        """Introspection of token info.
```
lsirac, line 395: "Introspection of token info" doesn't quite make sense -- let's replace with something else

google/auth/external_account.py
```diff
@@ -389,6 +391,48 @@ def with_token_uri(self, token_uri):
...
+
+        If the credential object has a token_info_url provided, we can
```
lsirac, line 397: Token_info_url should always be available. not sure if it is worth calling out here. we can just throw an error when it isn't and document that error

google/auth/external_account.py
```diff
@@ -389,6 +391,48 @@ def with_token_uri(self, token_uri):
         kwargs.update(token_url=token_uri)
         return self.__class__(**kwargs)
 
+    def introspect_token(self, request):
```
lsirac, line 394: For external account, this should be _introspect_token

google/auth/external_account.py
```diff
@@ -389,6 +391,48 @@ def with_token_uri(self, token_uri):
...
+                If no client auth given.
+            google.auth.exceptions.InvalidOperation:
+                If no token provided.
+            google.auth.exceptions.UserAccessTokenError:
+                If the credentials are invalid or expired.
+            google.auth.exceptions.TransportError:
+                If an error is encountered while calling the token
+                introspection endpoint.
+        """
+        # Token info introspection following RFC7662 standard:
```
lsirac, line 417: Maybe move this above?

google/auth/external_account.py
```diff
@@ -389,6 +391,48 @@ def with_token_uri(self, token_uri):
...
+        if not self._token_info_url:
+            raise exceptions.InvalidResource("Missing token_info_url")
+
+        if not self._client_auth:
+            raise exceptions.ClientCertError(
+                "Token introspection needs client basic auth"
+            )
+
+        if not self.token:
+            raise exceptions.InvalidOperation("Missing token for introspection")
```
lsirac, line 429: Maybe mention calling refresh before calling this function in the error message

google/auth/external_account_authorized_user.py
```diff
@@ -282,6 +282,36 @@ def refresh(self, request):
         if "refresh_token" in response_data:
             self._refresh_token = response_data["refresh_token"]
 
+    def introspect_token(self, request):
+        """Introspection of token info.
```
lsirac, line 286: same comments apply here

google/auth/pluggable.py
```diff
@@ -283,18 +283,20 @@ def revoke(self, request):
         response = json.loads(result.stdout.decode("utf-8"))
         self._validate_revoke_response(response)
 
-    @property
-    def external_account_id(self):
-        """Returns the external account identifier.
-
-        When service account impersonation is used the identifier is the service
-        account email.
+    def external_account_id(self, request):
```
lsirac, line 286: I don't think we should expose this here. Should be in external account, and private?

google/auth/external_account.py
BigTailWolf: In the workflow of `introspect_token`, `token_info_url` is always needed and will throw an error.
Do you mean we need to throw error if it's not configured in constructor?
Won't that lost backward compatibility?

BigTailWolf: For `external_account_authorized_user.Credentials` I didn't see the validate of `self._token_info_url` in constructor.
Also I saw the default parameter is `None`. Is that mean some `external_account_authorized_user.Credentials` can skip this field?

google/auth/external_account.py
BigTailWolf: Changed.

google/auth/external_account.py
BigTailWolf: Delete. Don't need to mention here. Because the implementation is in the Client and in the Client's comment it's already mentioned.

google/auth/external_account.py
lsirac: Yes, looks like when we implemented it to be initializable with minimal info (even though gcloud will always populate token info).

google/auth/pluggable.py
lsirac: Not sure I am understanding correctly as this is just calling introspect_token which can only be invoked from gcloud when using these credentials

BigTailWolf: > Not sure I am understanding correctly as this is just calling introspect_token which can only be invoked from gcloud when using these credentials

I think my understanding is different. We are open to any client (gcloud or 3rd party) that has basic auth to introspect the token they acquired. The introspect of token can be done for external_account type. And for pluggable auth specifically used for gcloud. Retrieving external account id is just one usage of that and we do that in the `pluggable_auth.Crentials` privately.

PR #1286, opened 2023-05-06
feat: universe domain support for service account

arithmetic1728:
Support universe_domain field from service account json file. The default value is "googleapis.com".

If the value is not default, then 
(1) for access token, we will always use self signed jwt flow. Note that domain wide delegation is not supported, so if subject is provided we will throw exceptions.
(2) for id token, we will always use iamcredentials.generateIdToken endpoint. Note that iam.serviceAccountTokenCreator role is required to call this endpoint.


BigTailWolf (approved):

google/oauth2/service_account.py
```diff
@@ -156,6 +159,9 @@ def __init__(
                 the JWT assertion used in the authorization grant.
             always_use_jwt_access (Optional[bool]): Whether self signed JWT should
                 be always used.
+            universe_domain (Optional[str]): The universe domain. The default
```
clundin25, line 162: Should this just be a str? What happens if `universe_domain` is `None`?

google/oauth2/service_account.py
```diff
@@ -262,20 +273,28 @@ def requires_scopes(self):
...
-    def with_scopes(self, scopes, default_scopes=None):
-        return self.__class__(
+    def _make_copy(self):
+        cred = self.__class__(
             self._signer,
             service_account_email=self._service_account_email,
-            scopes=scopes,
-            default_scopes=default_scopes,
+            scopes=self._scopes,
+            default_scopes=self._default_scopes,
```
clundin25, line 281: I think these objects need to be copied. Otherwise multiple objects will share pointers to these sequences

google/oauth2/service_account.py
```diff
@@ -418,6 +389,18 @@ def _make_authorization_grant_assertion(self):
...
     @_helpers.copy_docstring(credentials.Credentials)
     def refresh(self, request):
+        if (
+            self._universe_domain != _DEFAULT_UNIVERSE_DOMAIN
+            and not self._jwt_credentials
+        ):
+            raise exceptions.RefreshError(
+                "self._jwt_credentials is missing for non-default universe domain"
+            )
+        if self._universe_domain != _DEFAULT_UNIVERSE_DOMAIN and self._subject:
```
clundin25, line 399: nit: I think collapsing `self._universe_domain != _DEFAULT_UNIVERSE_DOMAIN ` into a single if statement, and then nesting the following conditions might improve readability.

google/oauth2/service_account.py
```diff
@@ -556,6 +540,11 @@ def __init__(
             additional_claims (Mapping[str, str]): Any additional claims for
                 the JWT assertion used in the authorization grant.
             quota_project_id (Optional[str]): The project ID used for quota and billing.
+            universe_domain (Optional[str]): The universe domain. The default
```
clundin25, line 543: Similar Q regarding `None` type

google/oauth2/service_account.py
arithmetic1728: Changed to str. Updated the logic: if `universe_domain` is `None` or "", it will be set to the default one.

google/oauth2/service_account.py
arithmetic1728: I initially wrote the way as you said, but it seems the coverage util has some issue, it complains some part of the code missed coverage even though I have the tests. Changing to the current way solved the issue, lol

google/oauth2/service_account.py
arithmetic1728: yes coped them now

google/oauth2/service_account.py
arithmetic1728: same treatment as above

google/oauth2/service_account.py
```diff
@@ -156,6 +159,9 @@ def __init__(
                 the JWT assertion used in the authorization grant.
             always_use_jwt_access (Optional[bool]): Whether self signed JWT should
                 be always used.
+            universe_domain (str): The universe domain. The default
+                universe domain is googleapis.com. If this value is the default
+                value, then self signed jwt will be used for token refresh.
```
TimurSadykov, line 164: nit: Replace "If this value is the default value" with "in case of default value" or "for default".. otherwise reads a bitt awkward, too many "the value/default" in a row )

google/oauth2/service_account.py
```diff
@@ -556,6 +543,11 @@ def __init__(
             additional_claims (Mapping[str, str]): Any additional claims for
                 the JWT assertion used in the authorization grant.
             quota_project_id (Optional[str]): The project ID used for quota and billing.
+            universe_domain (str): The universe domain. The default
+                universe domain is googleapis.com. If this value is the default
```
TimurSadykov, line 547: Same suggestion to maybe simplify: "If this value is the default value, then" -> "For default value" or such

google/oauth2/service_account.py
clundin25: What a pain lol

clundin25 (approved):

google/oauth2/service_account.py
```diff
@@ -287,18 +309,9 @@ def with_always_use_jwt_access(self, always_use_jwt_access):
...
-            default_scopes=self._default_scopes,
-            token_uri=self._token_uri,
-            subject=self._subject,
-            project_id=self._project_id,
-            quota_project_id=self._quota_project_id,
-            additional_claims=self._additional_claims.copy(),
-            always_use_jwt_access=always_use_jwt_access,
-        )
+        cred = self._make_copy()
+        cred._always_use_jwt_access = always_use_jwt_access
```
TimurSadykov, line 313: if the input value is false and original credential has TPC universe - do we want to throw? 

Otherwise could backfire for some existing code moved into TPC env.

google/oauth2/service_account.py
```diff
@@ -569,6 +561,14 @@ def __init__(
         self._quota_project_id = quota_project_id
         self._use_iam_endpoint = False
 
+        if not universe_domain:
+            self._universe_domain = _DEFAULT_UNIVERSE_DOMAIN
+        else:
+            self._universe_domain = universe_domain
+
+        if universe_domain != _DEFAULT_UNIVERSE_DOMAIN:
+            self._use_iam_endpoint = True
```
TimurSadykov, line 570: In the with_iam_endpoint method we probably want to check if universe is not a default before overriding in case of False input.

TimurSadykov (changes requested):

google/oauth2/service_account.py
arithmetic1728: done

google/oauth2/service_account.py
arithmetic1728: done

google/oauth2/service_account.py
arithmetic1728: added logic to throw

google/oauth2/service_account.py
arithmetic1728: added logic to throw because if universe is not default only the iam endpoint id token flow work

TimurSadykov (approved):
LGTM

PR #1292, opened 2023-05-17
Fix: Missing close session

vadimka123:


arithmetic1728:
fixed in #1315 , thank you @vadimka123 for your contribution!

PR #1296, opened 2023-05-17
feat: expose `universe_domain` for external account creds

BigTailWolf:


google/auth/external_account.py
```diff
@@ -458,6 +465,7 @@ def from_info(cls, info, **kwargs):
             credential_source=info.get("credential_source"),
             quota_project_id=info.get("quota_project_id"),
             workforce_pool_user_project=info.get("workforce_pool_user_project"),
+            universe_domain=info.get("universe", _DEFAULT_UNIVERSE_DOMAIN),
```
arithmetic1728, line 468: info.get("universe_domain",...) ?

google/auth/external_account.py
```diff
@@ -125,6 +130,7 @@ def __init__(
         self._scopes = scopes
         self._default_scopes = default_scopes
         self._workforce_pool_user_project = workforce_pool_user_project
+        self._universe_domain = universe_domain or _DEFAULT_UNIVERSE_DOMAIN
```
clundin25, line 133: I don't think you need the `or` portion of this statement, since `universe_domain` uses `_DEFAULT_UNIVERSE_DOMAIN` as a default value

BigTailWolf: It's a protection of user explicitly set `None` or empty string. I asked @arithmetic1728 about the implementation of 
https://github.com/googleapis/google-auth-library-python/blob/main/google/oauth2/service_account.py#L182-L185
is the same purpose.

clundin25: Hmm. Wouldn't throwing an invalid argument exception be better in that case? That seems like something users should be alerted to, rather then falling back to the default universe.

BigTailWolf: Maybe keep a consistency to `service_account.Credentials`? Or we can consider to change that part together as well.

clundin25: Let's let @arithmetic1728 chime in, but the change can be deferred to a separate PR.

arithmetic1728: When users provide None or empty string I think their intent is to use the default value so this treatment looks fine to me.  But if everyone wants to throw an exception, I don't mind changing it as well for consistency.

lsirac, line 133: should be self.universe_domain - right? @sai-sunder-s

tests/test_external_account.py
```diff
@@ -158,6 +159,7 @@ def make_credentials(
             quota_project_id=quota_project_id,
             scopes=scopes,
             default_scopes=default_scopes,
+            universe_domain=universe_domain,
```
lsirac, line 162: test retrieving it as well

google/auth/external_account.py
arithmetic1728: we made it private for service account for now because adc file doesn't have this field yet

lsirac: Any reason to keep it private here? It defaults to GDU. also the change was made for BYOID configs

lsirac: @viacheslav-rostovtsev

arithmetic1728: I think the reason is universe_domain is not finalized, which needs more discussion 
@TimurSadykov has more context

lsirac: ok lets keep it private

lsirac (approved):

PR #1298, opened 2023-05-21
feat: add metrics (part 1)

arithmetic1728:
This PR:
(1) list the metrics values needed
(2) add the `_metric_header_for_usage` method to the base credential class, which is used by the `before_request` method to add the metrics header for token usage. Children credentials classes can override this method for token usage metrics.

internal doc: go/googleapis-auth-metric-design


google/auth/metrics.py
```diff
@@ -0,0 +1,165 @@
...
+# Credential type
+CRED_TYPE_USER = "cred-type/u"
+CRED_TYPE_SA_ASSERTION = "cred-type/sa"
+CRED_TYPE_SA_JWT = "cred-type/jwt"
+CRED_TYPE_SA_MDS = "cred-type/mds"
+CRED_TYPE_SA_IMPERSONATE = "cred-type/imp"
+
+
+# Versions
+def auth_lib_version():
```
aeitzman, line 42: Any reason not to have these two as a single method that returns both language and library version? We should always be including both on the x-goog-api-client header as far as I know

google/auth/metrics.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def add_metric_header(headers, metric_header_value):
+    """Add x-goog-api-client header with the given value.
+
+    Args:
+        headers (Mapping[str, str]): The headers to which we will add the
+            metric header.
+        metric_header_value (Optional[str]): If value is None, do nothing;
+            if headers already has a x-goog-api-client header, append the value
```
aeitzman, line 156: If the x-goog-api-client header is already present shouldn there be a check for the individual parts of the header? I assume if the header already exists the language version section would already be there for example

arithmetic1728: For the token refresh requests, there is no x-goog-api-client header so we are fine.

For api requests where we add the token, client lib adds `x-goog-api-client` header. From auth lib side, we just append `cred-type/xx` to the header. 
- We don't need to send lang version because the header already has it
- We don't send auth version per discussion with concord team, because this will dramatically increase the size of concord table. For example, if currently we have a row for pubsub api call and the request_count field value is 100, it's likely these api calls use 5 different auth versions, 20 for each version, then the new table will need 5 rows, one row for each auth version).

tests/test_metrics.py
```diff
@@ -0,0 +1,77 @@
...
+    assert headers == {"x-goog-api-client": "foo bar"}
+
+    headers = {}
+    metrics.add_metric_header(headers, "bar")
+    assert headers == {"x-goog-api-client": "bar"}
+
+
+def test_versions():
+    assert metrics.python_and_auth_lib_version() == "gl-python/{} auth/{}".format(
+        platform.python_version(), version.__version__
```
wangyutongg, line 39: Can we mock the version here?

arithmetic1728: I don't find a good way to do so. I updated the test to save the `version.__version__` value, then assign it a fake value, do the test, then assign it back, so we can have some sort of "mock".

wangyutongg (approved):
LGTM

aeitzman (approved):

PR #1316, opened 2023-05-31
fix: misc fixes

arithmetic1728:
BEGIN_COMMIT_OVERRIDE
fix: fix "AttributeError: 'str' object has no attribute 'get'" 
docs: replacing abc.com with example.com
END_COMMIT_OVERRIDE

Fix reauth logic so it can handle token response where error response is a string instead of dict.  Similar issue see #857

### DIFF
--- a/google/oauth2/reauth.py
+++ b/google/oauth2/reauth.py
@@ -34,7 +34,6 @@ Those steps are:
 
 import sys
 
-import six
 from six.moves import range
 
 from google.auth import exceptions
@@ -336,7 +335,7 @@ def refresh_grant(
         request, token_uri, body, headers=metrics_header
     )
 
-    if not response_status_ok and isinstance(response_data, six.string_types):
+    if not response_status_ok and isinstance(response_data, str):
         raise exceptions.RefreshError(response_data, retryable=False)
 
     if (
--- a/google/oauth2/reauth.py
+++ b/google/oauth2/reauth.py
@@ -34,6 +34,7 @@ Those steps are:
 
 import sys
 
+import six
 from six.moves import range
 
 from google.auth import exceptions
@@ -334,6 +335,10 @@ def refresh_grant(
     response_status_ok, response_data, retryable_error = _client._token_endpoint_request_no_throw(
         request, token_uri, body, headers=metrics_header
     )
+
+    if not response_status_ok and isinstance(response_data, six.string_types):
+        raise exceptions.RefreshError(response_data, retryable=False)
+
     if (
         not response_status_ok
         and response_data.get("error") == _REAUTH_NEEDED_ERROR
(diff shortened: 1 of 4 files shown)


clundin25 (approved):

parthea (approved):

google/oauth2/reauth.py
```diff
@@ -334,6 +334,10 @@ def refresh_grant(
     response_status_ok, response_data, retryable_error = _client._token_endpoint_request_no_throw(
         request, token_uri, body, headers=metrics_header
     )
+
+    if not response_status_ok and isinstance(response_data, str):
```
glabdulhadi, line 338: enable_reauth_refresh flag is ignored here. is that fine? shouldn't enable_reauth_refresh flag be considered for whether exception needs to be raised or not?

tests/oauth2/test_reauth.py
```diff
@@ -326,6 +326,26 @@ def test_refresh_grant_failed(mock_metrics_header_value):
...
+        with pytest.raises(exceptions.RefreshError) as excinfo:
+            reauth.refresh_grant(
+                MOCK_REQUEST,
+                "token_uri",
+                "refresh_token",
+                "client_id",
+                "client_secret",
+                scopes=["foo", "bar"],
+                rapt_token="rapt_token",
+                enable_reauth_refresh=True,
```
glabdulhadi, line 343: in continuation from above comment.
In other words, i would like to understand what should be the expected behaviour of the method for the below test case?

`    with mock.patch(
        "google.oauth2._client._token_endpoint_request_no_throw"
    ) as mock_token_request:
        mock_token_request.return_value = (False, _REAUTH_NEEDED_ERROR, False)
        with pytest.raises(exceptions.RefreshError) as excinfo:
            reauth.refresh_grant(
                MOCK_REQUEST,
                "token_uri",
                "refresh_token",
                "client_id",
                "client_secret",
                scopes=["foo", "bar"],
                rapt_token="rapt_token",
                enable_reauth_refresh=True,
           )
           `
whether RefreshError exception be expected or whether reauth refresh flow (i.e., 354-364) is expected since enable_reauth_refresh=True and reponse_data=_REAUTH_NEEDED_ERROR?

glabdulhadi: @clundin25 , please let me know your thoughts on this.

arithmetic1728: I believe in case of reauth needed error, the oauth2 token endpoint always returns a json, in other words, something like `mock_token_request.return_value = (False, {"error":_REAUTH_NEEDED_ERROR, ...}, False)`. The server won't return `mock_token_request.return_value = (False, _REAUTH_NEEDED_ERROR, False)` as described in the test case you mentioned.

clundin25:
@arithmetic1728 perhaps `enable_reauth_refresh=True` should just be an early return in the function?

tests/oauth2/test_reauth.py
glabdulhadi: ok. then agree.

PR #1326, opened 2023-06-07
feat: Add public API load_credentials_from_dict

### LINKED ISSUE #1313: publicly expose _default._load_credentials_from_info
at the moment the public interface is just google.auth.default() and load_credentials_from_file, but if we have a data dictionary already due to it being loaded from a non file source it would be nice to pass it directly to `load_credentials_from_info` https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/_default.py#L133

the load from file function is just a super thin wrapper to load from info. 

afaics load_credentials_from_info is the most compatible across all the auth forms that gcp supports. (user default app, service account, external).

clundin25:
Add public API load_credentials_from_dict to allow creating a default credential object from a dictionary.

This resolves https://github.com/googleapis/google-auth-library-python/issues/1313.


tests/test__default.py
```diff
@@ -188,6 +188,20 @@ def test_load_credentials_from_missing_file():
     assert excinfo.match(r"not found")
 
 
+def test_load_credentials_from_dict_non_dict_object():
```
clundin25, line 191: `load_credentials_from_file` has extensive test coverage, and `load_credentials_from_dict` only varies by the input type.

google/auth/_default.py
```diff
@@ -130,6 +130,50 @@ def load_credentials_from_file(
...
+        google.auth.exceptions.DefaultCredentialsError: if the file is in the
+            wrong format or is missing.
+    """
+    if not isinstance(info, dict):
+        raise exceptions.DefaultCredentialsError(
+            "info object was of type {} but dict type was expected.".format(type(info))
+        )
+
+    return _load_credentials_from_info(
+        "dict object", info, scopes, default_scopes, quota_project_id, request
```
clundin25, line 173: The `filename` param is only used for logging purposes. I can make a more invasive change if that is desirable.

BigTailWolf (approved):

arithmetic1728 (approved):

PR #1327, opened 2023-06-08
fix: expiry in compute_engine.IDTokenCredentials

### LINKED ISSUE #1323: Wrong timezones in compute_engine.IDTokenCredentials expiry
The `expiry` of compute_engine.IDTokenCredentials is in the local timezone, but it's then compared to utc. This means that an expired token may be used. Expiry of all other credential types are correctly in UTC.

#### Environment details

  - OS: Linux
  - Python version: 3.11
  - pip version: 23.1.2
  - `google-auth` version: 2.19.1
  - Tested on GKE 1.24.12-gke.500 with Workload Identity, but AFAICT it would also fail anywhere on GCP

#### Steps to reproduce

Run on a GCE VM (or a GKE pod).

Configure Python to use some timezone far from UTC, eg `export TZ=America/New_York`.

```py
import google.auth.compute_engine.credentials

r = google.auth.transport.requests.Request()
creds = google.auth.compute_engine.credentials.IDTokenCredentials(r, target_audience="foo", use_metadata_identity_endpoint=True)
creds.refresh(r)

print(f"expiry: {creds.expiry}")
print(f"expired: {creds.expired}")
```

Here, `expired` incorrectly reports false, because it compares the local `expiry` with `utcnow`.

Another failure mode is in timezones with a positive offset (eg `Europe/Prague`), where the token will be treated as not-expired even after it actually expired.

All other credential types use utc for everything, so they don't have the problem. Even the compute engine OAuth2 credentials in the same file (ie just `Credentials`, not `IDTokenCredentials`).

Should be a very simple fix, to use UTC datetime everywhere.

juzna:
fixes #1323

### DIFF
--- a/google/auth/compute_engine/credentials.py
+++ b/google/auth/compute_engine/credentials.py
@@ -389,7 +389,7 @@ class IDTokenCredentials(
             six.raise_from(new_exc, caught_exc)
 
         _, payload, _, _ = jwt._unverified_decode(id_token)
-        return id_token, datetime.datetime.fromtimestamp(payload["exp"])
+        return id_token, datetime.datetime.utcfromtimestamp(payload["exp"])
 
     def refresh(self, request):
         """Refreshes the ID token.
--- a/tests/compute_engine/test_credentials.py
+++ b/tests/compute_engine/test_credentials.py
@@ -770,7 +770,7 @@ class TestIDTokenCredentials(object):
         }
 
         assert cred.token == SAMPLE_ID_TOKEN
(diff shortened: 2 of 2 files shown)


clundin25 (approved):

clundin25:
@juzna 

Can you update the test case as well? 

https://github.com/googleapis/google-auth-library-python/blob/main/tests/compute_engine/test_credentials.py#L773

arithmetic1728 (approved):

juzna:
Tests updated & commit ammended.

clundin25:
Thank you for your contribution !

PR #1332, opened 2023-06-12
feat: Add framework for BYOID metrics headers

aeitzman:


google/auth/identity_pool.py
```diff
@@ -216,6 +216,16 @@ def _parse_token_data(
             )
         return token
 
+    def _create_default_metrics_options(self):
+        metrics_options = super(Credentials, self)._create_default_metrics_options()
+        try:
+            if self._credential_source.get("file"):
```
clundin25, line 222:
```suggestion
            if self._credential_source.get("file", None):
```

What is the context behind the use of a `try` block?

google/auth/metrics.py
```diff
@@ -123,6 +125,15 @@ def reauth_continue():
     return "{} {}".format(python_and_auth_lib_version(), REQUEST_TYPE_REAUTH_CONTINUE)
 
 
+# x-goog-api-client header value for BYOID calls to the sts endpoint.
+# Example: "gl-python/3.7 auth/1.1 google-byoid-sdk source/aws sa-impersonation/True sa-impersonation/True"
+def byoid_metrics_header(metrics_options):
```
clundin25, line 130: Does this function have a unit test? It would be useful to add one that makes sure a whitespace is not appended

google/auth/identity_pool.py
aeitzman: Added the try because this method runs before the full __init__ completes for identity pool credentials since it gets called by the super().__init__. If the try isn't there, the error thrown comes from the metrics method instead of being the expected error. Will add a comment

aeitzman: example of test that would fail as well - https://github.com/googleapis/google-auth-library-python/blob/49d0f9eeb7887e0cc1f3b2686cf03a2c2dc4a1db/tests/test__default.py#LL535C1-L544C6

arithmetic1728:
Will the byoid metric header go to any of the oauth2.googleapis.com/token, IAM token endpoint, VM metadata server endpoints? Will the header go to chemist, or show up in cloud_apis concord table?

google/auth/identity_pool.py
clundin25: Is it possible to refactor the constructor to enforce this invariant before this function is called?

This approach makes the function very complex and tough to refactor. 

Perhaps this function should be refactored to be a property, so it is guaranteed that the object is fully instantiated?

google/auth/external_account.py
```diff
@@ -85,6 +86,7 @@ def __init__(
         default_scopes=None,
         workforce_pool_user_project=None,
         universe_domain=_DEFAULT_UNIVERSE_DOMAIN,
+        metrics_options=None,
```
arithmetic1728, line 89: Maybe not expose `metric_options` since it's internal details?

google/auth/external_account.py
```diff
@@ -140,6 +143,11 @@ def __init__(
             self._client_auth = None
         self._sts_client = sts.Client(self._token_url, self._client_auth)
 
+        # Set metrics options or create default if options were not provided.
+        self._metrics_options = (
+            metrics_options or self._create_default_metrics_options()
```
arithmetic1728, line 148: why not just use the default one?

aeitzman: We need to be able to pass the metrics options through the constructor because of the way impersonated credentials are handled in this library. If impersonation is used, the original credential (defined by the ADC file) is created. That credential knows that it is using impersonation because it has an impersonation_url passed in the constructor. Then, the original credential creates an impersonated credential object and passes a source credential to the impersonated credential. The source credential is just a copy of the original credential but with all the impersonation information removed. The source credential is the object that actually makes the call to sts, so we need to have some for it to know that the call to sts is for service account impersonation by passing the metrics object from the original credential. If we just use the create default, source metric has no way to know its being used for impersonation.

google/auth/external_account.py
aeitzman: Needs to be exposed either in the constructor or via an setter, see reply on comment below.

google/auth/external_account.py
arithmetic1728: I see. Thank you for the detailed explanation!

google/auth/external_account.py
arithmetic1728: I think a private setter might work better since metrics are internal details.

arithmetic1728:
> > Will the byoid metric header go to any of the oauth2.googleapis.com/token, IAM token endpoint, VM metadata server endpoints? Will the header go to chemist, or show up in cloud_apis concord table?
> 
> No, it is only going to the sts endpoint (and that is all that is planned for the forseeable future) which does not go through chemist or show up in the cloud_apis table.

Great. Thank you for the confirmation!

google/auth/external_account.py
aeitzman: Ok, just to make sure I understand, you're suggesting a private setter that is used here like its public (the original external account credential object would be calling it on the source credential after the source credential is created to pass to the impersonated object) but just marked private so users know they shouldn't be crafting/setting their own metrics options

arithmetic1728: yes, that's right. You can introduce a `def _with_metric_options(options)` method here to return a copy of the object but with the given options as `self._metric_options` (i.e. first create a copy of the object, then set `self._metric_options = options`). In the constructor, you can then simply set `self._metric_options = self._create_default_metrics_options()`.

google/auth/identity_pool.py
BigTailWolf: @clundin25 Why do we prefer an explicit `None` in your suggestion? I thought by default it's `None` already.

google/auth/metrics.py
aeitzman: added test

google/auth/external_account.py
```diff
@@ -404,7 +420,6 @@ def _initialize_impersonated_credentials(self):
             google.auth.exceptions.RefreshError: If the generateAccessToken
                 endpoint returned an error.
         """
-        # Return copy of instance with no service account impersonation.
```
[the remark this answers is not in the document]
aeitzman: no, I did not, added back

google/auth/identity_pool.py
clundin25: My thought was that the try/catch could be replaced by an if statement. Now that I see it's a workaround for the interaction with the constructor.

I would prefer seeing the logic refactored to not need such a try/catch, but my initial code suggestion is no longer relevant.

aeitzman: Its a bit difficult either way, _metrics_options is needed in the parent class constructor because the impersonated credential object is created there, and the _initialize_impersonated_credentials method needs to get the _metrics_options value, so even if we switch it to a property, it will run before the constructor finishes. Could move the super().__init__ to be after the credential source validation for identity pool but that seems a more significant change to fix this

BigTailWolf (approved):
LGTM

arithmetic1728 (approved):

arithmetic1728:
please run `python -m nox -s blacken` to fix the lint errors (and use `python -m nox -s lint` to verify), also it's complaining line 280 in `tests/test_identity_pool.py` missed test coverage.

google/auth/external_account.py
```diff
@@ -432,6 +448,19 @@ def _initialize_impersonated_credentials(self):
             ),
         )
 
+    def _create_default_metrics_options(self):
```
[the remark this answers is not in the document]
aeitzman: There is no guidance in the style guide for method naming besides that they should be descriptive and they should be lower_with_underscores

google/auth/external_account.py
```diff
@@ -432,6 +448,19 @@ def _initialize_impersonated_credentials(self):
...
+        metrics_options = {}
+        if self._service_account_impersonation_url:
```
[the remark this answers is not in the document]
aeitzman: I did them as strings instead of the bool to keep the header value consistently lowercase "true" across all languages, if I did that check, it would print "True" for the header value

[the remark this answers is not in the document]
lsirac: You can still simplify this using the same logic

lsirac (approved):

PR #1334, opened 2023-06-13
feat: adding meta header for trust boundary

BigTailWolf:
For most base class `google.auth.credentials.Credentials`, expose a field `_trust_boundary` default to `None`.
During the `before_request`. Apply the value to `X-Identity-Trust-Boundary` field if it's not `None`.

For planned credential types (service account and 3PI credentials) `google.oauth2.service_account.Credentials` and `google.auth.external_account.Credentials`. We inject an empty string value in constructor in order to have that field to be exposed during `before_request`.


google/auth/external_account.py
```diff
@@ -85,6 +85,7 @@ def __init__(
         default_scopes=None,
         workforce_pool_user_project=None,
         universe_domain=_DEFAULT_UNIVERSE_DOMAIN,
+        trust_boundary=None,
```
aeitzman, line 88: Why expose this constructor option right now if we aren't using it anywhere?

google/auth/external_account.py
```diff
@@ -131,6 +133,7 @@ def __init__(
         self._default_scopes = default_scopes
         self._workforce_pool_user_project = workforce_pool_user_project
         self._universe_domain = universe_domain or _DEFAULT_UNIVERSE_DOMAIN
+        self._trust_boundary = ""  # expose a placeholder trust boundary value.
```
aeitzman, line 136: Why set the empty string here instead of None? This is going to add an X-Identity-Trust-Boundary="" header to every request right? Is there some value to that right now?

google/auth/credentials.py
```diff
@@ -127,6 +130,8 @@ def apply(self, headers, token=None):
         headers["authorization"] = "Bearer {}".format(
             _helpers.from_bytes(token or self.token)
         )
+        if self._trust_boundary is not None:
+            headers["X-Identity-Trust-Boundary"] = self._trust_boundary
```
aeitzman, line 134: nit: should this all be lowercase? Probably doesn't matter but normally these x-xyz header keys seem to be lowercase in this lib

BigTailWolf: > nit: should this all be lowercase? Probably doesn't matter but normally these x-xyz header keys seem to be lowercase in this lib

I asked Annan for confirmation. She explicitly mentioned dashed camel case. IDK whether that matters for her. Maybe should ask whether can I change that to meet our consistency.

google/auth/external_account.py
BigTailWolf: Right now we wanted to inject `service_account.Credentials` and `external_account.Credentials` with empty string header. That would help check the coverage of requests once that GFE starts logging the header.
And for very base `credentails.Credentials` where it defined, the value is `None`. So that when we applying header, our `is not None` won't apply if it's other credentials inherit from that.

google/auth/credentials.py
aeitzman: That's fine if its what the service expects, just wanted to double check

google/auth/external_account.py
BigTailWolf: True. It can be either now or later when we really do the logic.

google/auth/external_account.py
aeitzman: What do you mean by the coverage of requests? Can't we just send some test requests when the backend starts logging? It seems a bit weird to me to start adding an empty header that isn't doing anything to every single request.

BigTailWolf: Our design of exposing those meta field in library is not guaranteed to be all the incoming requests to GFE. There will be some 3rd party not using SDK library. We don't know these percentile and won't know it until we really doing the data gathering. The phase 1 is we can do this data exposing then GFE can gathering data.

aeitzman: Reading the design doc, I'm still confused on why this is necessary, is the goal to see which languages support the trust boundary header? If so this doesn't seem like its sending useful information since the library doesn't actually support it right now for service accounts or external accounts (since customers can't actually use it, it only sends dummy data). Also, what information is this giving us that just checking the library version from the header and validating that the version used is equal or greater to the version this gets merged into wouldn't?

BigTailWolf: The actual purpose is to check is there a gap that "user upgraded to the latest library version and still sending requests without header for the type of credentials". (Should be none since it's a sanity check though)
Sounds not that necessary, but sure will be the starting point of how the library expose trust boundary.

lsirac: re: "user upgraded to the latest library version and still sending requests without header for the type of credentials"

I am not following either. how can that happen? checking the library version is enough for what you're wanting to know?

BigTailWolf: > re: "user upgraded to the latest library version and still sending requests without header for the type of credentials"
> 
> I am not following either. how can that happen? checking the library version is enough for what you're wanting to know?

Actually it shouldn't happen. And with the header gathering from GFE should proof that not happening.

aeitzman: Can't we also prove that with unit tests + 1 manual test?

BigTailWolf: > Can't we also prove that with unit tests + 1 manual test?

That will be covered by test. But not actually what received from GFE. I think the uber TLs expected to see the real data.

arithmetic1728 (approved):

aeitzman (approved):

lsirac (approved):

PR #1343, opened 2023-07-06
fix: Check if user access token is valid before refresh

clundin25:
- fix: Check if user access token credential is valid before refreshing.

### DIFF
--- a/google/oauth2/credentials.py
+++ b/google/oauth2/credentials.py
@@ -534,5 +534,6 @@ class UserAccessTokenCredentials(credentials.CredentialsWithQuotaProject):
 
     @_helpers.copy_docstring(credentials.Credentials)
     def before_request(self, request, method, url, headers):
-        self.refresh(request)
+        if not self.valid:
+            self.refresh(request)
         self.apply(headers)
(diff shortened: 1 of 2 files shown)


arithmetic1728 (approved):

arithmetic1728 (commented):
this does not work, because this type of credential just gets token by running gcloud auth print-access-token, it doesn't even set the token property

clundin25:
> this does not work, because this type of credential just gets token by running gcloud auth print-access-token, it doesn't even set the token property

Ah I see. So it should always refresh, as there is no way to track the expiry? We could consider querying the token info endpoint when the token is assigned, but I am not sure if that will save much performance

clundin25:
Or add a gcloud API to query the token expiry

arithmetic1728:
I think we might want to deprecate this credential type, and ask users to use user credentials instead. I added this credential type for mTLS usage because at that time user credential won't work. It's a google-auth specific credential.

clundin25:
Makes sense @arithmetic1728, I think we should go ahead and deprecate then!

PR #1347, opened 2023-07-11
fix: import pyopenssl.py from urllib3 to google-auth

arithmetic1728:
pyopenssl support will be removed from urllib3 library from v2.1.0, see https://github.com/urllib3/urllib3/issues/2680, which will completely break our mTLS feature. There we import the pyopenssl.py module to google-auth so google-auth can still use it with newer versions of urllib3. Most of the code is copied unchanged, with the following modifications:
(1) This module requires cryptography >= 2.1.0 and pyopenssl >= 0.14. Since google-auth requires pyopenssl>=20.0.0, cryptography>=38.0.3, we removed the version check logic in the module import code, and removed "\_validate_dependencies_met" method.
(2) copied to_bytes method here from https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/util.py
(3) removed unused extract_from_urllib3 method
(4) reordered the module imports
(5) replaced util.ssl_.PROTOCOL_TLS with ssl.PROTOCOL_TLS_CLIENT (same for PROTOCOL_TLS_CLIENT as well) since google-auth requires python >= 3.6 and python's ssl module has these constants defined.


clundin25:
This LGTM. Maybe you can add some more details on this code's origin? E.g. a commit hash or version tag, so we can more easily trace the fork point.

IMO we should hold off on merging this until v2.2.x of urllib3 is released. If we are lucky by that point we may not even need this change.

arithmetic1728:
> This LGTM. Maybe you can add some more details on this code's origin? E.g. a commit hash or version tag, so we can more easily trace the fork point.
> 
> IMO we should hold off on merging this until v2.2.x of urllib3 is released. If we are lucky by that point we may not even need this change.

Yep, added the version tag to the code. Let's hold off until v2.1.0 of urllib3 is released. v2.1.0 will remove the support for sure, so most likely we will need to merge the change at that time.

PR #1348, opened 2023-07-11
feat: Provide a way to read project from environment instead of checking credential

sai-sunder-s:
For external account credentials, the project id is obtained through an API call and that involves creating a token first. If project id is set already in environment, that is the value that is going to be returned by google.auth.default and the value read through the API is discarded.
Updating logic so that if env var is set, then dont make an API call for project.

### DIFF
--- a/google/auth/_default.py
+++ b/google/auth/_default.py
@@ -106,6 +106,8 @@ def load_credentials_from_file(
             for a workload identity pool resource (external account credentials).
             If not specified, then it will use a
             google.auth.transport.requests.Request client to make requests.
+        use_project_from_env (bool): If True, use the project ID from the
+            environment variables.    
 
     Returns:
         Tuple[google.auth.credentials.Credentials, Optional[str]]: Loaded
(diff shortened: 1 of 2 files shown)


google/auth/_default.py
```diff
@@ -79,7 +79,12 @@ def _warn_about_problematic_credentials(credentials):
 
 
 def load_credentials_from_file(
-    filename, scopes=None, default_scopes=None, quota_project_id=None, request=None
+    filename,
+    scopes=None,
+    default_scopes=None,
+    quota_project_id=None,
+    request=None,
+    use_project_from_env=False,
```
arithmetic1728, line 87: add some comments for this new parameter below line 108?

arithmetic1728 (approved):

BigTailWolf (approved):
LGTM

google/auth/_default.py
lsirac, line 87: I'm a little confused -- why are we passing this new var down? Don't we just need to skip the call in default if explicit_project_id is set?

sai-sunder-s: This method returns both the project and the cred itself. We still need the cred. We only do not want to use the cred to get project id, if already set in env.

PR #1349, opened 2023-07-12
fix: Skip checking projectid on cred if env var is set

sai-sunder-s:
For external account credentials, the project id is obtained through an API call and that involves creating a token first. If project id is set already in environment, that is the value that is going to be returned by google.auth.default and the value read through the API is discarded.
Updating logic so that if env var is set, then dont make an API call for project.


arithmetic1728 (approved):

tests/test__default.py
```diff
@@ -1030,6 +1030,35 @@ def test_default_environ_external_credentials_identity_pool_impersonated(
...
+
+    assert isinstance(credentials, identity_pool.Credentials)
+    assert not credentials.is_user
+    assert not credentials.is_workforce_pool
+    assert project_id == project_from_env
+    assert credentials.scopes == ["https://www.google.com/calendar/feeds"]
+
+    # The credential.get_project_id should have been used only in _get_external_account_credentials
+    assert get_project_id.call_count == 1
+
```
BigTailWolf, line 1062: Can we also add a test that getting from default value instead of `envioronment_vars.PROJECT`?

sai-sunder-s: Do you mean the scenario where env var is not set, so the project is obtained from credential.get_project_id?

BigTailWolf: No, not that scenario. I mean the scenario we don't have env var `PROJECT` set, but we have `LEGACY_PROJECT` set.

sai-sunder-s: Added another test for legacy project

lsirac (approved):

PR #1354, opened 2023-07-20
chore: Remove support for Python 3.6

clundin25:
No urgency to merge this, but I do want to use some features from Python 3.7.

3.6 has been EOL for sometime, and the core library has dropped support https://github.com/googleapis/python-api-core.


parthea:
Python 3.7 is also unsupported in the python community. https://devguide.python.org/versions/#unsupported-versions . I'd recommend dropping both at the same time. Could we consider raising a deprecation warning in the next month notifying users that support for both 3.6 and 3.7 will be dropped on a specific date?  We have a similar proposal in `google-cloud-python` which will be sent for review soon: https://github.com/googleapis/google-cloud-python/issues/11437

Only users who are running the latest version of `google-auth` on runtime versions which are unsupported would receive the warnings, and the warning should only appear once per session.

clundin25:
Historically I think `gcloud` has been the main reason to keep older versions, but now that we only support Python3 it might be easier to move faster.

Could we write a document that describes what versions we support and for how long? That might scale better than adding deprecations. Something akin to Go's version policy.

arithmetic1728 (approved):

BigTailWolf (approved):

PR #1368, opened 2023-08-11
feat: Add optional non blocking refresh for sync auth code

clundin25:


google/auth/credentials.py
```diff
@@ -83,6 +85,21 @@ def valid(self):
...
+    @property
+    def token_state(self):
+        if not self.valid or self.expired:
+            return TokenState.INVALID
+
+        # Credentials that can't expire are always treated as fresh.
+        if not self.expiry:
+            return TokenState.FRESH
+
+        refresh_window = _helpers.utcnow() >= (self.expiry - _helpers.REFRESH_THRESHOLD)
```
sai-sunder-s, line 97: rename this variable to something like is_fresh?

google/auth/credentials.py
```diff
@@ -157,11 +174,22 @@ def before_request(self, request, method, url, headers):
...
-        if not self.valid:
+
+        if self.token_state == TokenState.STALE:
+            if self._use_non_blocking_refresh:
+                self._refresh_worker.start_refresh(self, request)
+            else:
+                self.refresh(request)
+
+        if self.token_state == TokenState.INVALID:
             self.refresh(request)
```
sai-sunder-s, line 185: Multiple requests may end up doing this refresh right?
Can we have a lock on this sync refresh?

clundin25: The code is using a bounded queue which should reduce duplicate work. There is a risk that there is a small amount of duplicate work, if two threads queue work at the same time. The queue itself is thread safe + has locks.

I believe using the queue to reduce duplicate work is acceptable, and that multiple threads will perform the refresh occasionally. 

There should be no data corruption due to this

BigTailWolf (approved):

arithmetic1728 (approved):

google/oauth2/credentials.py
```diff
@@ -175,8 +177,11 @@ def __setstate__(self, d):
         self._enable_reauth_refresh = d.get("_enable_reauth_refresh")
         self._trust_boundary = d.get("_trust_boundary")
         self._universe_domain = d.get("_universe_domain")
+        self._token_state = d.get("_token_state")
```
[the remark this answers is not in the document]
clundin25: Great catch! I will delete this line

google/auth/_refresh_worker.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+            return
+        try:
+            cred.refresh(request)
+        except Exception as err:  # pragma: NO COVER
+            # This condition is covered by the unit test test_refresh_error but
+            # it can be flaky due to the scheduler.
+            _LOGGER.error(f"Background refresh failed due to: {err}")
+            if not self._error_queue.full():
+                self._error_queue.put_nowait(err)
```
westarle, line 158: with the code above, you can just stuff the error into a field.

google/auth/credentials.py
```diff
@@ -422,3 +466,9 @@ def signer(self):
         # pylint: disable=missing-raises-doc
         # (pylint doesn't recognize that this is abstract)
         raise NotImplementedError("Signer must be implemented.")
+
+
+class TokenState(Enum):
```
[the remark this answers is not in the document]
clundin25: Added a doc string for this class.

google/auth/credentials.py
```diff
@@ -70,10 +75,7 @@ def expired(self):
         if not self.expiry:
             return False
 
-        # Remove some threshold from expiry to err on the side of reporting
-        # expiration early so that we avoid the 401-refresh-retry loop.
-        skewed_expiry = self.expiry - _helpers.REFRESH_THRESHOLD
-        return _helpers.utcnow() >= skewed_expiry
+        return _helpers.utcnow() >= self.expiry
```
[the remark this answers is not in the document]
clundin25: Good point. This was a risky change. I updated the code to maintain backwards compatibility. 

I did not add a warning, but the documentation will suggest using token state.

google/auth/impersonated_credentials.py
```diff
@@ -259,7 +259,11 @@ def _update_token(self, request):
         """
 
         # Refresh our source credentials if it is not valid.
-        if not self._source_credentials.valid:
+        if (
+            not self._source_credentials.valid
```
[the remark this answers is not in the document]
clundin25: Good point, I've made this change

google/auth/_refresh_worker.py
```diff
@@ -0,0 +1,162 @@
...
+
+        Args:
+            cred: A credentials object.
+            request: A request object.
+        """
+        if cred is None or request is None:
+            raise e.InvalidValue(
+                "Unable to start refresh. cred and request must be valid and instantiated objects."
+            )
+
```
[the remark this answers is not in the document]
clundin25: Refactored the solution to be similar. I kept an error queue as a mechanism to communicate errors back to the RefreshManager. 

Let me know if there is a way that is aligned with the above.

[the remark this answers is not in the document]
clundin25: Actually, I see that I can add a field to the worker. I will make that change.

google/auth/_refresh_worker.py
```diff
@@ -62,101 +52,53 @@ def start_refresh(self, cred, request):
...
-      Drop all errors in the error queue.
-      """
-        try:
-            while not self._error_queue.empty():
-                _ = self._error_queue.get_nowait()
-        # This condition is unlikely but there is a possibility that an
-        # error gets queued between the empty and get calls
-        except queue.Empty:  # pragma: NO COVER
-            pass
+            if self._need_worker():  # pragma: NO COVER
```
westarle, line 61: nit: within a mutex, it's usually better to be explicit about accessing data and not call out to methods.
* this helps restrict the scope of the mutex
* this ensures the methods aren't called outside of a mutex

These methods are simple enough (and similar to the check above) so I'd just inline them.

google/auth/credentials.py
```diff
@@ -469,6 +483,13 @@ def signer(self):
 
 
 class TokenState(Enum):
+    """
+    Tracks the state of a token.
+    FRESH: The token is not expired and can be used normally.
```
westarle, line 488: The token is valid.  It is not expired or close to expired, or the token has no expiry.  To make it mutually exclusive to STALE.

google/auth/_refresh_worker.py
```diff
@@ -13,39 +13,29 @@
...
-
     def __init__(self):
-        """Initializes the worker."""
+        """Initializes the manager."""
 
-        self._refresh_queue = queue.Queue(self.MAX_REFRESH_QUEUE_SIZE)
-        # Bound the error queue to avoid infinitely growing the heap.
-        self._error_queue = queue.Queue(self.MAX_ERROR_QUEUE_SIZE)
         self._worker = None
+        self._lock = threading.Lock()
```
westarle, line 32: self._worker_lock = threading.Lock() # protects all access to _worker.

google/auth/_refresh_worker.py
clundin25: Makes sense, moved these to be inlined

google/auth/_refresh_worker.py
clundin25: Added this comment.

google/auth/credentials.py
```diff
@@ -171,11 +214,26 @@ def before_request(self, request, method, url, headers):
...
+        else:
+            self._blocking_refresh(request)
+
         metrics.add_metric_header(headers, self._metric_header_for_usage())
         self.apply(headers)
 
+    def with_non_blocking_refresh(self):
+        self._use_non_blocking_refresh = True
+
+    def get_background_refresh_error(self):
```
westarle, line 231: I think I would prefer to remove this from the public API:
- it is not relevant without the feature flag
- it isn't clear when you'd want to call it
- it's hard to implement safely

google/auth/credentials.py
```diff
@@ -422,3 +480,16 @@ def signer(self):
         # pylint: disable=missing-raises-doc
         # (pylint doesn't recognize that this is abstract)
         raise NotImplementedError("Signer must be implemented.")
+
+
+class TokenState(Enum):
+    """
+    Tracks the state of a token.
+    FRESH: The token is valid. It is not expired or close to expired, or the token has no expiry. To make it mutually exclusive to STALE.
```
westarle, line 491: sorry about this: the "To make it mutually exclusive to STALE" part wasn't meant to be part of the comment.

google/auth/_refresh_worker.py
```diff
@@ -0,0 +1,113 @@
...
+                "Unable to start refresh. cred and request must be valid and instantiated objects."
+            )
+
+        with self._lock:
+            if self._worker is not None and self._worker._error_info is not None:
+                # Reset error field to prevent deadlock for clients that try to
+                # rety this error.
+                err, self._worker._error_info = self._worker._error_info, None
+
+                raise e.RefreshError(
```
westarle, line 54: this has a kind of weird 'every 2nd call fails' pattern that doesn't seem quite right.

I would suggest we try to background refresh once, log and record the error, then do foreground refreshes from then on until we get a new token.

You might return `false` from `start_refresh` to cause the caller to call `refresh` on their own thread.  The caller would be responsible for calling "clear error" any time they have a good token from `refresh()`.

google/auth/_refresh_worker.py
```diff
@@ -0,0 +1,113 @@
...
+                    f"Could not start a background refresh. The background refresh previously failed with {self._worker._error_info}."
+                ) from err
+
+            if self._worker is None or not self._worker.is_alive():  # pragma: NO COVER
+                self._worker = RefreshThread(cred=cred, request=request)
+                self._worker.start()
+
+    def has_error(self):
```
westarle, line 62: I would remove this and `get_error` from the public API since they are subject to the lock and difficult to reason about outside of this class.

google/auth/credentials.py
```diff
@@ -154,6 +186,20 @@ def apply(self, headers, token=None):
         if self.quota_project_id:
             headers["x-goog-user-project"] = self.quota_project_id
 
+    def _blocking_refresh(self, request):
+        if not self.valid:
+            self.refresh(request)
+
+    def _non_blocking_refresh(self, request):
+        if (
```
westarle, line 194: following the comment above, this would become:

```
if self.token_state == TokenState.FRESH:
  return

need_refresh = True
if self.token_state == TokenState.STALE:
  need_refresh = not self.refresh_worker.start_refresh(self, request)

if need_refresh:
  self.refresh(request)
  self._refresh_worker.clear_error()

File: google/auth/_refresh_worker.py [hunk 17]
```diff
@@ -0,0 +1,113 @@
...
+    Organizes exactly one background job that refresh a token.
+    """
+
+    def __init__(self):
+        """Initializes the manager."""
+
+        self._worker = None
+        self._lock = threading.Lock()  # protects access to worker threads.
+
+    def start_refresh(self, cred, request):
```
westarle, line 34: is `request` immutable or can we create a copy here?

google/auth/_refresh_worker.py
clundin25: Sounds good!

google/auth/_refresh_worker.py
clundin25: Okay that sounds good!

google/auth/credentials.py
clundin25: Done! I kept the `INVALID` case, should a refresh was never called in the STALE state.

google/auth/_refresh_worker.py
clundin25: Can you expand on why copy it? I don't think it is necessary, as the request should not be re-used in other places.  

I think copying the request would be a breaking change. 

I'd have to do some work to understand the implication of deep copying a request object.

google/auth/_refresh_worker.py
```diff
@@ -47,40 +52,19 @@ def start_refresh(self, cred, request):
...
-          Optional[exceptions.Exception]
+    def clear_error(self):
         """
-        if not self._worker:
-            return None
-        err, self._worker._error_info = self._worker._error_info, None
-        return err
+      Removes any errors that were stored from previous background refreshes.
+      """
+        if self._worker:
```
westarle, line 66: Please guard these 2 lines with the lock.

google/auth/credentials.py
```diff
@@ -191,14 +191,16 @@ def _blocking_refresh(self, request):
...
 
-        if self.token_state == TokenState.INVALID:
+        if self.token_state == TokenState.STALE:
+            use_blocking_refresh_fallback = not self._refresh_worker.start_refresh(
+                self, request
+            )
+
+        if self.token_state == TokenState.INVALID or use_blocking_refresh_fallback:
             self.refresh(request)
+            self._refresh_worker.clear_error()
```
westarle, line 203: maybe leave a comment describing why we need to call this right here.

westarle (approved):

google/auth/credentials.py
clundin25: Done!

PR #1371, opened 2023-08-16
chore: Add deprecation notice for 3.7.

clundin25:


arithmetic1728:
it seems some unit test is needed to cover the new changes

clundin25:
> it seems some unit test is needed to cover the new changes

Good catch, I'll add those

arithmetic1728:
gcloud will drop 3.7 on sep 26, let's hold on until then

clundin25:
@arithmetic1728 sounds good 👍

clundin25:
@arithmetic1728 I am okay with waiting but I think this change is still safe, it is just a deprecation notice that would impact some users

arithmetic1728 (approved):

PR #1374, opened 2023-08-22
fix: detect correct auth when ADC env var is set but empty

### LINKED ISSUE #983: Default credentials fail to detect running on GCP when ADC env var is empty
In the final layer of a container, it is pretty common to clear out env vars by setting them to empty. When `GOOGLE_APPLICATION_CREDENTIALS` exists but is empty (e.g. after pulling packages from Google Artifact Registry), the default credentials for the pubsub transport fails to use the assigned service account / metadata server.

#### Environment details

  - OS type and version: `Linux 5.10.176+`
  - Python version: `Python 3.11.2`
  - pip version: `pip 23.0.1 from /usr/lib/python3/dist-packages/pip (python 3.11)`
  - `google-cloud-pubsub` version: `2.18.3`

#### Steps to reproduce

  1.  Set the application creds env var to an empty value
  2. Attempt to create a publisher client while on GCP / GKE with a service account assigned

#### Code example

```python
import asyncio
from google.pubsub_v1.services.publisher.async_client import PublisherAsyncClient
async def setup():
  publisher_client = PublisherAsyncClient(transport="grpc_asyncio")
  return publisher_client

loop = asyncio.new_event_loop()
loop.run_until_complete(setup())
```

#### Stack trace
```
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "<stdin>", line 2, in setup
  File "/usr/local/lib/python3.11/dist-packages/google/pubsub_v1/services/publisher/async_client.py", line 215, in __init__
    self._client = PublisherClient(
                   ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/google/pubsub_v1/services/publisher/client.py", line 492, in __init__
    self._transport = Transport(
                      ^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/google/pubsub_v1/services/publisher/transports/grpc_asyncio.py", line 198, in __init__
    super().__init__(
  File "/usr/local/lib/python3.11/dist-packages/google/pubsub_v1/services/publisher/transports/base.py", line 104, in __init__
    credentials, _ = google.auth.default(
                     ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/google/auth/_default.py", line 659, in default
    credentials, project_id = checker()
                              ^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/google/auth/_default.py", line 652, in <lambda>
    lambda: _get_explicit_environ_credentials(quota_project_id=quota_project_id),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/google/auth/_default.py", line 272, in _get_explicit_environ_credentials
    credentials, project_id = load_credentials_from_file(
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/google/auth/_default.py", line 116, in load_credentials_from_file
    raise exceptions.DefaultCredentialsError(
google.auth.exceptions.DefaultCredentialsError: File  was not found.
```

mbrancato:
Treat empty strings as if the env var was not set. This fixes an issue where an environment variable for `GOOGLE_APPLICATION_CREDENTIALS` is set, but it empty. Because the default sentinel is `None`, the auth library attempts to load this file. Some systems can set or update an env var, but not unset an env var (e.g. Docker).

Fixes: https://github.com/googleapis/python-pubsub/issues/983


BigTailWolf (approved):

chalmerlowe (approved):
LGTM

google/auth/_default.py
```diff
@@ -330,31 +330,31 @@ def _get_explicit_environ_credentials(quota_project_id=None):
     from google.auth import _cloud_sdk
 
     cloud_sdk_adc_path = _cloud_sdk.get_application_default_credentials_path()
-    explicit_file = os.environ.get(environment_vars.CREDENTIALS)
+    explicit_file = os.environ.get(environment_vars.CREDENTIALS, "")
```
daniel-sanche, line 333: I'm not sure if I like using "" to represent the empty state. That seems like it could lead to confusing situations down the line. Instead, I'd recommend:

1. changing this line to `os.environ.get(environment_vars.CREDENTIALS) or None`, which will replace empty strings with `None`

or

2. use `if explicit_file:` in place of the checks, which will respond to None and empty strings the same way, since both are falsy

daniel-sanche (approved):
Consider my comment, but I think this is also ok to merge as-is

google/auth/_default.py
chalmerlowe: Daniel:

Thanks for this comment. I considered similar ideas when evaluating this, and this is the conclusion I came to:

The variable `explicit_file` is a **local** variable within `_get_explicit_environ_credentials()`, so no external API is changed or affected and the way the code is currently written, it must be a `str`. Why? Because of the default behavior of the `.get()` method on a `dict` and because of the return object from `.get_application_default_credentials_path()` which is a `str`.


#### dict .get()
Environmental variables are stored as strings AND if the `CREDENTIALS` variable is not found, we return an empty string as the default value. There is no reason to return a `None`, it serves no additional purpose in the limited uses that we have for the variable `explicit_file`.
```
explicit_file = os.environ.get(environment_vars.CREDENTIALS, "")
```


#### _cloud_sdk.get_application_default_credentials_path()
This function returns a string.
```
def get_application_default_credentials_path():
    """Gets the path to the application default credentials file.

    The path may or may not exist.

    Returns:
        str: The full path to application default credentials.
    """
```

In the comparision, we are checking to see IF `explicit_file` was unset AND if not, whether the path matches the path referenced by `cloud_sdk_adc_path`. There is no need OR reason to prefer None over the empty string.
```
if explicit_file != "" and explicit_file == cloud_sdk_adc_path:
```

#### So why use the empty string?

Old Behavior: Defaulted to `None` if unset.
New Behavior: Defaults to `""` if unset (via `os.environ.get(..., "")`).
Result: The logic now treats "unset" and "set to empty string" identically (as `""`). The check `explicit_file != ""` correctly handles both cases (unset OR empty string) by skipping the file load, which prevents a crash. `cloud_sdk_adc_path` **always** returns a string path, so the comparison remains safe.

Before I had fully worked through the above logic, I had considered something such as:

`if explicit_file not in {None, ""}` and feel that while `if explicit_file not in {None, ""}` OR similar patterns (such as you recommend) where we look for both is a valid defensive pattern, the current PR's approach is slightly cleaner because it enforces type consistency. If we initialize `explicit_file` with a default of `""`, the variable is **always** a `str`, avoiding mixed types (`Optional[str]`). The current implementation `explicit_file != ""` is correct given the initialization change.

PR #1375, opened 2023-08-24
chore: remove duplicate variable

jackwotherspoon:
The global variable [`_DEFAULT_TOKEN_LIFETIME_SECS`](https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/impersonated_credentials.py#L40-L61) is set twice. Can we get rid of this duplicate?


clundin25 (approved):

BigTailWolf (approved):
LGTM

PR #1377, opened 2023-08-31
fix: missing ssj for impersonate cred

arithmetic1728:
b/294308809

For the impersonated credential, when the source cred is a service account and self signed jwt is enabled for source cred, after we create a copy of the source cred with IAM scopes, the source cred's internal _jwt_credential is lost so self signed jwt can no longer be used. We need to recreate one with the IAM scopes so self signed jwt can be used.


BigTailWolf (approved):
LGTM

PR #1380, opened 2023-09-06
fix: expose universe domain in credentials

arithmetic1728:
Expose `universe_domain` by adding a property in the credential base class.


clundin25 (approved):

tests/oauth2/test_service_account.py
```diff
@@ -71,6 +71,7 @@ def test_constructor_no_universe_domain(self):
             SIGNER, self.SERVICE_ACCOUNT_EMAIL, self.TOKEN_URI, universe_domain=None
         )
         assert credentials._universe_domain == service_account._DEFAULT_UNIVERSE_DOMAIN
```
sai-sunder-s, line 73: do we need to check the private variable as well now?

arithmetic1728: right, it's not necessary, deleted that

sai-sunder-s (approved):

Issue #1089, opened 2022-07-22
`TypeError: string indices must be integers` when using bucket.exists() in Google Cloud Code


Ark-kun:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/google-auth-library-python/issues

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS:
  - Python version: 3.9
  - `google-auth` version: google-auth==2.9.1

#### Steps to reproduce

  1. Go to Google's official Storage samples: https://github.com/googleapis/python-storage/tree/main/samples
  2. Click on the "List buckets" sample: https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googleapis/python-storage&page=editor&open_in_editor=samples/snippets/storage_list_buckets.py
  3. Try to run sample.

(Cloud Code was in ephemeral mode)
```
Traceback (most recent call last):
  File "/home/avolkov_ota/cloudshell_open/python-storage/samples/snippets/storage_list_buckets.py", line 35, in <module>
    list_buckets()
  File "/home/avolkov_ota/cloudshell_open/python-storage/samples/snippets/storage_list_buckets.py", line 27, in list_buckets
    for bucket in buckets:
  File "/usr/local/lib/python3.9/dist-packages/google/api_core/page_iterator.py", line 208, in _items_iter
    for page in self._page_iter(increment=False):
  File "/usr/local/lib/python3.9/dist-packages/google/api_core/page_iterator.py", line 244, in _page_iter
    page = self._next_page()
  File "/usr/local/lib/python3.9/dist-packages/google/api_core/page_iterator.py", line 373, in _next_page
    response = self._get_next_page_response()
  File "/usr/local/lib/python3.9/dist-packages/google/api_core/page_iterator.py", line 432, in _get_next_page_response
    return self.api_request(
  File "/home/avolkov_ota/.local/lib/python3.9/site-packages/google/cloud/storage/_http.py", line 73, in api_request
    return call()
  File "/usr/local/lib/python3.9/dist-packages/google/api_core/retry.py", line 283, in retry_wrapped_func
    return retry_target(
[...]

PR #1382, opened 2023-09-18
Fix: Less restrictive content-type header check for google authentication (ignores charset)

DanieleQuasimodo:
Less restrictive content-type header check for google authentication (now only the media-type is compared, ignoring charset, boundary, etc).
Should fix issue #1089.

Implemented as suggested in PEP-594 (due to deprecation of cgi.parse_header. See [https://peps.python.org/pep-0594/#cgi](https://peps.python.org/pep-0594/#cgi)).


arithmetic1728:
@DanieleQuasimodo thank you for your PR. Could you add some unit tests to cover your change (for example, https://github.com/googleapis/google-auth-library-python/blob/main/tests/test__helpers.py and https://github.com/googleapis/google-auth-library-python/blob/main/tests/compute_engine/test__metadata.py).

DanieleQuasimodo:
[These](https://github.com/googleapis/google-auth-library-python/pull/1382/commits/bda4f39ccb8efeffc0eaecfa227bad08a8e9162f) should do.
The _metadata test is probably a bit redundant (_metadata inner workings haven't really changed, except for invoking the new function which is already tested in the _helpers tests), but I still added it, just in case.

The forced push to my fork was just because I'm dumb, committed with the wrong account and had to ammend the author (so Google CLA is happy) 😅

arithmetic1728:
@DanieleQuasimodo The PR looks great, thanks! Could you run `python -m nox -s lint` to format the files (this caused kokoro build failure)? After that I will go ahead to merge the PR.

DanieleQuasimodo:
@arithmetic1728 I think now it should be ok.
I ran both blacken and lint

arithmetic1728 (approved):

PR #1384, opened 2023-09-19
fix: Trust boundary meta header renaming and using the schema from backend team.

BigTailWolf:
* A cached `self._trust_boundary` for service account and external account type is be initialized as "not opt-in" (`0x0`) of encoded_locations.
* The meta header is renamed from `x-identity-trust-boundary` to `x-allowed-locations`
* Fix the CI tests accordingly


TimurSadykov (approved):
LGTM

lsirac (approved):

PR #1389, opened 2023-09-26
fix: update urllib3 to >= 2.0.5

### LINKED ISSUE #1365: Support for urllib3 2
Hi - I see that urllib3 was pinned to <2.0 in #1283. Is there any plan to support 2.0 and above?

Thanks,
Tom

arithmetic1728:
fixes #1365 

urllib3 added back the pyopenssl support and removed the depreciation warning since 2.0.5, so we can upgrade the urllib3 dependency version now.

urllib3 made `RequestMethods` class private since 2.0.0 (changed from `urllib3.request.RequestMethods` to `urllib3._request_methods.RequestMethods`). We could copy the whole class and the related code into google-auth, but that would be a lot of code, so we let AuthorizedHttp inherit the private class instead, which seems to be the best solution so far.


clundin25 (approved):

danmoz:
Just a heads up, this is going to cause a lot of pain -- some major packages (for example, botocore) in the python ecosystem require `urllib3<2.0`. By requiring a version >= 2.0.5, we'll create an unresolvable dependency conflict.

An example of this is `botocore`, which [pins to <1.27](https://github.com/boto/botocore/blob/82e53f2191d9470b73826ade800c31b1a6b897ef/setup.cfg#L9)

arithmetic1728:
@danmoz thank you for the heads up, I will fix it asap.

arithmetic1728:
problem is fixed in release 2.23.2, https://github.com/googleapis/google-auth-library-python/pull/1391

PR #1405, opened 2023-11-03
feat: Implement trust boundary lookup and caching for Service Account credentials

BigTailWolf:
We are going to implement support for service account credentials trust boundary via a cache value.

## Usage of the Value

Whenever we apply a credentials to a request header, after we apply the authorization header with access_token, we are going to check the trust boundary value. If the trust boundary value is not `None`, we apply a header `x-allowed-locations` with the trust boundary value.

## Update of the Value

There are two ways we can do the update of the value:

* Injection: Provide a direct setter/getter to the trust boundary value.

* Trust boundary lookup: We are not going to include in this change since the global lookup endpoint is not launched yet.


google/oauth2/_client.py
```diff
@@ -505,3 +509,81 @@ def refresh_grant(
...
+                "locations": [],
+                "": "0x0"
+            }
+
+    Raises:
+        exceptions.TransportError: If the request to lookup endpoint fails.
+        exceptions.RefreshError: If the query response not 200 or 404.
+        exceptions.MalformedError: If the response not in valid format.
+    """
+    url = _TRUST_BOUNDARY_LOOKUP_ENDPOINT.format(universe, service_account)
```
aeitzman, line 553: Are the trust boundary endpoints going to be active for PSC environments? If so, this needs to be overriddable, if not, we should confirm that in a PSC environment, this flow is ignored. I think it would probably return not found, but not 100% sure, and if it doesn't this change would break all PSC users.

google/oauth2/_client.py
```diff
@@ -505,3 +509,81 @@ def refresh_grant(
...
+
+            {
+                "locations": [],
+                "": "0x0"
+            }
+
+    Raises:
+        exceptions.TransportError: If the request to lookup endpoint fails.
+        exceptions.RefreshError: If the query response not 200 or 404.
+        exceptions.MalformedError: If the response not in valid format.
```
aeitzman, line 551:
```suggestion
        exceptions.TransportError: If the request to the lookup endpoint fails.
        exceptions.RefreshError: If the query response status code is not 200 or 404.
        exceptions.MalformedError: If the response is not in a valid format.
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,81 @@ def refresh_grant(
...
+            HTTP requests.
+
+        universe (str): Universe domain. (GDU/TPC universe)
+
+        service_account (str): The service account to lookup.
+
+        access_token (str): A valid access_token used as a basic auth.
+
+    Returns:
+        Mapping[str,list|str, str]: A response would be a dictionary contains
```
aeitzman, line 528:
```suggestion
        Mapping[str,list|str, str]: A response would be a dictionary containing
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,81 @@ def refresh_grant(
...
+
+            e.g:
+            {
+                "locations": [
+                    "us-central1", "us-east1", "europe-west1", "asia-east1"
+                ],
+                "encoded_locations": "0xA30"
+            }
+
+            In case of a universe global lookup not launched yet, a default
```
aeitzman, line 540:
```suggestion
            In the universe global lookup is not launched yet, a default
```

google/oauth2/service_account.py
```diff
@@ -441,6 +447,12 @@ def refresh(self, request):
             self.token = access_token
             self.expiry = expiry
 
+        # Either the trust boundary has not yet fetch or been removed.
```
aeitzman, line 450:
```suggestion
        # Either the trust boundary has been removed or not yet been fetched.
```

google/oauth2/service_account.py
```diff
@@ -429,6 +429,12 @@ def refresh(self, request):
                 "domain wide delegation is not supported for non-default universe domain"
             )
 
+        # If a refresh is triggered under the case of credential not expired
+        # that means some abnormal cases proactively trigger this refresh, we
+        # need to update the trust boundary cache.
+        if self.valid:
```
aeitzman, line 435: I'm a bit confused, why does the trust boundary need to be refreshed here?

google/oauth2/_client.py
BigTailWolf: If we don't launch in any environment, we will go for a flow for `404` which fall backs to the current behavior

google/oauth2/service_account.py
BigTailWolf: The `refresh` method can be triggered by client code. In the library normal flow, we have 
https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/credentials.py#L174-L175
```
        if not self.valid:
            self.refresh(request)
```
which is the flow of refresh in a case of credentials expired.

But one example of client `gcloud`, we are also do a `refresh` when the abnormal retryable case of authentication issue happens.

google/oauth2/_client.py
aeitzman: Actually, thinking about it more, I'm worried about this breaking the flow in general. The trust boundary backend isn't actually live yet right? And even when it does go live it isn't actually required for BYOID? If the trust boundary service is broken (say returning a 503) this logic would break auth for everyone. I would think we would want to ignore it in that case.

google/oauth2/service_account.py
aeitzman: but why do we want to reset the trust boundary in that case?

google/oauth2/_client.py
BigTailWolf: The `refresh` method of external_account don't do this part. We just have the hard coded "enable all" values in the credential objects and apply it to requests when "apply" it.

BigTailWolf: @aeitzman Did some modification only processing the potential errors under `200` return. For others we fall back to the old logic to give a default trust boundary.

google/oauth2/service_account.py
BigTailWolf: That means an abnormal auth status happens, which could include trust boundary error. We only do the reset trust boundary in that case. Otherwise, a trust boundary value will be carried from the creating of credential object till the entire lifecycle end.

google/oauth2/_client.py
aeitzman: Sure, sounds good. Let's wait to finalize on this until we get a good answer from the backend team though.

arithmetic1728 (approved):
LGTM

google/auth/credentials.py
```diff
@@ -236,6 +223,40 @@ def with_non_blocking_refresh(self):
         self._use_non_blocking_refresh = True
 
 
+class CredentialsWithTrustBoundary(Credentials):
+    """Abstract base for credentials supporting trust boundary factory"""
```
aeitzman, line 227: nit: supporting trust boundary factory? Or trust boundaries in general? Also can you add a short explanation of what trust boundaries are?

google/auth/credentials.py
```diff
@@ -236,6 +223,40 @@ def with_non_blocking_refresh(self):
...
+        """
+        if self._trust_boundary is not None:
+            headers["x-allowed-locations"] = self._trust_boundary["encoded_locations"]
+
+    def lookup_trust_boundary(self, request):
+        """Lookup trust boundary shall be implemented by subclasses"""
+        raise NotImplementedError("Missing definition of trust boundary lookup")
+
+    def set_trust_boundary(self, trust_boundary):
+        """Set trust boudnary value to the credential"""
```
aeitzman, line 252:
```suggestion
        """Sets the trust boundary value to the credential"""
```

google/auth/credentials.py
```diff
@@ -236,6 +223,40 @@ def with_non_blocking_refresh(self):
         self._use_non_blocking_refresh = True
 
 
+class CredentialsWithTrustBoundary(Credentials):
+    """Abstract base for credentials supporting trust boundary factory"""
+
+    def apply(self, headers, token=None):
+        super().apply(headers, token)
+        """Trust boundary value will be a cached value from global lookup.
```
aeitzman, line 231: nit: Should this docstring be above the super call? Also, can the docstring describe what the function is doing, it's a little confusing currently

google/auth/credentials.py
```diff
@@ -236,6 +223,40 @@ def with_non_blocking_refresh(self):
...
+                "us-central1", "us-east1", "europe-west1", "asia-east1"
+            ]
+            "encoded_locations": "0xA30"
+        }
+        """
+        if self._trust_boundary is not None:
+            headers["x-allowed-locations"] = self._trust_boundary["encoded_locations"]
+
+    def lookup_trust_boundary(self, request):
+        """Lookup trust boundary shall be implemented by subclasses"""
```
aeitzman, line 248: nit: can you describe what this function is supposed to do? Also, does this need to be public? Would an external user ever need to retrieve the trust boundary?

google/auth/external_account.py
```diff
@@ -203,6 +200,10 @@ def _constructor_args(self):
             args.pop("workforce_pool_user_project")
         return args
 
+    def lookup_trust_boundary(self, reqeust):
+        """Current implementation of inject dummy headers"""
```
aeitzman, line 204: Can you rewrite this comment to describe what the function is actually supposed to do, and then have another sentence that explains currently it is just retrieving the default values.

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
         request, token_uri, body, can_retry=can_retry
     )
     return _handle_refresh_grant_response(response_data, refresh_token)
+
+
+def lookup_trust_boundary(request, url, headers):
+    """ Implements the global lookup of service account trust boundary
```
aeitzman, line 515:
```suggestion
    """ Implements the global lookup of a service account trust boundary.
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
...
+            In the universe global lookup is not launched yet, a default
+            trust boundary of "all" will be returned.
+
+            {
+                "locations": [],
+                "": "0x0"
+            }
+
+    Raises:
+        exceptions.TransportError: If the request to lookup endpoint fails.
```
aeitzman, line 547:
```suggestion
        exceptions.TransportError: If the request to the lookup endpoint fails.
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
...
+        exceptions.RefreshError: If the query response not 200.
```
aeitzman, line 548:
```suggestion
        exceptions.RefreshError: If the query response is not 200.
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
...
+        exceptions.MalformedError: If the response not in valid format.
```
aeitzman, line 549:
```suggestion
        exceptions.MalformedError: If the response is not in a valid format.
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
...
+
+            e.g:
+            {
+                "locations": [
+                    "us-central1", "us-east1", "europe-west1", "asia-east1"
+                ],
+                "encoded_locations": "0xA30"
+            }
+
+            In the universe global lookup is not launched yet, a default
```
aeitzman, line 538:
```suggestion
            If the universe global lookup is not launched yet, a default
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
...
+    Args:
+        request (google.auth.transport.Request): A callable used to make
+            HTTP requests.
+
+        url (str): The url of the look up request.
+
+        headers (mapping): Basic auth headers for the request.
+
+    Returns:
+        Mapping[str,list|str]: A response would be a dictionary containing the
```
aeitzman, line 526:
```suggestion
        Mapping[str,list|str]: A dictionary containing the
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
...
+            "locations" as a list of allowed locations in string and
```
aeitzman, line 527:
```suggestion
            "locations" as a list of allowed locations as strings and
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
...
+    Raises:
+        exceptions.TransportError: If the request to lookup endpoint fails.
+        exceptions.RefreshError: If the query response not 200.
+        exceptions.MalformedError: If the response not in valid format.
+    """
+    try:
+        response = request(method="GET", url=url, headers=headers)
+    except Exception as e:
+        raise exceptions.TransportError(
+            "Failed to make request to trust boundary global lookup endpoint {}".format(
```
aeitzman, line 555:
```suggestion
            "Failed to make request to the trust boundary global lookup endpoint {}".format(
```

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
...
+        )
+
+    response_body = (
+        response.data.decode("utf-8")
+        if hasattr(response.data, "decode")
+        else response.data
+    )
+    if response.status != http_client.OK:
+        raise exceptions.RefreshError(
+            "Failed look up trust boundary, {}, {}".format(
```
aeitzman, line 567:
```suggestion
            "Trust boundary look up failed , {}, {}".format(
```

google/oauth2/_client.py
```diff
@@ -43,6 +43,10 @@
     "https://iamcredentials.googleapis.com/v1/"
     + "projects/-/serviceAccounts/{}:generateIdToken"
 )
+_TRUST_BOUNDARY_LOOKUP_ENDPOINT = (
```
aeitzman, line 46: This seems to be duplicated in service account

google/oauth2/service_account.py
```diff
@@ -81,13 +83,20 @@
...
 from google.oauth2 import _client
 
+_LOGGER = logging.getLogger(__name__)
+
 _DEFAULT_TOKEN_LIFETIME_SECS = 3600  # 1 hour in seconds
 _GOOGLE_OAUTH2_TOKEN_ENDPOINT = "https://oauth2.googleapis.com/token"
+_TRUST_BOUNDARY_LOOKUP_ENDPOINT = (
+    "iamcredentials.{}/v1/projects/-/serviceAccounts/{}/trustBoundary"
+)
+TRUST_BOUNDARY_LOOKUP_ENABLED_ENV = "TRUST_BOUNDARY_LOOKUP_ENABLED"
```
aeitzman, line 93: can you add a comment explaining what this is?

google/oauth2/service_account.py
```diff
@@ -302,6 +314,13 @@ def with_scopes(self, scopes, default_scopes=None):
         cred._default_scopes = default_scopes
         return cred
 
+    def lookup_trust_boundary(self, request):
```
aeitzman, line 317: Add docstring

google/oauth2/service_account.py
```diff
@@ -436,6 +455,26 @@ def refresh(self, request):
...
+            try:
+                self._trust_boundary = self.lookup_trust_boundary(request)
+            except Exception as err:
+                if not self._trust_boundary:
+                    raise err
+                # if we already have trust boundary values, we log the error
+                # and keep the cached trust boundary value to let it fail
+                # afterwards.
+                else:
+                    _LOGGER.error(f"trust boundary refresh failed due to {err}")
```
aeitzman, line 476: should this be a warning instead of an error? Also, explain that a cached value was used instead.

google/oauth2/service_account.py
```diff
@@ -436,6 +455,26 @@ def refresh(self, request):
...
+        if (
+            self.valid
+            and self.universe_domain in _client.TRUST_BOUNDARY_ENABLED_UNIVERSES
+            and self._trust_boundary_enabled
+        ):
+            try:
+                self._trust_boundary = self.lookup_trust_boundary(request)
+            except Exception as err:
+                if not self._trust_boundary:
+                    raise err
```
aeitzman, line 471: can the exception message explain that trust boundary lookup failed and a cache was not available?

google/oauth2/service_account.py
```diff
@@ -448,6 +487,14 @@ def refresh(self, request):
             self.token = access_token
             self.expiry = expiry
 
+        # In case of trust boundary never been fetched.
+        if (
+            self._trust_boundary is None
+            and self.universe_domain in _client.TRUST_BOUNDARY_ENABLED_UNIVERSES
+            and self._trust_boundary_enabled
```
aeitzman, line 494: should we throw an error or log something if the user is trying to enable trust boundaries but isn't in a supported domain? Seems like we just ignore it silently

google/auth/credentials.py
```diff
@@ -236,6 +223,40 @@ def with_non_blocking_refresh(self):
...
+        raise NotImplementedError("Missing definition of trust boundary lookup")
+
+    def set_trust_boundary(self, trust_boundary):
+        """Sets the trust boundary value to the credential"""
+        self._trust_boundary = trust_boundary
```
clundin25, line 253: Use a python property instead: https://stackoverflow.com/questions/6618002/using-property-versus-getters-and-setters

google/auth/external_account.py
```diff
@@ -130,10 +130,7 @@ def __init__(
         self._default_scopes = default_scopes
         self._workforce_pool_user_project = workforce_pool_user_project
         self._universe_domain = universe_domain or credentials.DEFAULT_UNIVERSE_DOMAIN
-        self._trust_boundary = {
-            "locations": [],
-            "encoded_locations": "0x0",
-        }  # expose a placeholder trust boundary value.
+        self._trust_boundary = credentials.DEFAULT_TRUST_BOUNDARY
```
clundin25, line 133: This needs to a deep copy or trust boundary modifications would impact other credential objects.

google/auth/credentials.py
```diff
@@ -24,6 +24,7 @@
 from google.auth import metrics
 from google.auth._refresh_worker import RefreshThreadManager
 
+DEFAULT_TRUST_BOUNDARY = {"locations": [], "encoded_locations": "0x0"}
```
clundin25, line 27: I would not use a global dictionary like this, it is prone to race conditions. 

It would be shared by every credential object instantiated with it.

google/auth/external_account.py
clundin25: It is not clear what this function is for. Also there is a typo for `request`.

google/oauth2/_client.py
clundin25: It is okay for oauth2 module to depend on auth module.

google/oauth2/_client.py
```diff
@@ -505,3 +509,77 @@ def refresh_grant(
         request, token_uri, body, can_retry=can_retry
     )
     return _handle_refresh_grant_response(response_data, refresh_token)
+
+
+def lookup_trust_boundary(request, url, headers):
+    """ Implements the global lookup of a service account trust boundary.
```
clundin25, line 515: Can you add how and when this function should be used.

google/oauth2/service_account.py
clundin25: Is this documented in an AIP?

google/oauth2/service_account.py
```diff
@@ -192,7 +200,12 @@ def __init__(
             self._additional_claims = additional_claims
         else:
             self._additional_claims = {}
-        self._trust_boundary = {"locations": [], "encoded_locations": "0x0"}
+        self._trust_boundary_enabled = (
+            os.environ.get(TRUST_BOUNDARY_LOOKUP_ENABLED_ENV) == "1"
```
clundin25, line 204: This is inconsistent with https://google.aip.dev/auth/4110, where "true|false" is used

My suggestion:

Define this in an AIP.

Check for the presence of the ENV variable to enable it. It can be unset to "toggle" off.

google/oauth2/service_account.py
clundin25, line 93: I think "ENABLE_GOOGLE_TRUST_BOUNDARY" would be more clear and aligned with existing ENV variables.

google/oauth2/service_account.py
```diff
@@ -302,6 +314,13 @@ def with_scopes(self, scopes, default_scopes=None):
         cred._default_scopes = default_scopes
         return cred
 
+    def lookup_trust_boundary(self, request):
+        url = _TRUST_BOUNDARY_LOOKUP_ENDPOINT.format(
+            self.universe_domain, self.service_account_email
+        )
+        headers = {"Authorization": "Basic " + self.token}
```
clundin25, line 321: I'm not familiar with this endpoint, but should Basic be Bearer?

google/oauth2/service_account.py
```diff
@@ -436,6 +455,26 @@ def refresh(self, request):
                 "domain wide delegation is not supported for non-default universe domain"
             )
 
+        # If a refresh is triggered under the case of credential not expired
+        # that means some abnormal cases proactively trigger this refresh, we
+        # will attemp to fetch trust boundary but tolerant on errors if we
+        # already have cached trust boundaries.
+        if (
+            self.valid
```
clundin25, line 463: This attribute is deprecated. New code should use the token_state attribute.

google/oauth2/service_account.py
```diff
@@ -436,6 +455,26 @@ def refresh(self, request):
                 "domain wide delegation is not supported for non-default universe domain"
             )
 
+        # If a refresh is triggered under the case of credential not expired
+        # that means some abnormal cases proactively trigger this refresh, we
+        # will attemp to fetch trust boundary but tolerant on errors if we
+        # already have cached trust boundaries.
```
clundin25, line 461: Can you expand on why the trust boundary cache needs to be refreshed?

Are trust boundaries expected to change periodically?

google/oauth2/service_account.py
```diff
@@ -436,6 +455,26 @@ def refresh(self, request):
...
+        if (
+            self.valid
+            and self.universe_domain in _client.TRUST_BOUNDARY_ENABLED_UNIVERSES
+            and self._trust_boundary_enabled
+        ):
+            try:
+                self._trust_boundary = self.lookup_trust_boundary(request)
+            except Exception as err:
+                if not self._trust_boundary:
```
clundin25, line 470:
```suggestion
                if not isinstance(self._trust_boundary, dict):
```

google/oauth2/service_account.py
clundin25, line 321: Can this be an f string?

google/oauth2/service_account.py
clundin25: +1 to warning. 
```suggestion
                    _LOGGER.warning(f"Failed to refresh trust boundary cache due to {err}. Falling back to existing cache.")
```

google/oauth2/service_account.py
```diff
@@ -436,6 +455,26 @@ def refresh(self, request):
...
             )
 
+        # If a refresh is triggered under the case of credential not expired
+        # that means some abnormal cases proactively trigger this refresh, we
+        # will attemp to fetch trust boundary but tolerant on errors if we
+        # already have cached trust boundaries.
+        if (
+            self.valid
+            and self.universe_domain in _client.TRUST_BOUNDARY_ENABLED_UNIVERSES
+            and self._trust_boundary_enabled
```
clundin25, line 465: I think this condition should be wrapped in a method. I see it in multiple locations.

google/oauth2/service_account.py
```diff
@@ -448,6 +487,14 @@ def refresh(self, request):
             self.token = access_token
             self.expiry = expiry
 
+        # In case of trust boundary never been fetched.
```
clundin25, line 490: Can this be merged with the above section? We should only fetch the trust boundary in one section of code.

tests/oauth2/test__client.py
```diff
@@ -619,3 +619,81 @@ def test__token_endpoint_request_no_throw_with_retry(can_retry):
...
+        mock_response.data = data.get("response_data")
+
+        mock_request = mock.create_autospec(transport.Request)
+        mock_request.return_value = mock_response
+
+        if not expect_error:
+            trust_boundary = _client.lookup_trust_boundary(mock_request, "mock_url", {})
+            assert trust_boundary == data.get("expect_trust_boundary")
+            # assert trust_boundary['locations'] == data.get('expect_trust_boundary')['locations']
+            # assert trust_boundary['encoded_locations'] == data.get('expect_trust_boundary')['encoded_locations']
```
clundin25, line 696: nit: Don't commit comments.

tests/oauth2/test__client.py
```diff
@@ -619,3 +619,81 @@ def test__token_endpoint_request_no_throw_with_retry(can_retry):
...
+def test_lookup_trust_boundary_request_error():
+    mock_request = mock.create_autospec(transport.Request)
+    mock_request.side_effect = Exception("Mock Request Error")
+
+    with pytest.raises(exceptions.TransportError):
+        _client.lookup_trust_boundary(mock_request, "mock_url", {})
+
+
+def test_lookup_trust_boundary():
+    test_data = [
```
clundin25, line 633: nit: unroll this into multiple test cases, otherwise it is hard to tell which test failed. Maybe use a pytest fixture? 

https://docs.pytest.org/en/stable/how-to/parametrize.html

tests/oauth2/test__client.py
```diff
@@ -619,3 +619,81 @@ def test__token_endpoint_request_no_throw_with_retry(can_retry):
...
+    for data in test_data:
+        expect_error, msg = data.get("expect_error"), data.get("expect_error_message")
+        mock_response = mock.create_autospec(transport.Response, instance=True)
+        mock_response.status = data.get("response_status")
+        mock_response.data = data.get("response_data")
+
+        mock_request = mock.create_autospec(transport.Request)
+        mock_request.return_value = mock_response
+
+        if not expect_error:
```
clundin25, line 692: I think this should be split into a separate test case, to simplify the test code.

tests/oauth2/test_service_account.py
```diff
@@ -481,6 +486,173 @@ def test_refresh_success(self, jwt_grant):
...
+        credentials._enable_trust_boundary()
+        token = "token"
+        jwt_grant.return_value = (
+            token,
+            _helpers.utcnow() + datetime.timedelta(seconds=500),
+            {},
+        )
+        request = mock.create_autospec(transport.Request, instance=True)
+
+        # Refresh credentials
```
clundin25, line 501: nit: Comments should describe why instead of what, so this comment is redundant.

[go/comments-codelab](https://goto.google.com/comments-codelab)

google/oauth2/service_account.py
BigTailWolf: The difference is in this case, the token is a new token and the case above is using a valid token.

tests/oauth2/test__client.py
BigTailWolf: Fixture is a good call.
I have a plan to refactoring of each function mapping to one test with fixtures as unit test.
A whole setup scenario of a workflow as an integration test.
This is a long term gradually refactoring by a flexible pace.

tests/oauth2/test__client.py
BigTailWolf: One `if` to control we test the happy case and the error cases. I think this is to simplify test into one method have one associate test.

google/oauth2/service_account.py
BigTailWolf: Trust boundary won't be changed periodically. A one time set can last years long.
But during a runtime of a workload. Someone on the customer admin side can edit policies and cause a trust boundary change.
In this case, a request routing to a region which is no longer valid trust boundary will get `401` from backend.
And this `401` will naturally cause a refresh in most of the client as a common practice.

google/oauth2/service_account.py
BigTailWolf: This endpoint is not done yet. The current design is that would be "Basic"

google/oauth2/service_account.py
BigTailWolf: Thanks. Got it.

google/oauth2/service_account.py
BigTailWolf: The case is we are going to support injection on this phase. And through the injection scenario. We never do the "look_up". When we onboard the "look_up", I think "LOOKUP_ENABLED" is more descriptive.

google/auth/credentials.py
BigTailWolf: How about set the constant itself as a deepcopy?

google/oauth2/service_account.py
BigTailWolf: Only for presents or not? I thought when we explicitly set some variable, it should be exactly match a value we want. like `true`?

BigTailWolf: By the way. This is just a temporary solution for the phase since the backend is not finished on lookup endpoint.
Once they are done, we should go through this flow any way.

google/oauth2/_client.py
BigTailWolf: My bad to not delete this after I put all the default in the `auth.credentials`

google/oauth2/service_account.py
BigTailWolf: To check the enable of lookup itself is a temporary solution. That's for injection test phase when backend is not ready on lookup endpoint. Will be totally removed. Then since it's an `and` condition, it won't be true anyway if we are not in GDU.

google/oauth2/service_account.py
BigTailWolf: LOOUP_ENABLED check is a temporary solution and will be totally removed shortly.

lsirac: Can you expand on this a bit more? Once we release this, we can't remove it

BigTailWolf: Yeah, the backend team want to launch trust boundary injection first at this stage. And the lookup must be bypassed. But in the end they want the attempt lookup in the refresh flow. The `TRUST_BOUNDARY_LOOKUP_ENABLED` is just a temporary solution and would be removed once the lookup endpoint is ready.

lsirac (changes requested):
Thanks Jin. Please add a description to this PR

google/auth/credentials.py
```diff
@@ -259,7 +342,7 @@ def with_quota_project_from_environment(self):
 
 
 class CredentialsWithTokenUri(Credentials):
-    """Abstract base for credentials supporting ``with_token_uri`` factory"""
+    """Abstract base for credentials supporting ``with_token_uri`` factory."""
```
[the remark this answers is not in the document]
BigTailWolf: Oh, it's not passing the `docstring` check. Previously there's no such check. It's fixed by formatting tool.

google/auth/credentials.py
```diff
@@ -236,8 +222,105 @@ def with_non_blocking_refresh(self):
...
+        super().apply(headers, token)
+        if self._trust_boundary is not None:
+            headers["x-allowed-locations"] = self._trust_boundary["encoded_locations"]
+
+    def lookup_trust_boundary(self, request):
+        """Lookup trust boundary shall be implemented by subclasses.
+
+        Upon the lookup, we send a request to the global lookup endlpoint and then
+        parse the valid response. Service account credentials, workload identity
+        pools and workforce pools implementation will be various.
```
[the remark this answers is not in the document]
BigTailWolf: The endpoint would be different.

google/auth/external_account.py
BigTailWolf: I put a deep copy at the `DEFAULT_TRUST_BOUNDARY`. Would that be OK?

google/oauth2/service_account.py
```diff
@@ -436,6 +462,28 @@ def refresh(self, request):
                 "domain wide delegation is not supported for non-default universe domain"
             )
 
+        # If a refresh is triggered under the case of credential not expired
+        # that means some abnormal cases proactively trigger this refresh, we
+        # will attemp to fetch trust boundary but tolerant on errors if we
+        # already have cached trust boundaries.
+        if (
```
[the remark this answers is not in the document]
BigTailWolf: No. The STS endpoint doesn't require trust boundary header and using "dynamic routing" on their end for now. And the basic auth for the trust boundary lookup is using access_token.

google/auth/external_account.py
clundin25: No, I think there is a misunderstanding. The code is hitting this error https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument. By creating a DEFAULT dict, the dict is shared by all credential objects, because in this line of code you are grabbing a pointer to that dict.

Deep copying once at the global namespace is a NOOP, a deep copy in the constructor would accomplish what I believe your intent for the default is.

My suggestion is to use `DEFAULT_TRUST_BOUNDARY` as a singleton if you want to share the cache across all credential types, otherwise codify the default dict in your constructor, to avoid a race condition.

google/oauth2/service_account.py
clundin25: @westarle any opinions on this? 

What is the justification for the temporary environment variable?

google/auth/external_account.py
BigTailWolf: Oh, I made a copy on each use case.

google/auth/credentials.py
```diff
@@ -24,6 +23,7 @@
 from google.auth import metrics
 from google.auth._refresh_worker import RefreshThreadManager
 
+DEFAULT_TRUST_BOUNDARY = copy.deepcopy({"locations": [], "encoded_locations": "0x0"})
```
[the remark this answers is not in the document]
BigTailWolf: fix. using it in all the assign operations.

google/oauth2/service_account.py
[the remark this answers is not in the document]
lsirac: Then isn't the one call at the bottom enough?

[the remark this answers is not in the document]
BigTailWolf: The bottom call is using the refreshed access_token.

google/oauth2/service_account.py
lsirac: Why not just not do the lookup if trust boundary info is provided when the credentials are created? What is the environment variable for?

google/auth/credentials.py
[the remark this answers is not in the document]
lsirac: Ok, can you change the wording?

google/auth/external_account.py
```diff
@@ -203,6 +200,16 @@ def _constructor_args(self):
             args.pop("workforce_pool_user_project")
         return args
 
+    def lookup_trust_boundary(self, request):
+        """Current implementation of inject dummy headers
+        As we are not going to publish external_account lookup trust boundary
```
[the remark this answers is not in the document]
BigTailWolf: Changed to a private function. This is only called during refresh when lookup is enabled.

google/oauth2/service_account.py
BigTailWolf: When the credential instance created we don't yet have a valid access_token. Without a valid access token we cannot do basic auth for lookup trust boundary.

google/oauth2/service_account.py
BigTailWolf: Hi @clundin25 , we decided to get rid of a temporary control since any release of SDK code won't get back. For this change, we just avoid this part of the logic because the endpoint is not ready yet.

clundin25:
@BigTailWolf is this PR ready for another round of review?

BigTailWolf:
Hi @clundin25 and @lsirac , I removed the global lookup flow in the refresh. Now we only do supports on injection.

google/auth/credentials.py
```diff
@@ -11,19 +11,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-
```
clundin25, line 15: Avoid committing unrelated whitespace changes, it makes the diff larger.

google/auth/credentials.py
```diff
@@ -11,19 +11,20 @@
...
 from enum import Enum
+import json
 import os
 
 from google.auth import _helpers, environment_vars
 from google.auth import exceptions
 from google.auth import metrics
 from google.auth._refresh_worker import RefreshThreadManager
 
+DEFAULT_TRUST_BOUNDARY = {"locations": [], "encoded_locations": "0x0"}
```
clundin25, line 27: I think this needs a warning that this dictionary should not be used directly, but copied from as a base template.

google/auth/credentials.py
```diff
@@ -52,16 +53,18 @@ def __init__(self):
...
         """Optional[str]: Project to use for quota and billing purposes."""
         self._trust_boundary = None
         """Optional[dict]: Cache of a trust boundary response which has a list
         of allowed regions and an encoded string representation of credentials
         trust boundary."""
         self._universe_domain = DEFAULT_UNIVERSE_DOMAIN
-        """Optional[str]: The universe domain value, default is googleapis.com
-        """
+        """Optional[str]: The universe domain value, default is
+        googleapis.com."""
```
clundin25, line 67: nit:Avoid checking in white psace changes

google/auth/credentials.py
```diff
@@ -98,9 +101,7 @@ def valid(self):
 
     @property
     def token_state(self):
-        """
-        See `:obj:`TokenState`
-        """
+        """See `:obj:`TokenState`"""
```
clundin25, line 104: I'm curious why so many formatting changes happened.

Did you use `nox -s blacken`?

google/auth/credentials.py
BigTailWolf: Hi Carl, those are auto generated from lint fix tools since we failed the previous lint check.

google/auth/credentials.py
```diff
@@ -236,8 +223,113 @@ def with_non_blocking_refresh(self):
         self._use_non_blocking_refresh = True
 
 
+class CredentialsWithTrustBoundary(Credentials):
+    """Abstract base for credentials supporting trust boundary.
+
+    A credential with trust boundary will carry a trust boundary info as a
+    cache. The cache value can be either or fetch from a global lookup or
```
clundin25, line 230:
```suggestion
    cache. The cache value can either be fetched from a global lookup or
```

google/auth/credentials.py
```diff
@@ -236,8 +223,113 @@ def with_non_blocking_refresh(self):
...
+    injected via ``set_trust_boundary``.
+
+    Upon we apply credential headers to a request, a trust boundary value will
+    be added as a header. GFE will use the value to routing.
```
clundin25, line 234:
```suggestion
    On apply, a trust boundary value will be added as a header. GFE will use the trust-boundary header for routing.
```

google/auth/credentials.py
```diff
@@ -236,8 +223,113 @@ def with_non_blocking_refresh(self):
...
+    """
+
+    def apply(self, headers, token=None):
+        """Trust boundary value will be a cached value from global lookup.
```
clundin25, line 238: What does "global lookup" mean in this context?

BigTailWolf:
Hi @clundin25 , I'm going to close this change as I got an update on the change from backend on their design.

google/auth/credentials.py
```diff
@@ -236,8 +223,113 @@ def with_non_blocking_refresh(self):
...
+            {
+                "locations": [
+                    "us-central1", "us-east1", "europe-west1", "asia-east1"
+                ],
+                "encoded_locations": "0xA30"
+            }
+        """
+        super().apply(headers, token)
+        if self.trust_boundary is not None:
+            headers["x-allowed-locations"] = self.trust_boundary["encoded_locations"]
```
clundin25, line 256: Is it enough to store just the encoded locations? Where does locations get used?

google/auth/credentials.py
```diff
@@ -236,8 +223,113 @@ def with_non_blocking_refresh(self):
...
+        parse the valid response. Service account credentials, workload identity
+        pools and workforce pools will have different endpoints.
+
+        Args:
+            request (google.auth.transport.Request): A callable used to make
+                HTTP requests.
+
+            url (str): The url of the look up request.
+
+            headers (mapping): Basic auth headers for the request.
```
clundin25, line 271: This doesn't match the method parameters, which one is correct?

PR #1409, opened 2023-11-08
feat: add universe domain support for VM cred

arithmetic1728:


clundin25 (approved):

google/auth/compute_engine/credentials.py
```diff
@@ -131,6 +132,14 @@ def service_account_email(self):
     def requires_scopes(self):
         return not self._scopes
 
+    @property
+    def universe_domain(self):
+        if self._universe_domain_cached:
+            return self._universe_domain
+        self._universe_domain = _metadata.get_universe_domain()
+        self._universe_domain_cached = True
```
TimurSadykov, line 140: We want to cache only definitive response, like 404 or a value. Do I understand correctly that this line will work only in those cases and others will throw TransportException?

arithmetic1728: yes, if it's not 404 or a value, then _metadata.get_universe_domain() will throw an exception at line 139, the caching logic is at line 140, so it won't be called.

TimurSadykov (approved):
LGTM

PR #1419, opened 2023-11-15
feat: Modify refresh window from 30s to 3 mins and 45 seconds

clundin25:
Modify the token refresh window. This will give a larger window for any token refreshes, giving retries and high latency environments more room to breathe.

The MDS stack that was previously a 30s window is now a 4 minute window.


arithmetic1728 (approved):

arithmetic1728:
the description above is a list of commit messages which look very confusing

clundin25:
> the description above is a list of commit messages which look very confusing

Thanks for catching that, I made a mistake when I created my branch.

PR #1433, opened 2023-12-06
fix: fix vm universe_domain bug

arithmetic1728:
missed a parameter for universe_domain request


google/auth/compute_engine/credentials.py
```diff
@@ -74,6 +75,7 @@ def __init__(
         self._scopes = scopes
         self._default_scopes = default_scopes
         self._universe_domain_cached = False
+        self._universe_domain_request = google_auth_requests.Request()
```
clundin25, line 78: Why does this request object get saved?

arithmetic1728: All req objects are passed in except the one used in the universe_domain property. I am thinking if we need to pass in a custom req object some day, we can override this _universe_domain_request property.

clundin25: Okay gotcha

clundin25 (approved):

PR #1437, opened 2023-12-08
fix: external account user cred universe domain support

arithmetic1728:


google/auth/external_account_authorized_user.py
```diff
@@ -297,6 +303,20 @@ def with_token_uri(self, token_uri):
         kwargs.update(token_url=token_uri)
         return self.__class__(**kwargs)
 
+    def with_universe_domain(self, universe_domain):
```
BigTailWolf, line 306: Shall we consider to move this method move to a common super class?

arithmetic1728: yes, created a common super class, thank you for your suggestion!

lsirac (approved):
Thanks, please hold off on submitting for now

clundin25 (approved):

Issue #1444, opened 2023-12-20
Getting Keyerrror _refresh_handler with old tokens


sompodsign:
**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.
![image](https://github.com/googleapis/google-auth-library-python/assets/47793716/24b218ae-5b14-45db-b94c-e3bff98bd6f3)

clundin25:
Hi @sompodsign,

Thank you for this issue. Do you have any more details on how the token got into this state?

sompodsign:
Thanks!. I can see the update on main branch. Could you please let me know how the release is work and when the fix is going to be released?

clundin25:
We have been in a release freeze since Dec 14. This freeze expires tonight, at 12am PST. We will cut a release tomorrow, on Jan 3rd.

Thanks!

clundin25:
Release v2.26.0 with the fix is now out! Feel free to re-open should the issue continue. Thanks!

PR #1445, opened 2023-12-20
fix: Guard delete statements. Add default fallback for _use_non_blocking_refresh.

clundin25:


clundin25:
Fixes https://github.com/googleapis/google-auth-library-python/issues/1444

westarle (approved):

Issue #1441, opened 2023-12-15
add underlying cause for `TransportError`


gchux:
https://github.com/googleapis/google-auth-library-python/blob/0afc61a3a9c3d5035c913ad4a7568e8a888e2ec9/google/auth/compute_engine/_metadata.py#L217-L220

Is it possible to save a reference to the last exception thrown? at 

https://github.com/googleapis/google-auth-library-python/blob/0afc61a3a9c3d5035c913ad4a7568e8a888e2ec9/google/auth/compute_engine/_metadata.py#L204

That way the underlying reason for the `TransportError` would be available even if `DEBUG` logging level is not enabled.

clundin25:
Should be a small and easy fix. Just add a variable store the last seen exception, and inherit it using the `from` syntax. 

https://docs.python.org/3/tutorial/errors.html#exception-chaining

PR #1446, opened 2023-12-21
Fix issue #1441

gchux:
Simplifying troubleshooting by capturing last seen exception 
to set it as the direct cause of the `exceptions.TransportError`


arithmetic1728:
could you add a unit test for this? thanks!

PR #1450, opened 2024-01-10
fix: read universe_domain for external account authorized user

arithmetic1728:


tests/data/external_account_authorized_user_non_gdu.json
```diff
@@ -0,0 +1,10 @@
+{
+    "type": "external_account_authorized_user",
+    "audience": "//iam.googleapis.com/locations/global/workforcePools/$WORKFORCE_POOL_ID/providers/$PROVIDER_ID",
```
lsirac, line 3: endpoints should should not be pointing to googleapis, but instead the value of universe_domain 

also missing new line eof

lsirac (approved):
LGTM, just make sure this doesn't cause any regressions with gcloud given the previous fix we submitted

tests/data/external_account_authorized_user_non_gdu.json
arithmetic1728: updated

google/auth/external_account_authorized_user.py
```diff
@@ -342,6 +342,7 @@ def from_info(cls, info, **kwargs):
             revoke_url=info.get("revoke_url"),
             quota_project_id=info.get("quota_project_id"),
             scopes=info.get("scopes"),
+            universe_domain=info.get("universe_domain"),
```
clundin25, line 345: Unrelated, but should all of these `get` alls be guarded or have a default? I am curious because I assume a stale file can cause a runtime exception.

clundin25 (approved):

google/auth/external_account_authorized_user.py
arithmetic1728: I think the required fields are checked in https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/external_account_authorized_user.py#L226-L229

I added a default to my change. Without adding it, it won't cause any issues because the constructor sets the default if universe_domain is None. But just added the default for the best.

PR #1463, opened 2024-02-05
feat: adding universe domain support for downscroped credentials

BigTailWolf:


google/auth/credentials.py
```diff
@@ -24,6 +24,8 @@
 from google.auth import metrics
 from google.auth._refresh_worker import RefreshThreadManager
 
+DEFAULT_UNIVERSE_DOMAIN = "googleapis.com"
```
lsirac, line 27: I think there are other instances of this, e.g. in external account. Can you consolidate? 

Also, "googleapis.com" is used below. Should be replaced there & elsewhere

tests/test_downscoped.py
```diff
@@ -496,6 +504,20 @@ def test_default_state(self):
         assert not credentials.expired
         # No quota project ID set.
         assert not credentials.quota_project_id
+        assert credentials.universe_domain == DEFAULT_UNIVERSE_DOMAIN
+
+    def test_create_with_customized_universe_domain(self):
```
lsirac, line 509: Can you add more tests? e.g. that request is sent to STS with the custom universe domain

google/auth/downscoped.py
```diff
@@ -467,7 +471,10 @@ def __init__(
         self._source_credentials = source_credentials
         self._credential_access_boundary = credential_access_boundary
         self._quota_project_id = quota_project_id
-        self._sts_client = sts.Client(_STS_TOKEN_URL)
+        self._universe_domain = universe_domain or credentials.DEFAULT_UNIVERSE_DOMAIN
```
lsirac, line 474: Can you check that the universe domain matches the one in source creds?

lsirac (changes requested):

google/auth/downscoped.py
```diff
@@ -437,7 +437,11 @@ class Credentials(credentials.CredentialsWithQuotaProject):
     """
 
     def __init__(
-        self, source_credentials, credential_access_boundary, quota_project_id=None
+        self,
+        source_credentials,
+        credential_access_boundary,
+        quota_project_id=None,
+        universe_domain=credentials.DEFAULT_UNIVERSE_DOMAIN,
```
arithmetic1728, line 444: add a doc string for `universe_domain` in the `Args` part below?

arithmetic1728 (approved):

google/auth/credentials.py
BigTailWolf: Great. In the beginning I wanted to do those refactoring together and I realized the PR is big.
Then i planned to open a PR only to refactor the `DEFAULT_UNIVERSE_DOMAIN` at once.

google/auth/downscoped.py
BigTailWolf: It is covered in the test L#507
```
        assert credentials.universe_domain == DEFAULT_UNIVERSE_DOMAIN
```

lsirac: What I mean here is if you pass source creds with universe domain A, but pass universe domain B here. We should error out in this case

BigTailWolf: I think the test supposed to be:
1. Implicit default. Pass nothing for universe domain and validate the universe domain as the default.
2. Explicitly pass None. And valid the universe domain is equals to default.
3. Explicitly pass a universe domain A, and validate the universe domain equals to A

Now working on the test above. Any missing validation?

tests/test_downscoped.py
```diff
@@ -506,6 +543,49 @@ def test_with_quota_project(self):
 
         assert quota_project_creds.quota_project_id == "project-foo"
 
+    @mock.patch("google.auth._helpers.utcnow", return_value=datetime.datetime.min)
+    def test_refresh_on_tpc_universe(self, unused_utcnow):
```
lsirac, line 547:
```suggestion
    def test_refresh_on_custom_universe(self, unused_utcnow):
```

lsirac (approved):

PR #1476, opened 2024-02-14
fix: make requests import conditional for gce universe domain

sai-sunder-s:
In [#1433](https://github.com/googleapis/google-auth-library-python/issues/1433) a dependency for requests was introduced unconditionally i.e. requests lib was required for anyone use gce mds even if they are not using TPC. This PR makes the import conditional.

### DIFF
--- a/google/auth/compute_engine/credentials.py
+++ b/google/auth/compute_engine/credentials.py
@@ -148,8 +148,9 @@ class Credentials(
     def universe_domain(self):
         if self._universe_domain_cached:
             return self._universe_domain
-        
+
         from google.auth.transport import requests as google_auth_requests
+
         self._universe_domain = _metadata.get_universe_domain(
             google_auth_requests.Request()
         )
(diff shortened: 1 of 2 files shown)


google/auth/compute_engine/credentials.py
```diff
@@ -84,7 +83,7 @@ def __init__(
         self._scopes = scopes
         self._default_scopes = default_scopes
         self._universe_domain_cached = False
-        self._universe_domain_request = google_auth_requests.Request()
+        self._universe_domain_request = None
```
clundin25, line 86: nit: Consider exposing the request as an optional parameter.

arithmetic1728 (approved):

arithmetic1728 (approved):

google/auth/compute_engine/credentials.py
sai-sunder-s: Removed the universe_domain_request completely per @arithmetic1728 's suggestion

Issue #1478, opened 2024-02-15
Breaking change when upgraded from 2.23.3 to 2.27.0


dzejeu:
## Error description
We use [luigi](https://github.com/spotify/luigi) in our project and after recent update of our python dependencies we started to receive following errors.
```
/usr/local/lib/python3.10/site-packages/luigi/worker.py  701 Luigi unexpected framework error while scheduling MyTask()
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/luigi/worker.py", line 796, in add
    current = queue.get()
  File "<string>", line 2, in get
  File "/usr/local/lib/python3.10/multiprocessing/managers.py", line 833, in _callmethod
    raise convert_to_error(kind, result)
multiprocessing.managers.RemoteError:
---------------------------------------------------------------------------
Unserializable message: Traceback (most recent call last):
  File "/usr/local/lib/python3.10/multiprocessing/managers.py", line 308, in serve_client
    send(msg)
  File "/usr/local/lib/python3.10/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/usr/local/lib/python3.10/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: cannot pickle '_thread.lock' object
```

## Workaround
I found out that dependency that caused it was `google-auth`. Bringing it down from `2.27.0` to `2.23.2` while keeping other dependencies updated solved the issue. It looks like this [PR](https://github.com/googleapis/google-auth-library-python/pull/1447) may be related to this behaviour.

## Background
We use luigi with `parallel_scheduling=true` option which means `multiprocessing.Queue` is used underneath and it requires every object in it to be pickle-able. 

## Environment details

  - OS: Debian 12 / Ubuntu 20.04 (same issue on both)
  - Python version: 3.10
  - pip version: 23.3.1
  - `google-auth` version: 2.27.0
  - `luigi` version: 3.5.0

clundin25:
Hi @dzejeu,

Can you share a repro?

I just tested on the latest commit in`main`, and I can pickle this code, so my repro is missing some details.

```
❯ git rev-parse HEAD && python -c 'import pickle; from google.auth._refresh_worker import RefreshThreadManager; q = pickle.dumps(RefreshThreadManager()); print(q); print(pickle.loads(q))'
30743b0937fe0c45c77c824514a0c27539f5745b
[...]

PR #1479, opened 2024-02-15
fix: Typo when setting the state for the pickle deserializer.

clundin25:


google/auth/_refresh_worker.py
```diff
@@ -75,7 +75,7 @@ def __getstate__(self):
 
     def __setstate__(self, state):
         """Pickle helper that deserializes the _lock attribute."""
-        state["_key"] = threading.Lock()
+        state["_lock"] = threading.Lock()
```
westarle, line 78: is there a test we could add?

clundin25: ACK - Going to wait to hear back on https://github.com/googleapis/google-auth-library-python/issues/1478 before doing more

westarle (approved):

tests/test__refresh_worker.py
```diff
@@ -150,7 +150,10 @@ def test_refresh_dead_worker():
 
 def test_pickle():
     w = _refresh_worker.RefreshThreadManager()
+    # For some reason isinstance cannot interpret threading.Lock as a type.
```
merwok, line 153: The reason is that threading.Lock is a function, returning instances of _thread.lock (internal class)

PR #1481, opened 2024-02-20
feat: Use float instead of int for GCE_METADATA_TIMEOUT

ludoch:
Move the logic to read GCE_METADATA_TIMEOUT env variable so we can better control it in app_identity component.

Related internal change is 607513254.


estrellis:
The change LGTM

google/auth/compute_engine/_metadata.py
```diff
@@ -119,6 +116,11 @@ def ping(request, timeout=_METADATA_DEFAULT_TIMEOUT, retry_count=3):
     #       could lead to false negatives in the event that we are on GCE, but
     #       the metadata resolution was particularly slow. The latter case is
     #       "unlikely".
+    try:
+        timeout = float(os.getenv("GCE_METADATA_TIMEOUT", timeout))
+    except ValueError:  # pragma: NO COVER
+        pass
```
clundin25, line 122: I would not capture a ValueError here. Additionally, the timeout variable would not be usable, so the downstream call would have problms.

estrellis: FWIW the original code caught the ValueError. The value is set by the user.
I am thinking if the Exception happens timeout will retain the passed in/default value

clundin25: In that case, can you remove the pragma, and include test coverage for this change?

google/auth/compute_engine/_metadata.py
```diff
@@ -53,10 +53,7 @@
 
 # Timeout in seconds to wait for the GCE metadata server when detecting the
 # GCE environment.
-try:
-    _METADATA_DEFAULT_TIMEOUT = int(os.getenv("GCE_METADATA_TIMEOUT", 3))
-except ValueError:  # pragma: NO COVER
-    _METADATA_DEFAULT_TIMEOUT = 3
+_METADATA_DEFAULT_TIMEOUT = 3
```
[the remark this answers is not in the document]
clundin25: Can you revert this change? It seems safe to keep as is.

google/auth/compute_engine/_metadata.py
```diff
@@ -119,6 +116,9 @@ def ping(request, timeout=_METADATA_DEFAULT_TIMEOUT, retry_count=3):
     #       could lead to false negatives in the event that we are on GCE, but
     #       the metadata resolution was particularly slow. The latter case is
     #       "unlikely".
+    if timeout is None:
+        timeout = float(os.getenv("GCE_METADATA_TIMEOUT", _METADATA_DEFAULT_TIMEOUT))
```
clundin25, line 120: Please add a test case, and then this LGTM.

PR #1484, opened 2024-02-28
feat: refactor AWS and identity pool credentials to use suppliers

aeitzman:


BigTailWolf (approved):

BigTailWolf:
Please fix lint

google/auth/aws.py
```diff
@@ -106,9 +106,9 @@ def get_request_options(
             Mapping[str, str]: The AWS signed request dictionary object.
         """
         # Get AWS credentials.
-        access_key = aws_security_credentials.get("access_key_id")
-        secret_key = aws_security_credentials.get("secret_access_key")
-        security_token = aws_security_credentials.get("security_token")
+        access_key = aws_security_credentials.accessKeyId
+        secret_key = aws_security_credentials.secretAccessKey
+        security_token = aws_security_credentials.sessionToken
```
clundin25, line 111: nit: Simplify the code by using the data class directly, instead of unrolling it into variables.

google/auth/aws.py
```diff
@@ -338,211 +338,141 @@ def _generate_authentication_header_map(
     return authentication_header
 
 
-class Credentials(external_account.Credentials):
-    """AWS external account credentials.
-    This is used to exchange serialized AWS signature v4 signed requests to
-    AWS STS GetCallerIdentity service for Google access tokens.
-    """
+class AwsSecurityCredentials(object):
+    """An object that models AWS security credentials with an optional session token."""
```
clundin25, line 342: This seems like a good candidate for a Dataclass since this codebase now uses at a minimum Python 3.7.

google/auth/aws.py
```diff
@@ -338,211 +338,141 @@ def _generate_authentication_header_map(
...
-        Retrieve the AWS region from the AWS_REGION or AWS_DEFAULT_REGION
-        environment variable or from the AWS metadata server availability-zone
-        if not found in the environment variable.
 
-        Check AWS credentials in environment variables. If not found, retrieve
-        from the AWS metadata server security-credentials endpoint.
+class AwsSecurityCredentialsSupplier(metaclass=abc.ABCMeta):
+    """Base class for AWS security credential suppliers. This can be implemented with custom logic to retrieve
+    AWS security credentials to exchange for a Google Cloud access token. The AWS external account credential does
+    not cache the AWS security credentials, so caching logic should be added in the implementation.
```
clundin25, line 388: nit: Maybe call out the caching in an Info or Warning block. I see it alluded to multiple times.

google/auth/aws.py
```diff
@@ -338,211 +338,141 @@ def _generate_authentication_header_map(
...
-            imdsv2_session_token = None
 
-        # Initialize the request signer if not yet initialized after determining
-        # the current AWS region.
-        if self._request_signer is None:
-            self._region = self._get_region(
-                request, self._region_url, imdsv2_session_token
-            )
-            self._request_signer = RequestSigner(self._region)
+class _InternalAwsSecurityCredentialsSupplier(AwsSecurityCredentialsSupplier):
```
clundin25, line 431: I think `Internal` is implied by the `_` convention. I'd prefer a name like `Default`

google/auth/aws.py
```diff
@@ -338,211 +338,141 @@ def _generate_authentication_header_map(
...
+        if env_aws_access_key_id and env_aws_secret_access_key:
+            return AwsSecurityCredentials(
+                env_aws_access_key_id, env_aws_secret_access_key, env_aws_session_token
+            )
 
-    def _get_region(self, request, url, imdsv2_session_token):
-        """Retrieves the current AWS region from either the AWS_REGION or
-        AWS_DEFAULT_REGION environment variable or from the AWS metadata server.
+        imdsv2_session_token = self._get_imdsv2_session_token(request)
+        # Get role name.
```
clundin25, line 460: nit: Avoid comments that describe "What". The variable names and method names are self-documenting

google/auth/aws.py
```diff
@@ -338,211 +338,141 @@ def _generate_authentication_header_map(
...
+        credentials = self._get_metadata_security_credentials(
+            request, role_name, imdsv2_session_token
+        )
 
-        Returns:
-            str: The current AWS region.
+        return AwsSecurityCredentials(
+            credentials.get("AccessKeyId"),
+            credentials.get("SecretAccessKey"),
+            credentials.get("Token"),
```
clundin25, line 471: Should there be any type of validation here? Is it guaranteed that the credential variable has these keys?

google/auth/aws.py
```diff
@@ -579,53 +510,23 @@ def _get_region(self, request, url, imdsv2_session_token):
...
-        env_aws_session_token = os.environ.get(environment_vars.AWS_SESSION_TOKEN)
-        if env_aws_access_key_id and env_aws_secret_access_key:
-            return {
-                "access_key_id": env_aws_access_key_id,
-                "secret_access_key": env_aws_secret_access_key,
-                "security_token": env_aws_session_token,
-            }
+    def _get_imdsv2_session_token(self, request):
+        if request is not None and self._imdsv2_session_token_url is not None:
+            headers = {"X-aws-ec2-metadata-token-ttl-seconds": "300"}
```
clundin25, line 515: Why 300? Can you make it a self-documenting variable.

google/auth/aws.py
```diff
@@ -579,53 +510,23 @@ def _get_region(self, request, url, imdsv2_session_token):
...
-        role_name = self._get_metadata_role_name(request, imdsv2_session_token)
+            imdsv2_session_token_response = request(
+                url=self._imdsv2_session_token_url, method="PUT", headers=headers
+            )
 
-        # Get security credentials.
-        credentials = self._get_metadata_security_credentials(
-            request, role_name, imdsv2_session_token
-        )
+            if imdsv2_session_token_response.status != 200:
```
clundin25, line 521: Use a http status object instead of a magic number. See `transport/__init__.py` for examples.

google/auth/aws.py
```diff
@@ -579,53 +510,23 @@ def _get_region(self, request, url, imdsv2_session_token):
...
+                raise exceptions.RefreshError(
+                    "Unable to retrieve AWS Session Token",
+                    imdsv2_session_token_response.data,
```
clundin25, line 524: Should this string be formatted?

google/auth/aws.py
```diff
@@ -722,24 +623,169 @@ def _get_metadata_role_name(self, request, imdsv2_session_token):
...
+        *args,
+        **kwargs
+    ):
+        """Instantiates an AWS workload external account credentials object.
+
+        Args:
+            audience (str): The STS audience field.
+            subject_token_type (str): The subject token type.
+            token_url (str): The STS endpoint URL.
+            credential_source (Mapping): The credential source dictionary used
```
clundin25, line 648: This documentation is missing how to build a credential source dict

google/auth/aws.py
```diff
@@ -722,24 +623,169 @@ def _get_metadata_role_name(self, request, imdsv2_session_token):
...
+        environment_id = credential_source.get("environment_id") or ""
+        self._aws_security_credentials_supplier = _InternalAwsSecurityCredentialsSupplier(
+            credential_source
+        )
+        self._cred_verification_url = credential_source.get(
+            "regional_cred_verification_url"
+        )
+
+        # Get the environment ID. Currently, only one version supported (v1).
+        matches = re.match(r"^(aws)([\d]+)$", environment_id)
```
clundin25, line 682: Add a comment with an example string that this regex is matching.

google/auth/aws.py
```diff
@@ -722,24 +623,169 @@ def _get_metadata_role_name(self, request, imdsv2_session_token):
...
+        if matches:
+            env_id, env_version = matches.groups()
+        else:
+            env_id, env_version = (None, None)
+
+        if env_id != "aws" or self._cred_verification_url is None:
+            raise exceptions.InvalidResource(
+                "No valid AWS 'credential_source' provided"
+            )
+        elif int(env_version or "") != 1:
```
clundin25, line 692: `int("")` will result in a `ValueError` being raised.

google/auth/aws.py
```diff
@@ -722,24 +623,169 @@ def _get_metadata_role_name(self, request, imdsv2_session_token):
...
+        # The full, canonical resource name of the workload identity pool
+        # provider, with or without the HTTPS prefix.
+        # Including this header as part of the signature is recommended to
+        # ensure data integrity.
+        request_headers["x-goog-cloud-target-resource"] = self._target_resource
+
+        # Serialize AWS signed request.
+        # Keeping inner keys in sorted order makes testing easier for Python
+        # versions <=3.5 as the stringified JSON string would have a predictable
+        # key order.
```
clundin25, line 775: Can the tests be rewritten to not rely on order? This does not seem like a good reason to sort the data.

google/auth/external_account.py
```diff
@@ -52,6 +52,43 @@
 _CLOUD_RESOURCE_MANAGER = "https://cloudresourcemanager.googleapis.com/v1/projects/"
 
 
+class SupplierContext(object):
```
clundin25, line 55: Looks like another candidate for a Dataclass

google/auth/external_account.py
```diff
@@ -52,6 +52,43 @@
...
+    """A context object that contains information about the requested third party credential that is passed
+    to AWS security credential and subject token suppliers."""
+
+    def __init__(self, subject_token_type, audience):
```
clundin25, line 59: Audience could default to None if it is optional

google/auth/identity_pool.py
```diff
@@ -47,6 +48,107 @@
...
+            raise exceptions.RefreshError(
+                "Unable to retrieve Identity Pool subject token", response_body
+            )
+        token_content = response_body, self._url
+        return _parse_token_data(
+            token_content, self._format_type, self._subject_token_field_name
+        )
+
+
+def _parse_token_data(token_content, format_type="text", subject_token_field_name=None):
```
clundin25, line 129: I would prefer a DataClass or NameTuple instead of an anonymous tuple. This code will be harder to refactor since all the Tuple code paths need to be traced manually.

aeitzman: changed to name tuple

google/auth/aws.py
aeitzman: done.

google/auth/external_account.py
aeitzman: Whoops, forgot to fix the comments here from the original code I took from a different part of the repo. Audience is not optional.

google/auth/aws.py
aeitzman: yes, the response we get back will have at least accessKeyId and SecretAccesskey, and token is optional. This code and the rest of the comments on the AWS credential aren't actually changed at all btw, just refactored and moved around a bit to support the supplier pattern we are using for programmatic. Since this isn't actually getting changed and won't cause a regression, are you OK if I pull all your suggestions into a to-do bug to go and do some of these fixes in a different PR targeted at that?

google/auth/aws.py
```diff
@@ -722,24 +623,169 @@ def _get_metadata_role_name(self, request, imdsv2_session_token):
...
+        subject_token_type,
+        token_url,
+        credential_source=None,
+        *args,
+        **kwargs
+    ):
+        """Instantiates an AWS workload external account credentials object.
+
+        Args:
+            audience (str): The STS audience field.
```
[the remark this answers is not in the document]
aeitzman: That would be a breaking change right? This is an existing credential type, and the "audience" parameter name is used everywhere for external account credentials.

google/auth/aws.py
clundin25: I am okay with you tracking non-regressions in a separate bug and following up with small PRs if you prefer.

google/auth/aws.py
```diff
@@ -338,211 +338,141 @@ def _generate_authentication_header_map(
...
-                to be exchanged for Google access tokens.
-            args (List): Optional positional arguments passed into the underlying :meth:`~external_account.Credentials.__init__` method.
-            kwargs (Mapping): Optional keyword arguments passed into the underlying :meth:`~external_account.Credentials.__init__` method.
+            accessKeyId (str): The AWS security credential access key ID.
+            secretAccessKey (str): The AWS security credential secret access key.
+            sessionToken (Optional[str]): The optional AWS session token. Used for temporary security credentials.
+        """
+        self._accessKeyId = accessKeyId
+        self._secretAccessKey = secretAccessKey
+        self._sessionToken = sessionToken
```
clundin25: Can we use snake_case instead of camelCase?

google/auth/aws.py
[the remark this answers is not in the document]
clundin25: I see, can this be hardcoded to the STS audience, or do users need to configure the audience?

Feel free to leave it as is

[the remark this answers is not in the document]
aeitzman: Users need to configure the audience, the audience is what tells us what workforce or workload pool provider the customer is using, there is not hard coded value to be used here.

google/auth/identity_pool.py
```diff
@@ -38,15 +38,122 @@
...
+        self._path = path
+        self._format_type = format_type
+        self._subject_token_field_name = subject_token_field_name
+
+    @_helpers.copy_docstring(SubjectTokenSupplier)
+    def get_subject_token(self, context, request):
+        if not os.path.exists(self._path):
+            raise exceptions.RefreshError("File '{}' was not found.".format(self._path))
+
+        with io.open(self._path, "r", encoding="utf-8") as file_obj:
```
clundin25, line 97: nit: open is a builtin, you don't need to use the io module.

google/auth/identity_pool.py
```diff
@@ -38,15 +38,122 @@
...
+            google.auth.exceptions.RefreshError: If an error is encountered during
+                subject token retrieval logic.
+
+        Returns:
+            str: The requested subject token string.
+        """
+        raise NotImplementedError("")
+
+
+class _TokenContent(NamedTuple):
```
clundin25, line 79: Please add a doc string. Location is particularly tricky, since it may be a file path or a URL.

google/auth/aws.py
clundin25: I would like to follow up on this. It would be best to avoid sorting this data twice.

google/auth/aws.py
clundin25: Ping

google/auth/aws.py
clundin25: Ping

google/auth/aws.py
```diff
@@ -338,211 +335,111 @@ def _generate_authentication_header_map(
...
-        self,
-        audience,
-        subject_token_type,
-        token_url,
-        credential_source=None,
-        *args,
-        **kwargs
-    ):
-        """Instantiates an AWS workload external account credentials object.
+    accessKeyId: str
```
clundin25, line 348: Avoid camel case, in this case the style guide suggests snake case

clundin25 (approved):
Additional follow up tracked in b/328671918

Issue #1426, opened 2023-12-01
I can't load any package


sabaadmin:
#### Environment details

  - OS:windows11
  - Python version:3.12.0
  - pip version:23.3.1
  - `google-auth` version:2.24.0

#### Steps to reproduce

  1. Install google-auth
  2. Try import module

![image](https://github.com/googleapis/google-auth-library-python/assets/94890302/c4523a62-7a16-41f0-9d59-b0ef5f5ad8bc)
There is a file in the pip directory but I can't load it

parthea:
Can you please share the output after running `which python`?

Please also try running `python -m pip install google-auth`.

Please also check if the path that you see after running `which python` appears in in the output of this command
```
import sys
sys.path
```

sabaadmin:
![image](https://github.com/googleapis/google-auth-library-python/assets/94890302/59ae5beb-5122-4d2f-9814-8f752337b292)
which python gives an error

parthea:
Sorry about that. `where python` may work , as `which python` is for unix.

Please also try running
```
python -c "import os, sys; print(os.path.dirname(sys.executable))"
```

parthea:
@clundin25 Do you have availability to reproduce the issue in Windows Power Shell?

clundin25:
I was not able to reproduce.

```
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

Try the new cross-platform PowerShell https://aka.ms/pscore6

PS C:\Users\clundin> python --version
Python 3.12.0
PS C:\Users\clundin> python -m pip install google-auth
Collecting google-auth
  Downloading google_auth-2.24.0-py2.py3-none-any.whl.metadata (4.7 kB)
Collecting cachetools<6.0,>=2.0.0 (from google-auth)
  Obtaining dependency information for cachetools<6.0,>=2.0.0 from https://files.pythonhosted.org/packages/a2/91/2d843adb9fbd911e0da45fbf6f18ca89d07a087c3daa23e955584f90ebf4/cachetools-5.3.2-py3-none-any.whl.metadata
  Downloading cachetools-5.3.2-py3-none-any.whl.metadata (5.2 kB)
Collecting pyasn1-modules>=0.2.1 (from google-auth)
  Downloading pyasn1_modules-0.3.0-py2.py3-none-any.whl (181 kB)
[...]

PR #1487, opened 2024-02-29
feat: Make init.py to can load google.auth and google.outh2 with import google

CureSaba:
For fix this issue https://github.com/googleapis/google-auth-library-python/issues/1426
You can load google.auth and google.outh2 with import google

### DIFF
--- a/google/__init__.py
+++ b/google/__init__.py
@@ -1,0 +1,2 @@
+import google.auth
+import google.oauth2


parthea (changes requested):
If this change is accepted I believe we could run into issues like https://github.com/googleapis/google-cloud-python/issues/11399 where there is a conflict with other packages that use the `google` namespace.

clundin25:
I don't see any benefits to introducing this change.

CureSaba:
> If this change is accepted I believe we could run into issues like [googleapis/google-cloud-python#11399](https://github.com/googleapis/google-cloud-python/issues/11399) where there is a conflict with other packages that use the `google` namespace.

CureSaba:
This pull request has been stopped to avoid conflicts with other libraries that use google mamespace.

PR #1494, opened 2024-03-11
feat: Adds support for custom suppliers in AWS and Identity Pool credential instantiation

aeitzman:


BigTailWolf (approved):
LGTM

google/auth/aws.py
```diff
@@ -637,33 +661,54 @@ def __init__(
...
-        # Get the environment ID. Currently, only one version supported (v1).
-        matches = re.match(r"^(aws)([\d]+)$", environment_id)
-        if matches:
-            env_id, env_version = matches.groups()
+        if aws_security_credentials_supplier:
+            self._aws_security_credentials_supplier = aws_security_credentials_supplier
+            self._cred_verification_url = (
+                _DEFAULT_AWS_REGIONAL_CREDENTIAL_VERIFICATION_URL
+            )
+            _LOGGER.info(
```
clundin25, line 681: I think this qualifies as a debug level log.

google/auth/aws.py
```diff
@@ -637,33 +661,54 @@ def __init__(
...
+            if matches:
+                env_id, env_version = matches.groups()
+            else:
+                env_id, env_version = (None, None)
+
+            if env_id != "aws" or self._cred_verification_url is None:
+                raise exceptions.InvalidResource(
+                    "No valid AWS 'credential_source' provided"
+                )
+            elif int(env_version or "") != 1:
```
clundin25, line 704: This still has the ValueError from the previous PR?

google/auth/aws.py
aeitzman: I added this based off your previous comment - https://github.com/googleapis/google-auth-library-python/pull/1484#discussion_r1508223845, I may have not understood what you are asking for

google/auth/aws.py
aeitzman: This was part of the code that was just moved around that I was going to handle in b/328671918, not an actual change/regression made by this feature

google/auth/aws.py
clundin25: Sorry my comment was vague - I meant a docstring block like [this](https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/credentials.py#L383) to visually elevate this info in the documentation.

google/auth/aws.py
clundin25: Got it - so there will be a PR after this

google/auth/aws.py
aeitzman: Ahh, makes sense, I will just remove the logging blocks then and add that instead - thanks

google/auth/aws.py
aeitzman: Yeah, first PR was doing a refactor to support suppliers, this PR adds support for users to create credentials using their own suppliers, then there will be a follow up PR that fixes the older issues you pointed out in the previous PR. Trying to keep the individual PR's smaller is possible since this is such a big change

google/auth/aws.py
```diff
@@ -603,20 +610,37 @@ def __init__(
...
+                to be exchanged for Google access tokens. Either a credential source or an AWS security
+                credentials supplier must be provided.
+                Example credential_source for AWS credential::
+
+                    {
+                        "environment_id": "aws1",
+                        "regional_cred_verification_url": "https://sts.{region}.amazonaws.com?Action=GetCallerIdentity&Version=2011-06-15",
+                        "region_url": "http://169.254.169.254/latest/meta-data/placement/availability-zone",
+                        "url": "http://169.254.169.254/latest/meta-data/iam/security-credentials",
+                        imdsv2_session_token_url": "http://169.254.169.254/latest/api/token"
```
clundin25, line 638: Are these fields documented anywhere? Maybe a link to the credential_source type could be useful?

google/auth/aws.py
```diff
@@ -603,20 +610,37 @@ def __init__(
...
+                    }
+            aws_security_credentials_supplier (AwsSecurityCredentialsSupplier): Optional AWS security credentials supplier.
```
clundin25, line 640: Should this be a union type?

clundin25: Also should be optional

google/auth/aws.py
aeitzman: Unfortunately, the credential_source has always just been an generic map instead of a defined credential_source type in this library, which makes instantiating the credentials in code kind of weird. I think it would be valuable to add typing for that, but that isn't really in the scope of this PR. Also for any credential that uses credential_source, we would just expect users to build it from the ADC file anyway, which is why I think there wasn't much effort put into having the credentials easily create able from code in the past.

google/auth/external_account.py
```diff
@@ -444,6 +452,14 @@ def with_universe_domain(self, universe_domain):
         new_cred._metrics_options = self._metrics_options
         return new_cred
 
+    def _should_initialize_impersonated_credentials(self):
+        if (
+            self._service_account_impersonation_url
+            and not self._impersonated_credentials
```
clundin25, line 458:
```suggestion
            and self._impersonated_credentials is None
```

Same nit as above

google/auth/external_account.py
```diff
@@ -444,6 +452,14 @@ def with_universe_domain(self, universe_domain):
...
+        ):
+            return True
+        return False
```
clundin25, line 461:
```suggestion
            return self._service_account_impersonation_url is not None and self._impersonated_credentials is None
```

google/auth/identity_pool.py
```diff
@@ -219,10 +232,28 @@ def __init__(
...
+            )
+        if credential_source is not None and subject_token_supplier is not None:
+            raise exceptions.InvalidValue(
+                "Identity pool credential cannot have both a credential source and a subject token supplier."
+            )
+
+        if subject_token_supplier is not None:
+            self._subject_token_supplier = subject_token_supplier
+            _LOGGER.info(
+                "Subject token supplier is being used. Note that the external account credential does not cache the subject token so caching should be implemented in the supplier."
```
clundin25, line 247: Same here, I think this could be a doc string note or warning (Sorry for the misunderstanding).

google/auth/identity_pool.py
```diff
@@ -295,8 +326,20 @@ def _create_default_metrics_options(self):
                 metrics_options["source"] = "file"
             else:
                 metrics_options["source"] = "url"
+        else:
+            metrics_options["source"] = "programmatic"
         return metrics_options
 
+    def _has_custom_supplier(self):
+        return not self._credential_source
```
clundin25, line 334:
```suggestion
        return self._credential_source is None
```

nit: Style guide

clundin25 (approved):

google/auth/aws.py
clundin25: Misunderstood the doc string, but I suppose `credential_source` can be overloaded to use this type as well, to achieve the mutual exclusion.

Not a big deal, marking the type as Optional is enough to close this.

google/auth/aws.py
clundin25: Okay - Would be nice to see these fields documented a bit more but feel free to close this.

PR #1501, opened 2024-03-18
fix: refactor tech debt in aws and identity pool credentials

aeitzman:
implements fixes suggested in #1484


google/auth/aws.py
```diff
@@ -690,7 +695,7 @@ def __init__(
                 "regional_cred_verification_url"
             )
 
-            # Get the environment ID. Currently, only one version supported (v1).
+            # Get the environment ID, i.e. "aws1". Currently, only one version supported (1).
```
BigTailWolf, line 698: Is this change from `v1` to `1` just want to clarify we are using the `env_version` literally `1`?

BigTailWolf (commented):
Just a nit question

google/auth/aws.py
```diff
@@ -69,6 +69,8 @@
 _DEFAULT_AWS_REGIONAL_CREDENTIAL_VERIFICATION_URL = (
     "https://sts.{region}.amazonaws.com?Action=GetCallerIdentity&Version=2011-06-15"
 )
+# IMDSV2 session token lifetime. This is set to a low value because the session token is used immediately.
+_IMDSV2_SESSION_TOKEN_TTL = "300"
```
clundin25, line 73: Can you include the unit as well?
```suggestion
_IMDSV2_SESSION_TOKEN_TTL _SECONDS= "300"
```

clundin25 (approved):

google/auth/aws.py
aeitzman: Correct, the value we expect for environment ID is "aws1"

PR #1506, opened 2024-03-28
fix: fix id_token iam endpoint for non-gdu service credentials

arithmetic1728:
The iam id token endpoint should use the given universe domain, instead of googleapis.com


google/auth/iam.py
```diff
@@ -67,7 +82,7 @@ def _make_signing_request(self, message):
         message = _helpers.to_bytes(message)
 
         method = "POST"
-        url = _SIGN_BLOB_URI.format(self._service_account_email)
+        url = _IAM_SIGN_ENDPOINT.format(self._service_account_email) + "?alt=json"
```
clundin25, line 85: Was the missing `?alt=json` a bug?

clundin25 (approved):

google/auth/iam.py
BigTailWolf: Shall we move the `?alt=json` part to the `_IAM_SIGN_ENDPOINT`

TimurSadykov (commented):
LGTM

PR #1514, opened 2024-04-10
fix: makes default token_url universe aware

aeitzman:
Also adds tests for credentials that use the default token_url


google/auth/external_account.py
```diff
@@ -147,7 +147,12 @@ def __init__(
         super(Credentials, self).__init__()
         self._audience = audience
         self._subject_token_type = subject_token_type
+        self._universe_domain = universe_domain or credentials.DEFAULT_UNIVERSE_DOMAIN
```
clundin25, line 150: Is this already available from the parent class?

google/auth/external_account.py
```diff
@@ -52,7 +52,7 @@
 # Cloud resource manager URL used to retrieve project information.
 _CLOUD_RESOURCE_MANAGER = "https://cloudresourcemanager.googleapis.com/v1/projects/"
 # Default Google sts token url.
-_DEFAULT_TOKEN_URL = "https://sts.googleapis.com/v1/token"
+_DEFAULT_TOKEN_URL = "https://sts.{universe_domain}/v1/token"
```
clundin25, line 55: This is fine for now but if the URL continues to get complex it should be wrapped in a type.

clundin25 (approved):

google/auth/external_account.py
aeitzman: SG, should not get more complex than this, but will make that change if it does.

google/auth/external_account.py
aeitzman: The defaulting is, but it doesn't set the universe_domain from the constructor.

PR #1519, opened 2024-05-01
fix: Add exception handler when a linux device is missing a security key but a reauth challenge was issued for a key.

clundin25:


arithmetic1728 (approved):

clundin25:
Opting to fix this in gcloud instead.

PR #1527, opened 2024-05-23
feat: adds X509 workload cert logic

aeitzman:
Adds support for X509 workload logic that reads from the certificate_config.json file. This PR does not actually call that logic yet - hooks for actually using the new logic will come when the credential type is added.


google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
     return metadata
```
andyrzhao, line 87: nit: maybe "jsondata" instead of "metadata"

google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
...
 
 
+def _get_workload_x509(certificate_config_path=None):
```
andyrzhao, line 90: nit: _get_workload_cert

google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
...
+    workload = data["cert_configs"]["workload"]
+
+    cert_path = workload["cert_path"]
+    key_path = workload["key_path"]
+
+    return _read_cert_and_key_files(cert_path, key_path)
+
+
+def _get_cert_config_path(certificate_config_path=None):
+    """Gets the certificate configuration full path. If an override is provided,
```
andyrzhao, line 121: Current comment suggests that if override is provided and does not exist, we will fall back to the other 2 options, but the code actually returns "None" in this case. Let's update the comment for clarity:
"""Gets the certificate configuration full path using the following order of precedence:
1. Explicit override, if set
2. Environment variable, if set
3. Well-known location

Returns "None" if the selected config file does not exist.
...

andyrzhao (commented):
LGTM w.r.t business logic.

lsirac (approved):

google/auth/transport/_mtls_helper.py
```diff
@@ -16,13 +16,15 @@
...
 import logging
-from os import path
+from os import environ, path
 import re
 import subprocess
 
 from google.auth import exceptions
 
 CONTEXT_AWARE_METADATA_PATH = "~/.secureConnect/context_aware_metadata.json"
+_CERTIFICATE_CONFIGURATION_DEFAULT_PATH = "~/.config/gcloud/certificate_config.json"
```
clundin25, line 26: Will this work on Windows?

google/auth/transport/_mtls_helper.py
```diff
@@ -63,20 +65,20 @@ def _check_dca_metadata_path(metadata_path):
     return metadata_path
 
 
-def _read_dca_metadata_file(metadata_path):
-    """Loads context aware metadata from the given path.
+def _read_json_file(path):
```
clundin25, line 68: I think this function's name is too vague. Can you rename it to provide more context?

google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
...
 
 
+def _get_workload_x509(certificate_config_path=None):
+    """Read the cert and key files specified in the certificate config provided.
+    If no config path is provided, check the environment variable and well known
+    gcloud location.
+
+    Args:
+        certificate_config_path (string): The certificate config path. If no path is provided,
+        the well known gcloud location will be used.
```
clundin25, line 97: nit: The env variable will be checked as well?

google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
...
+        Tuple[bytes, bytes]: client certificate bytes in PEM format and key
+            bytes in PEM format.
+
+    Raises:
+        google.auth.exceptions.ClientCertError: if problems occurs when retrieving
+        the certificate or key information.
+    """
+    absolute_path = _get_cert_config_path(certificate_config_path)
+    if absolute_path is None:
+        return None, None
```
clundin25, line 109: Should this be an exception?

google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
...
+    If no config path is provided, check the environment variable and well known
+    gcloud location.
+
+    Args:
+        certificate_config_path (string): The certificate config path. If no path is provided,
+        the well known gcloud location will be used.
+
+    Returns:
+        Tuple[bytes, bytes]: client certificate bytes in PEM format and key
+            bytes in PEM format.
```
clundin25, line 101: The function seems to actually return a `Tuple[Optional[bytes], Optional[bytes]]`

google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
...
+    Raises:
+        google.auth.exceptions.ClientCertError: if problems occurs when retrieving
+        the certificate or key information.
+    """
+    absolute_path = _get_cert_config_path(certificate_config_path)
+    if absolute_path is None:
+        return None, None
+    data = _read_json_file(certificate_config_path)
+
+    workload = data["cert_configs"]["workload"]
```
clundin25, line 112: This JSON needs some validation, otherwise there is a risk of uncaught key errors

google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
...
+    return cert_data, key_data
+
+
+def _read_cert_file(cert_path):
+    with open(cert_path, "rb") as cert_file:
+        cert_data = cert_file.read()
+
+    cert_match = re.findall(_CERT_REGEX, cert_data)
+    if len(cert_match) != 1:
+        raise exceptions.ClientCertError("Certificate file is invalid")
```
clundin25, line 159: The exception should include the cert_path and possibly some debugging steps.

google/auth/transport/_mtls_helper.py
```diff
@@ -85,6 +87,90 @@ def _read_dca_metadata_file(metadata_path):
...
+    return cert_match[0]
+
+
+def _read_key_file(key_path):
+    with open(key_path, "rb") as key_file:
+        key_data = key_file.read()
+
+    key_match = re.findall(_KEY_REGEX, key_data)
+    if len(key_match) != 1:
+        raise exceptions.ClientCertError("Private key file is invalid")
```
clundin25, line 169: Same here

google/auth/transport/_mtls_helper.py
aeitzman: renamed to load Json file, but that's not much better. Not sure what would be more descriptive, this method does simply just read a file and calls Json.load while wrapping in the appropriate error if it fails - open to suggestions

google/auth/transport/_mtls_helper.py
aeitzman: I don't think we want to throw here, since generally if we don't find a valid file we would want the calling method to just call the next certificate method (i.e. ECP/secure connect) - will revisit when this is actually called as part of the cert logic

google/auth/transport/_mtls_helper.py
andyrzhao: With the changes in this PR, this function is used to load json configs for both secure connect as well as workload, hence the generic naming. I think it's fine as is - maybe add a function comment to clarify usage. It does seem more suitable in a generic util file, but I don't have a strong opinion here.

google/auth/transport/_mtls_helper.py
clundin25: Sounds good.

clundin25 (approved):

andyrzhao (approved):

BigTailWolf (approved):

google/auth/transport/_mtls_helper.py
aeitzman: added another sentence in the comment explaining the method is used for both x509 and secure connect, this definitely could be moved to a util file in the future though, just didn't think that was necessary for this PR.

PR #1529, opened 2024-06-03
feat: add support for async rest

ohmayr:


google/auth/aio/_credentials_async.py
```diff
@@ -0,0 +1,171 @@
...
+
+"""Interfaces for credentials."""
+
+import abc
+import inspect
+
+from google.auth import credentials
+
+
+class Credentials(credentials.Credentials, metaclass=abc.ABCMeta):
```
clundin25, line 24: IMO, inheriting from the sync classes is a poor design choice and will make the async code unwieldy.

PR #1541, opened 2024-06-20
feat: adds support for X509 workload credential type

aeitzman:


google/auth/external_account.py
```diff
@@ -40,6 +41,7 @@
 from google.auth import exceptions
 from google.auth import impersonated_credentials
 from google.auth import metrics
+from google.auth.transport.requests import _MutualTlsAdapter
```
clundin25, line 44: Is it possible to keep this self-contained to the transport module? (I don't see this being used)

google/auth/identity_pool.py
```diff
@@ -255,76 +265,93 @@ def __init__(
...
+                sum(
+                    map(
+                        bool,
+                        [
+                            self._credential_source_file,
+                            self._credential_source_url,
+                            self._credential_source_certificate,
+                        ],
                     )
                 )
```
clundin25, line 288: This should be refactored to a simpler helper function.

google/auth/identity_pool.py
```diff
@@ -255,76 +265,93 @@ def __init__(
...
             self._credential_source_file = credential_source.get("file")
             self._credential_source_url = credential_source.get("url")
```
clundin25, line 267: Is it possible to use a data class for each credential types parameters? There is a lot of horizontal parameters / invariants now.

google/auth/external_account.py
```diff
@@ -393,12 +395,18 @@ def get_project_id(self, request):
     @_helpers.copy_docstring(credentials.Credentials)
     def refresh(self, request):
         scopes = self._scopes if self._scopes is not None else self._default_scopes
+        auth_request = request
+
+        # if mtls is required, wrap the incoming request in a partial to set the cert.
+        if self._should_add_mtls():
+            print("mtls yeah")
```
clundin25, line 402: looks like this line should be deleted.

aeitzman: whoops, thanks

google/auth/identity_pool.py
aeitzman: Are you OK if that is done in a separate PR?

clundin25: Sounds good 👍

tests/test_external_account.py
```diff
@@ -665,6 +677,56 @@ def test_refresh_without_client_auth_success(
...
+            "requested_token_type": "urn:ietf:params:oauth:token-type:access_token",
+            "subject_token": "subject_token_0",
+            "subject_token_type": self.SUBJECT_TOKEN_TYPE,
+        }
+        request = self.make_mock_request(status=http_client.OK, data=response)
+        credentials = self.make_credentials()
+
+        credentials.refresh(request)
+
+        expected_cert_path = ("test1", "test2")
```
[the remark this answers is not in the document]
aeitzman: Fixed to real values, its a tuple since it returns both a cert and private key file location (should be more self explanatory now since I changed the function name)

google/auth/external_account.py
```diff
@@ -393,12 +394,19 @@ def get_project_id(self, request):
     @_helpers.copy_docstring(credentials.Credentials)
     def refresh(self, request):
         scopes = self._scopes if self._scopes is not None else self._default_scopes
+        auth_request = request
```
clundin25, line 397: nit: I don't think it is necessary to create a new variable.

google/auth/external_account.py
```diff
@@ -393,12 +394,19 @@ def get_project_id(self, request):
...
+
+        # If mtls is required, wrap the incoming request in a partial to set the cert.
```
clundin25, line 399: nit:
```suggestion
        # Inject client certificate into request.
```

google/auth/external_account.py
```diff
@@ -523,6 +531,31 @@ def _create_default_metrics_options(self):
 
         return metrics_options
 
+    def _should_add_mtls(self):
```
clundin25, line 534: nit:

```suggestion
    def _mtls_required(self):
```

google/auth/external_account.py
```diff
@@ -523,6 +531,31 @@ def _create_default_metrics_options(self):
...
+        X509 workload identity federation.
+
+        Returns:
+            Tuple[str, str]: The cert and key file locations as strings in a tuple.
+
+        Raises:
+            NotImplementedError: When the current credential is not configured for
+                mTLS.
+        """
+        raise NotImplementedError("_get_mtls_cert must be implemented.")
```
clundin25, line 557:
```suggestion
        raise NotImplementedError("_get_mtls_cert_and_key_location must be implemented.")
```

google/auth/external_account.py
```diff
@@ -523,6 +531,31 @@ def _create_default_metrics_options(self):
...
+        """Returns a boolean representing whether the current credential is configured
+        for mTLS and should add a certificate to the outgoing calls to the sts and service
+        account impersonation endpoint.
+
+        Returns:
+            bool: True if the credential is configured for mTLS, False if it is not.
+        """
+        return False
+
+    def _get_mtls_cert_and_key_location(self):
```
clundin25, line 544: nit:

I wonder if replacing location with file path is appropriate?

google/auth/external_account.py
```diff
@@ -523,6 +531,31 @@ def _create_default_metrics_options(self):
...
+        """Gets the file locations for a certificate and private key file
+        to be used for configuring mTLS for the sts and service account
+        impersonation calls. Currently only expected to return a value when using
+        X509 workload identity federation.
+
+        Returns:
+            Tuple[str, str]: The cert and key file locations as strings in a tuple.
```
clundin25, line 551: I think this can be a follow up but I think in the future we may want to refactor this and the transport layer to wrap the cert and key as new sources get added.

google/auth/identity_pool.py
```diff
@@ -255,76 +265,70 @@ def __init__(
...
                 )
-            else:
+            elif self._credential_source_url:
                 self._subject_token_supplier = _UrlSupplier(
                     self._credential_source_url,
                     self._credential_source_format_type,
                     self._credential_source_field_name,
                     self._credential_source_headers,
                 )
+            else:
```
clundin25, line 298:
```suggestion
            else: # self._credential_source_certificate
```

google/auth/identity_pool.py
```diff
@@ -339,6 +343,76 @@ def _constructor_args(self):
             args.update({"subject_token_supplier": self._subject_token_supplier})
         return args
 
+    def _validate_certificate_credential_source(self):
```
clundin25, line 346: nit:
```suggestion
    def _validate_certificate_config(self):
```

google/auth/identity_pool.py
```diff
@@ -339,6 +343,76 @@ def _constructor_args(self):
...
+        if self._certificate_config_location:
+            if use_default:
+                raise exceptions.MalformedError(
+                    "Invalid certificate configuration, certificate_config_location cannot be specified when use_default_certificate_config = true."
+                )
+        else:
+            if not use_default:
+                raise exceptions.MalformedError(
+                    "Invalid certificate configuration, use_default_certificate_config should be true if no certificate_config_location is provided."
+                )
```
clundin25, line 362:
```suggestion
        if self._certificate_config_location and use_default:
                raise exceptions.MalformedError(
                    "Invalid certificate configuration, certificate_config_location cannot be specified when use_default_certificate_config = true."
                )
        if not self._certificate_config_location and not use_default:
                raise exceptions.MalformedError(
                    "Invalid certificate configuration, use_default_certificate_config should be true if no certificate_config_location is provided."
                )
```

clundin25: nit

google/auth/identity_pool.py
```diff
@@ -339,6 +343,76 @@ def _constructor_args(self):
...
+
+    def _validate_file_url_credential_source(self, credential_source):
```
clundin25, line 364: nit: 
```suggestion
    def _validate_url_config(self, credential_source):
```

google/auth/identity_pool.py
```diff
@@ -339,6 +343,76 @@ def _constructor_args(self):
...
+        self._credential_source_headers = credential_source.get("headers")
```
clundin25, line 365: Is this assignment intentional? Seems this function does more than validation.

google/auth/identity_pool.py
```diff
@@ -339,6 +343,76 @@ def _constructor_args(self):
...
+                "subject_token_field_name"
+            )
+            if self._credential_source_field_name is None:
+                raise exceptions.MalformedError(
+                    "Missing subject_token_field_name for JSON credential_source format"
+                )
+        else:
+            self._credential_source_field_name = None
+
+    def _validate_only_one_source(self):
```
clundin25, line 390:
```suggestion
    def _validate_single_source(self):
```

google/auth/identity_pool.py
```diff
@@ -339,6 +343,76 @@ def _constructor_args(self):
...
+            > 1
+        ):
+            raise exceptions.MalformedError(
+                "Ambiguous credential_source. 'file', 'url', and 'certificate' are mutually exclusive.."
+            )
+        if (
+            not self._credential_source_file
+            and not self._credential_source_url
+            and not self._credential_source_certificate
+        ):
```
clundin25, line 411: nite:

```suggestion
        credential_sources = [self._credential_source_file, self._credential_source_url, self._credential_source_certificate]
        valid_credential_sources = list(filter(lambda source: source is not None, credential_sources))
        
        if len(valid_credential_sources) > 1: 
            raise exceptions.MalformedError(
                "Ambiguous credential_source. 'file', 'url', and 'certificate' are mutually exclusive.."
            )
        if len(valid_credential_sources) != 1:
```

google/auth/transport/_mtls_helper.py
```diff
@@ -105,9 +105,50 @@ def _get_workload_cert_and_key(certificate_config_path=None):
...
+    cert_path, key_path = _get_workload_cert_and_key_paths(certificate_config_path)
+
+    if cert_path is None and key_path is None:
+        return None, None
+
+    return _read_cert_and_key_files(cert_path, key_path)
+
+
+def _get_cert_config_path(certificate_config_path=None):
+    """Gets the certificate configuration full path using the following order of precedence:
```
clundin25, line 118: nit:
```suggestion
    """Get the certificate configuration path based on the following order:
```

clundin25 (approved):

google/auth/external_account.py
aeitzman: If we just set modify the input request, won't that change the behavior of the request instance after the refresh is called with x509? i.e. if someone is re-using the same request for two credential refresh calls, one for X509, and one for file based

PR #1551, opened 2024-07-05
feat: implement base classes for credentials and request sessions

ohmayr:
This PR implements base classes for credentials and request sessions to capture the common logic between synchronous and asynchronous code.


google/auth/transport/_requests_base.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2016 Google LLC
```
vchudnov-g, line 1:
```suggestion
# Copyright 2024 Google LLC
```

google/auth/_credentials_base.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Args:
+            headers (Mapping): The HTTP request headers.
+            token (Optional[str]): If specified, overrides the current access
+                token.
+        """
+        headers["authorization"] = "Bearer {}".format(
+            _helpers.from_bytes(token or self.token)
+        )
+        """Trust boundary value will be a cached value from global lookup.
```
vchudnov-g, line 175: I realize this was in the original file, but it's weird having a string floating as a comment rather than a proper Python comment.

google/auth/_credentials_base.py
```diff
@@ -0,0 +1,204 @@
...
+        """Optional[str]: The universe domain value, default is googleapis.com
+        """
+
+        self._use_non_blocking_refresh = False
+        # TODO(omairn): RefreshThreadManager() may be specific to sync refresh and may need to be separately implemented for async. If so,
+        # move it back to credentials.py
+        self._refresh_worker = RefreshThreadManager()
+
+    @property
+    # TODO(omairn): We could alternatively not implement this here and use asyncio.get_event_loop().time() for the case of async
```
vchudnov-g, line 70: Use your GitHJub username in TODOs....

vchudnov-g (approved):
LGTM, but get sign-off from auth team.

Quick observation: When refactoring file contents into a new file but still keeping the old file, it is helpful to reviewers if you can make the new file share the old file's previous history, so reviewers can see the diff's between the old file and the new as well as between the the old file's previous version and new version. Instructions are here: https://stackoverflow.com/a/44036771/23828460 (also see https://www.mankier.com/1/git-cp#Synopsis)

No need to do it now after the fact, but keep it in mind for future refactorings.

google/auth/transport/requests.py
```diff
@@ -588,12 +589,12 @@ def request(
...
-
     def close(self):
         if self._auth_request_session is not None:
             self._auth_request_session.close()
         super(AuthorizedSession, self).close()
+
+    @property
+    def is_mtls(self):
+        """Indicates if the created SSL channel is mutual TLS."""
+        return self._is_mtls
```
clundin25, line 600: nit: Try to keep whitespace changes out of the diff.

google/auth/transport/_requests_base.py
```diff
@@ -0,0 +1,84 @@
...
+import os
+
+import abc
+from google.auth import transport
+
+
+_DEFAULT_TIMEOUT = 120  # in second
+
+
+class _BaseAuthorizedSession(metaclass=abc.ABCMeta):
```
clundin25, line 26: My thoughts are similar here re:codesharing. Since the async code is a "colored" [function](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/) I think it's safe to assume that the sync code and async code should never be in the same callstack.

google/auth/_credentials_base.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        if not self.expiry:
+            return False
+        # Remove some threshold from expiry to err on the side of reporting
+        # expiration early so that we avoid the 401-refresh-retry loop.
+        skewed_expiry = self.expiry - _helpers.REFRESH_THRESHOLD
+        return _helpers.utcnow() >= skewed_expiry
+
+    @property
+    def valid(self):
```
clundin25, line 90: Similar comment here.

google/auth/_credentials_base.py
```diff
@@ -0,0 +1,204 @@
...
+
+        self._use_non_blocking_refresh = False
+        # TODO(omairn): RefreshThreadManager() may be specific to sync refresh and may need to be separately implemented for async. If so,
+        # move it back to credentials.py
+        self._refresh_worker = RefreshThreadManager()
+
+    @property
+    # TODO(omairn): We could alternatively not implement this here and use asyncio.get_event_loop().time() for the case of async
+    # since utcnow() is a sync function.
+    def expired(self):
```
clundin25, line 72: This property is deprecated, so I don't think a base class should include it.

google/auth/_credentials_base.py
```diff
@@ -0,0 +1,204 @@
...
+from enum import Enum
+
+from google.auth import _helpers
+from google.auth import metrics
+from google.auth._refresh_worker import RefreshThreadManager
+
+_DEFAULT_UNIVERSE_DOMAIN = "googleapis.com"
+
+
+class _BaseCredentials(metaclass=abc.ABCMeta):
```
clundin25, line 28: I am not convinced a shared base class between sync and async credentials is necessary.  I do not think the code is large enough / complex enough that the tradeoff between reduced code size and an additional inheritance is worthwhile.

Do you mind sharing some of the benefits?

google/auth/transport/requests.py
ohmayr: addressed.

google/auth/_credentials_base.py
ohmayr: I agree that introducing a base class here is not necessary. But it will prevent code duplication for when we implement the async class.

Avoiding repetition is the main benefit. If this is a sticking point we can omit the base class for now so it doesn't block the Async REST project, and it can be added later. Let me know what you think!

vchudnov-g: I would argue for having the base class, even if small. It's an investment in simplicity and lowering future maintenance by having shared logic in one place. Not having had this done in the past in other code has become a source of friction for us subsequently working on that code.

While there is a trivial performance cost in using inheritance [O(µs)], IIUC this code would not be used in any tight inner loops where it would be noticeable by the user. (Correct me if I'm wrong, please.)

google/auth/_credentials_base.py
ohmayr: Moved back to credentials.

google/auth/_credentials_base.py
ohmayr: Moved back to credentials.

google/auth/_credentials_base.py
```diff
@@ -0,0 +1,73 @@
...
+
+        Raises:
+            google.auth.exceptions.RefreshError: If the credentials could
+                not be refreshed.
+        """
+        # pylint: disable=missing-raises-doc
+        # (pylint doesn't recognize that this is abstract)
+        raise NotImplementedError("Refresh must be implemented")
+
+    def _apply(self, headers, token=None):
```
clundin25, line 63: Should this be a public method?

clundin25: Also - Is a partially implemented abstract class a common pattern? Is that something we want to keep?

ohmayr: This method implements common logic between our sync and async code that is meant to be used internally within our subclasses.

Isn't an abstract class meant to be a partially implemented class?

clundin25: You are right. I apologize, I was thinking of this as an `interface`.

clundin25 (approved):

parthea (approved):

parthea:
```
nox > flake8 --import-order-style=google --application-import-names=google,tests,system_tests google tests tests_async
google/auth/credentials.py:26:1: I100 Import statements are in the wrong order. 'from google.auth._credentials_base import _BaseCredentials' should be before 'from google.auth._refresh_worker import RefreshThreadManager'
google/auth/transport/requests.py:41:1: I100 Import statements are in the wrong order. 'import google.auth.transport._mtls_helper' should be before 'from google.auth.transport._requests_base import _BaseAuthorizedSession'
nox > Command flake8 --import-order-style=google --application-import-names=google,tests,system_tests google tests tests_async failed with exit code 1
nox > Session lint failed.
```

parthea (approved):

google/auth/transport/_requests_base.py
vchudnov-g: In terms of the `async` semantics, yes that is a valid consideration. But there are still common parts of the logic are re-used between sync and async and those would make sense to consider factoring out. (I think in the rant you linked to, that corresponds to the "blue" functions.)

PR #1558, opened 2024-07-11
feat(auth): Update get_client_ssl_credentials to support X.509 workload certs

andyrzhao:
This PR adds support for loading X.509 workload cert in addition to existing context_aware_metadata based cert.


google/auth/transport/mtls.py
```diff
@@ -24,10 +24,15 @@ def has_default_client_cert_source():
...
+    return (
+        _mtls_helper._check_config_path(_mtls_helper.CONTEXT_AWARE_METADATA_PATH)
+        is not None
+    ) or (
+        _mtls_helper._check_config_path(
+            _mtls_helper.CERTIFICATE_CONFIGURATION_DEFAULT_PATH
+        )
+        is not None
     )
-    return metadata_path is not None
```
clundin25, line 36: Can you refactor this to an early return flow?
 
This is hard to read.

clundin25 (approved):

google/auth/transport/mtls.py
andyrzhao: like this?
    if _mtls_helper._check_config_path(_mtls_helper.CONTEXT_AWARE_METADATA_PATH) is not None:
        return true
    elif _mtls_helper._check_config_path(_mtls_helper.CERTIFICATE_CONFIGURATION_DEFAULT_PATH) is not None:
        return true
    else:
        return false

clundin25:
```
if _mtls_helper._check_config_path(_mtls_helper.CONTEXT_AWARE_METADATA_PATH) is not None:
    return true
if _mtls_helper._check_config_path(_mtls_helper.CERTIFICATE_CONFIGURATION_DEFAULT_PATH) is not None:
    return true
return false
```

Slightly tweaked

PR #1559, opened 2024-07-12
feat: implement async `StaticCredentials` using access tokens

ohmayr:
This PR implements the following:
- `google.auth.aio.credentials.Credentials` which is an abstract class for asynchronous credentials.
- `google.auth.aio.credentials.StaticCredentials` to allow users to configure a valid access token.
- `google.auth.aio.credentials.AnonymousCredential` for testing purposes.


google/auth/aio/credentials.py
```diff
@@ -0,0 +1,74 @@
...
+
+        Args:
+            request (google.auth.transport.Request): The object used to make
+                HTTP requests.
+            method (str): The request's HTTP method or the RPC method being
+                invoked.
+            url (str): The request's URI or the RPC service's URI.
+            headers (Mapping): The request's headers.
+        """
+        await self.apply(headers)
```
clundin25, line 73: Need to add a check:

```
if token_state is STALE or token_state is INVALID:
  await refresh
```

ohmayr: `refresh` isn't implemented here yet. I believe the token state is also not stored here for now (this logic is deferred) because it'll be the user's responsibility to ensure that the token is valid.

google/auth/aio/credentials.py
```diff
@@ -0,0 +1,84 @@
...
+
+    async def apply(self, headers, token=None):
+        """Apply the token to the authentication header.
+
+        Args:
+            headers (Mapping): The HTTP request headers.
+            token (Optional[str]): If specified, overrides the current access
+                token.
+        """
+        self._apply(headers, token=token)
```
clundin25, line 54: I think already the shared base class is showing cracks. A shared private method for an abstract class seems like a code smell

google/oauth2/aio/credentials.py
```diff
@@ -0,0 +1,53 @@
...
+
+from google.auth.aio import credentials
+
+
+class Credentials(credentials.Credentials):
+    """Asynchronous Credentials using OAuth 2.0 access and refresh tokens.
+
+    """
+
+    def __init__(self, token):
```
clundin25, line 42: I don't think the token should be a part of the constructor.

ohmayr: Why not? It's a part of the synchronous version of this method here: https://github.com/googleapis/google-auth-library-python/blob/main/google/oauth2/credentials.py#L73

google/auth/aio/credentials.py
ohmayr: @clundin25  what's wrong with having a protected method in an abstract class? Can you elaborate more on this?

clundin25: Similar comment here - I was thinking of the abstract class as an interface instead of a partial class. My apologies, this is OK

google/oauth2/aio/credentials.py
clundin25: I don't love that API as `token` is dependent on `expiry`. It may not be obvious that both are required, and when expiry is omitted the token is always `valid`.

My 2c, leave it for now and maybe we introduce it using a builder pattern.

ohmayr: I've introduced a builder class here: https://github.com/googleapis/google-auth-library-python/pull/1559/commits/1bb65171b2caedd39498bd8cb2fddbda4d66b7f7. Let me know what you think!


- IIUC from your comment, both `token `and `expiry` are required. Given that the class expects the token to never expire / user's responsibility to re-initialize the credentials with a new token and make the request (if needed), do we want to introduce `expiry` and `setExpiry` at this point?

    -  if we are to move forward with this approach, we should be emitting this information to the user and also lmk that the credentials' class will not attempt to automatically refresh the token before making the request.

clundin25: I think it should be in the same style as the existing code: https://github.com/googleapis/google-auth-library-python/blob/main/google/oauth2/credentials.py#L282.

clundin25: I think a static token class makes sense for your use case.

It is similar to https://github.com/googleapis/google-auth-library-python/blob/main/google/oauth2/credentials.py#L578 but does not support refresh.

google/auth/_credentials_base.py
```diff
@@ -0,0 +1,72 @@
...
+    :meth:`refreshed <refresh>` and used, credentials should be considered
+    immutable. Various credentials will accept configuration such as private
+    keys, scopes, and other options. These options are not changeable after
+    construction. Some classes will provide mechanisms to copy the credentials
+    with modifications such as :meth:`ScopedCredentials.with_scopes`.
+    """
+
+    def __init__(self):
+        self.token = None
+        """str: The bearer token that can be used in HTTP headers to make
```
vchudnov-g, line 43: Style: This is really non-standard format. Class attributes are described in the class pydoc, not with this standalone string in the `__init__`. See https://google.github.io/styleguide/pyguide.html#384-classes.

Please fix here and elsewhere.

tests/oauth2/test_credentials_async.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_default_state(self):
+        credentials = (credentials_async.CredentialsBuilder()
+                       .setToken(token=None)
+                       .build())
+        assert credentials.token is None
+
+    @pytest.mark.asyncio
+    async def test_token_usage_metrics(self):
+        credentials = (credentials_async.CredentialsBuilder()
+                       .setToken(token="token")
```
vchudnov-g, line 45: I suggest using a word that has nothing to do with the problem space as a value.
```suggestion
                       .setToken(token="orchid")
```

tests/test_credentials_async.py
```diff
@@ -0,0 +1,49 @@
...
+
+def test_credentials_constructor():
+    credentials = CredentialsImpl()
+    assert not credentials.token
+
+
+@pytest.mark.asyncio
+async def test_before_request():
+    credentials = CredentialsImpl()
+    request = "token"
```
vchudnov-g, line 32: Ditto here and below: use words as unrelated as possible from the problem domain.

tests/test_credentials_async.py
```diff
@@ -0,0 +1,49 @@
...
+    headers = {}
+    credentials.token = "token"
```
vchudnov-g, line 34: Unless the values are required to be identical, use distinct values in your test (i.e. not `token` in both places here)

tests/oauth2/test_credentials_async.py
ohmayr: oauth2 credentials and the corresponding test cases were removed. But I've addressed this for the other test cases.

tests/test_credentials_async.py
ohmayr: Addressed!

tests/test_credentials_async.py
ohmayr: Addressed!

google/oauth2/aio/credentials.py
ohmayr: Addressed!

google/auth/aio/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        Args:
+            token (str): The OAuth 2.0 or JWT access token.
+        """
+        super(StaticCredentials, self).__init__()
+        self.token = token
+
+    @_helpers.copy_docstring(Credentials)
+    async def refresh(self, request):
+        raise exceptions.InvalidOperation("Static credentials cannot be refreshed.")
```
ohmayr, line 110: @clundin25 Do we want to raise a `RefreshError` instead to match what we have in the docstring?

google/auth/_credentials_base.py
```diff
@@ -37,12 +37,14 @@ class _BaseCredentials(metaclass=abc.ABCMeta):
     keys, scopes, and other options. These options are not changeable after
     construction. Some classes will provide mechanisms to copy the credentials
     with modifications such as :meth:`ScopedCredentials.with_scopes`.
+    
+    Attributes:
+        token (str): The bearer token that can be used in HTTP headers to make
```
clundin25, line 42: Should this be `Optional[str]`?

google/auth/aio/credentials.py
vchudnov-g: That would seem sensible here.

google/auth/aio/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+            method (str): The request's HTTP method or the RPC method being
+                invoked.
+            url (str): The request's URI or the RPC service's URI.
+            headers (Mapping): The request's headers.
+        """
+        await self.apply(headers)
+
+
+class StaticCredentials(Credentials):
+    """Asynchronous Credentials using access tokens.
```
clundin25, line 88:
```suggestion
    """Asynchronous Credentials representing an immutable access token.
```

google/auth/aio/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+
+    StaticCredentials does not support :meth `refresh` and assumes that the configured
+    token is valid and not expired. StaticCredentials will never attempt to 
+    refresh the token.
+    """
+
+    def __init__(self, token):
+        """
+        Args:
+            token (str): The OAuth 2.0 or JWT access token.
```
clundin25, line 103:
```suggestion
            token (str): The access token.
```

google/auth/aio/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        super(StaticCredentials, self).__init__()
+        self.token = token
+
+    @_helpers.copy_docstring(Credentials)
+    async def refresh(self, request):
+        raise exceptions.InvalidOperation("Static credentials cannot be refreshed.")
+
+    @_helpers.copy_docstring(Credentials)
+    async def before_request(self, request, method, url, headers):
```
clundin25, line 113: Can you add a comment that points out to future maintainers that `before_request` should never try to refresh.

google/auth/aio/credentials.py
clundin25: No this looks great

google/auth/aio/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+        The optional ``token`` argument is not supported.
+
+        Raises:
+            google.auth.exceptions.InvalidValue: If a token was specified.
+        """
+        if token is not None:
+            raise exceptions.InvalidValue("Anonymous credentials don't support tokens.")
+
+    async def before_request(self, request, method, url, headers):
+        """Anonymous credentials do nothing to the request."""
```
clundin25, line 141:
```suggestion
        """Anonymous credentials do nothing to the request."""
        pass
```

clundin25:
A few tiny nits, but this LGTM.

google/auth/aio/credentials.py
clundin25: IMO

`RefreshError` could be due to a configuration / auth issue. Since this would be due to a bug in the application, I feel `InvalidOperation` is appropriate.

(I acknowledge there is no defined guidance here)

ohmayr: Thanks for the clarification!

google/auth/aio/credentials.py
ohmayr: addressed.

google/auth/aio/credentials.py
ohmayr: addressed.

google/auth/aio/credentials.py
ohmayr: addressed.

google/auth/aio/credentials.py
ohmayr: addressed.

google/auth/_credentials_base.py
ohmayr: addressed.

clundin25 (approved):

PR #1569, opened 2024-08-01
feat: Implement asynchronous timeout context manager

ohmayr:
This PR implements an asynchronous timeout guard which will be used for asynchronous requests.

Fixes: googleapis/google-cloud-python#15175

### DIFF
--- a/google/auth/aio/transport/aiohttp.py
+++ b/google/auth/aio/transport/aiohttp.py
@@ -45,17 +45,21 @@ async def timeout_guard(timeout):
         elapsed = time.monotonic() - start
         remaining = total_timeout - elapsed
         if remaining <= 0:
-            raise TimeoutError(f"Context manager exceeded the configured timeout of {total_timeout}s.")
+            raise TimeoutError(
+                f"Context manager exceeded the configured timeout of {total_timeout}s."
+            )
         return remaining
-    
+
     async def with_timeout(coro):
         try:
             remaining = _remaining_time()
             response = await asyncio.wait_for(coro, remaining)
             return response
         except (asyncio.TimeoutError, TimeoutError):
-            raise TimeoutError(f"The operation {coro} exceeded the configured timeout of {total_timeout}s.")
-    
+            raise TimeoutError(
+                f"The operation {coro} exceeded the configured timeout of {total_timeout}s."
+            )
+
     try:
         yield with_timeout
 
--- a/google/auth/aio/transport/aiohttp.py
+++ b/google/auth/aio/transport/aiohttp.py
@@ -14,3 +14,37 @@
 
 """Transport adapter for Asynchronous HTTP Requests.
 """
+
+
+from google.auth.exceptions import TimeoutError
+
+import asyncio
+import time
+from contextlib import asynccontextmanager
+
+
+@asynccontextmanager
+async def timeout_guard(timeout):
+    start = time.monotonic()
+    total_timeout = timeout
+
+    def _remaining_time():
+        elapsed = time.monotonic() - start
+        remaining = total_timeout - elapsed
+        if remaining <= 0:
+            raise TimeoutError(f"Context manager exceeded the configured timeout of {total_timeout}s.")
+        remaining
+    
+    async def with_timeout(op):
+        try:
+            remaining = _remaining_time()
+            response = await asyncio.wait_for(op, remaining)
+            return response
+        except (asyncio.TimeoutError, TimeoutError):
+            raise TimeoutError(f"The operation {op} exceeded the configured timeout of {total_timeout}s.")
+    
+    try:
+        yield with_timeout
+
+    finally:
+        _remaining_time()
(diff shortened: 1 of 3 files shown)


google/auth/aio/transport/requests.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    start = time.monotonic()
+    total_timeout = timeout
+
+    def _remaining_time():
+        elapsed = time.monotonic() - start
+        remaining = total_timeout - elapsed
+        if remaining <= 0:
+            raise TimeoutError(f"Context manager exceeded the configured timeout of {total_timeout}s.")
+        remaining
```
clundin25, line 49: Why is the variable here?

clundin25: Should it be returned?

google/auth/aio/transport/requests.py
```diff
@@ -0,0 +1,63 @@
...
+    
+    async def with_timeout(coro):
+        try:
+            remaining = _remaining_time()
+            response = await asyncio.wait_for(coro, remaining)
+            return response
+        except (asyncio.TimeoutError, TimeoutError):
+            raise TimeoutError(f"The operation {coro} exceeded the configured timeout of {total_timeout}s.")
```
clundin25, line 57:
```suggestion
        except (asyncio.TimeoutError, TimeoutError) as e:
            raise TimeoutError(f"The operation {coro} exceeded the configured timeout of {total_timeout}s.") from e
```

Should we include the underlying timeout too?

tests/transport/aio/test_requests.py
```diff
@@ -0,0 +1,90 @@
...
+
+@pytest.fixture
+async def simple_async_task():
+        await asyncio.sleep(0.1)
+        return True
+
+@pytest.fixture
+async def long_running_async_task():
+    await asyncio.sleep(0.3)
+    return True
```
clundin25, line 29: Can you run `nox -s blacken` to fix the formatting?

tests/transport/aio/test_requests.py
```diff
@@ -0,0 +1,90 @@
...
+
+import google.auth.aio.transport.requests as requests_aio
+import pytest # type: ignore
+import asyncio
+from google.auth.exceptions import TimeoutError
+from unittest.mock import patch
+
+@pytest.fixture
+async def simple_async_task():
+        await asyncio.sleep(0.1)
```
clundin25, line 23: Perhaps we should mock `asyncio.sleep` so the test code doesn't have the extra wall clock time.

tests/transport/aio/test_requests.py
ohmayr: Addressed in: https://github.com/googleapis/google-auth-library-python/pull/1569/commits/9e9aab4a876b6de6bd15a6b75b4a1d989a1eb3ff

google/auth/aio/transport/requests.py
ohmayr: This is a good catch! It should've been returned. Fixed now.

It changes how the test cases work. It now makes sense to me why some of the tests were unintentionally working due to the missing `return` statement.

Essentially, assertions shouldn't happen within the context manager because it would exit before executing those statements in some cases. I've addressed this as well.

Thank you @clundin25!

google/auth/aio/transport/requests.py
ohmayr: Addressed in https://github.com/googleapis/google-auth-library-python/pull/1569/commits/3fb323c655097c1c4d902d91cf268089489e32c4

tests/transport/aio/test_requests.py
ohmayr: @clundin25 I think mocking `asyncio.sleep` may defeat the purpose of testing the passage of time in the timeout context manager. But I do see the added cost of time. 

_what are some other options?_ 

-> use `0.01` instead of `0.1`.
-> mock `time.monotonic` (which we're using in our context manager to track time) to return a defined iterable. [The only issue with this approach is that `asyncio.wait_for` also uses `time.monotonic` so this approach can lead to unexpected behaviour.
-> use an external library similar to `freezegun` but for asynchoronous code; maybe `pytest-freezegun`?.

Let me know what you think!

clundin25: I'll add a "nit", I don't think it is necessary to do this right away. With mocks these tests would ideally be deterministic. 

For example `test_timeout_with_single_async_task_within_bounds` could have a mock `asyncio.wait_for` that returns immediately. 

`test_timeout_with_multiple_async_tasks_out_of_bounds` could have a mock that returns on the first call, and then raises a timeout on the second.

> use 0.01 instead of 0.1. 

I think the test would be flaky / non-deterministic 

> mock time.monotonic

This sounds like it work, I think.

> use an external library similar to freezegun but for asynchoronous code; maybe pytest-freezegun

If it exists I think this is also a reasonable approach.

ohmayr: Addressed in: https://github.com/googleapis/google-auth-library-python/pull/1569/commits/bd089259ac890fbc4f0e240c58162b2536342543.

clundin25 (approved):

PR #1575, opened 2024-08-07
feat: Implement asynchronous `AuthorizedSession` api response class

ohmayr:
This class implements an interface for the asynchronous `Response` as well as an implemented class that wraps an `aiohttp.ClientResponse`.

### DIFF
--- a/google/auth/aio/transport/__init__.py
+++ b/google/auth/aio/transport/__init__.py
@@ -23,3 +23,29 @@ to support HTTP libraries. :class:`Request` defines the interface expected by
 :mod:`google.auth` to make asynchronous requests. :class:`Response` defines the interface
 for the return value of :class:`Request`.
 """
+
+import abc
+
+
+class Response(metaclass=abc.ABCMeta):
+    """Asynchronous HTTP Response Interface."""
+
+    @abc.abstractmethod
+    async def status_code(self):
+        """int: The HTTP status code."""
+        raise NotImplementedError("status_code must be implemented.")
+
+    @abc.abstractmethod
+    def headers(self):
+        """Mapping[str, str]: The HTTP response headers."""
+        raise NotImplementedError("headers must be implemented.")
+
+    @abc.abstractmethod
+    def content(self):
+        """bytes: The raw response content."""
+        raise NotImplementedError("content must be implemented.")
+
+    @abc.abstractmethod
+    async def close(self):
+        """Release the connection back to the connection pool."""
+        raise NotImplementedError("close must be implemented.")
--- a/google/auth/aio/transport/__init__.py
+++ b/google/auth/aio/transport/__init__.py
@@ -77,5 +77,5 @@ class Response(metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     async def close(self):
-        """Close the response."""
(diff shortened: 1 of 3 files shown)


google/auth/aio/transport/__init__.py
```diff
@@ -23,3 +23,29 @@
...
+        """int: The HTTP status code."""
+        raise NotImplementedError("status_code must be implemented.")
+
+    @abc.abstractmethod
+    def headers(self):
+        """Mapping[str, str]: The HTTP response headers."""
+        raise NotImplementedError("headers must be implemented.")
+
+    @abc.abstractmethod
+    def content(self):
```
clundin25, line 44: Should this be async?

ohmayr: The `content` property exposed by `aiohttp` is synchronous. I believe it's synchronous for `httpx` as well but I'll double check.

Do we want to make this async?

ohmayr: As per our offline discussion, I've updated this to be async.

clundin25 (approved):

google/auth/aio/transport/__init__.py
ohmayr: After putting in some more thought, given that we're declaring `content` as a property, it's expected to be computationally inexpensive. We may also not want to unnecessarily give back control to the event loop.

Therefore, having an async property may not be ideal. Let's discuss this further. Will put this PR on hold until then.

ohmayr: We've decided to update content to be a method rather than an attribute so we can keep it as async. We've also updated its return type to be an `AsyncGenerator`.

google/auth/aio/transport/__init__.py
```diff
@@ -23,3 +23,59 @@
...
+        """Read the entire response content as bytes.
+
+        Returns:
+            bytes: The entire response content.
+        """
+        raise NotImplementedError("read must be implemented.")
+
+    @abc.abstractmethod
+    async def close(self):
+        """Close the response."""
```
clundin25, line 80: When should the response be closed? After you are finished reading / streaming content?

What does closing do, is it for cleaning up memory?

google/auth/aio/transport/aiohttp.py
```diff
@@ -12,5 +12,57 @@
...
+from google.auth import _helpers
+from typing import AsyncGenerator, Dict
+
+
+class Response(transport.Response):
+
+    """
+    Instances of Response class are returned by 
+    ``google.auth.aio.transport.requests.AuthorizedSession`` and provide methods to interact
+    with the response data.
```
clundin25, line 34:
```suggestion
   Represents an HTTP response and it's data. It is returned by 
    ``google.auth.aio.transport.requests.AuthorizedSession``.
```

clundin25 (approved):

google/auth/aio/transport/__init__.py
ohmayr: Addressed.

google/auth/aio/transport/aiohttp.py
ohmayr: Addressed.

PR #1576, opened 2024-08-09
chore: Add anonymous credentials json file

dhpollack:
This allows one to use a credentials json file to create anonymous credentials.  

https://googleapis.dev/python/google-auth/latest/_modules/google/auth/credentials.html#AnonymousCredentials

As the docs for `AnonymousCredentials` themselves states, this only useful for anonymous services and emulators.  But it would mean that one could include a credentials file in a project that used an emulator and authenticate in the exact same way as with real credentials.  Otherwise one needs to add application logic to allow for anonymous credentials and real credentials.


dhpollack:
stale

PR #1579, opened 2024-08-11
feat: Implement asynchronous `AuthorizedSession` api request class

ohmayr:
This class implements the following:
- an interface for making asynchronous requests which is consumed by the AuthorizedSession API i.e. `google.auth.aio.transport.Request`.
- a concrete class which leverages `aiohttp.ClientSession` to make asynchronous requests i.e. `google.auth.aio.transport.aiohttp.Request`.

### DIFF
--- a/google/auth/aio/transport/__init__.py
+++ b/google/auth/aio/transport/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Transport - Async HTTP client library support.
+"""Transport - Asynchronous HTTP client library support.
 
 :mod:`google.auth.aio` is designed to work with various asynchronous client libraries such
 as aiohttp. In order to work across these libraries with different
@@ -23,3 +23,49 @@ to support HTTP libraries. :class:`Request` defines the interface expected by
 :mod:`google.auth` to make asynchronous requests. :class:`Response` defines the interface
 for the return value of :class:`Request`.
 """
+
+import abc
+
+
+class Request(metaclass=abc.ABCMeta):
+    """Interface for a callable that makes HTTP requests.
+
+    Specific transport implementations should provide an implementation of
+    this that adapts their specific request / response API.
+
+    .. automethod:: __call__
+    """
+
+    @abc.abstractmethod
+    async def __call__(
+        self, url, method="GET", body=None, headers=None, timeout=None, **kwargs
+    ):
+        """Make an HTTP request.
+
+        Args:
+            url (str): The URI to be requested.
+            method (str): The HTTP method to use for the request. Defaults
+                to 'GET'.
+            body (bytes): The payload / body in HTTP request.
+            headers (Mapping[str, str]): Request headers.
+            timeout (Optional[int]): The number of seconds to wait for a
+                response from the server. If not specified or if None, the
+                transport-specific default timeout will be used.
+            kwargs: Additionally arguments passed on to the transport's
+                request method.
+
+        Returns:
+            Response: The HTTP response.
+
+        Raises:
+            google.auth.exceptions.TransportError: If any exception occurred.
+        """
+        # pylint: disable=redundant-returns-doc, missing-raises-doc
+        # (pylint doesn't play well with abstract docstrings.)
+        raise NotImplementedError("__call__ must be implemented.")
+
+    async def close(self):
+        """
+        Close the underlying session.
+        """
+        raise NotImplementedError("close must be implemented.")
--- a/google/auth/aio/transport/__init__.py
+++ b/google/auth/aio/transport/__init__.py
@@ -94,10 +94,10 @@ class Request(metaclass=abc.ABCMeta):
     async def __call__(
         self,
         url: str,
-        method: Optional[str] = "GET",
-        body: bytes = None,
-        headers: Mapping[str, str] = None, # TODO (ohmayr): headers in the response be the same type?
-        timeout: Optional[float] = None,
+        method: Optional[str],
+        body: bytes,
+        headers: Mapping[str, str],
+        timeout: Optional[float],
         **kwargs
     ) -> Response:
         """Make an HTTP request.
--- a/google/auth/aio/transport/__init__.py
+++ b/google/auth/aio/transport/__init__.py
@@ -25,7 +25,7 @@ for the return value of :class:`Request`.
 """
 
 import abc
-from typing import AsyncGenerator, Dict
+from typing import AsyncGenerator, Dict, Mapping, Optional
 
 
 class Response(metaclass=abc.ABCMeta):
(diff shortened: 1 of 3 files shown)


google/auth/aio/transport/aiohttp.py
```diff
@@ -14,3 +14,103 @@
...
+
+try:
+    import aiohttp
+except ImportError as caught_exc:  # pragma: NO COVER
+    raise ImportError(
+        "The aiohttp library is not installed from please install the aiohttp package to use the aiohttp transport."
+    ) from caught_exc
+from google.auth import exceptions
+
+_DEFAULT_TIMEOUT = 180  # in seconds
```
clundin25, line 28:
```suggestion
_DEFAULT_TIMEOUT_SECONDS = 180
```

google/auth/aio/transport/aiohttp.py
```diff
@@ -14,3 +14,103 @@
...
+
+
+class Request:
```
clundin25, line 31: Should this inherit from the abstract class?

google/auth/aio/transport/aiohttp.py
```diff
@@ -14,3 +14,103 @@
...
+        session (aiohttp.ClientSession): An instance :class:`aiohttp.ClientSession` used
+            to make HTTP requests. If not specified, a session will be created.
+
+    .. automethod:: __call__
+    """
+
+    def __init__(self, session=None):
+        # TODO(ohmayr): Evaluate if we want auto_decompress=False.
+        # and if we want to update it in the passed in session.
+        self.session = session or aiohttp.ClientSession(auto_decompress=False)
```
clundin25, line 57: Let's just use the defaults, and if anyone needs to set a custom parameter they can pass in their own session.

What do you think?

google/auth/aio/transport/aiohttp.py
```diff
@@ -14,3 +14,103 @@
...
+        """
+
+        try:
+            # TODO (ohmayr): verify the timeout type. We may need to pass
+            # in aiohttp.ClientTimeout. Alternatively, we can incorporate
+            # per request timeout within the timeout_guard context manager.
+            response = await self.session.request(
+                method, url, data=body, headers=headers, timeout=timeout, **kwargs
+            )
+            # TODO(ohmayr): Wrap this with Custom Response.
```
clundin25, line 99: When will the TODOs be addressed?

google/auth/aio/transport/aiohttp.py
```diff
@@ -14,3 +14,103 @@
...
+            return response
+
+        except aiohttp.ClientError as caught_exc:
+            new_exc = exceptions.TransportError(caught_exc)
```
clundin25, line 103: Is it necessary to build the exception with the caught exception when you are raising with context? 

Maybe instead write an error message and simply raise it. I am curious why this pattern is in the codebase.
```suggestion
            new_exc = exceptions.TransportError(f"Failed to send request to {url}.")
```

tests/transport/aio/test_aiohttp.py
```diff
@@ -13,4 +13,48 @@
...
 import google.auth.aio.transport.aiohttp as auth_aiohttp
-import pytest # type: ignore
+import pytest  # type: ignore
+import pytest_asyncio
+import asyncio
+from google.auth import exceptions
+from aioresponses import aioresponses
+
+# TODO (ohmayr): Verify if we want to optionally run these test cases
+# if aohttp is installed instead of raising an exception.
```
clundin25, line 23: I think it makes sense to raise the exception so these aren't silently skipped.

ohmayr: Okay makes sense. I've kept it as is and removed the TODO.

google/auth/aio/transport/aiohttp.py
ohmayr: Addressed.

google/auth/aio/transport/aiohttp.py
ohmayr: Addressed.

google/auth/aio/transport/aiohttp.py
ohmayr: Sounds good to me. I think we can work around by not setting `auto_decompress` so it should be fine or come back to it later if always needed.

What I'm unsure about is why `trust_env` is set [here](https://github.com/googleapis/google-auth-library-python/blob/18c2ec1b571d506c0dbcffc483aa5e7b95e1b246/google/auth/transport/_aiohttp_requests.py#L311) as False.

By default the value is set to True i.e. it'll automatically pull up any proxy settings in the environment unlike what we've been doing in our experimental code.

google/auth/aio/transport/aiohttp.py
ohmayr: addressed.

google/auth/aio/transport/aiohttp.py
ohmayr: addressed.

google/auth/aio/transport/__init__.py
```diff
@@ -79,3 +79,53 @@ async def read(self) -> bytes:
...
+    .. automethod:: __call__
+    """
+
+    @abc.abstractmethod
+    async def __call__(
+        self,
+        url: str,
+        method: Optional[str],
+        body: bytes,
+        headers: Mapping[str, str],
```
ohmayr, line 99: @clundin25 
Do we want to change the `Response.headers` type to a `Mapping` instead of a `Dict` to match this?

google/auth/aio/transport/aiohttp.py
```diff
@@ -113,4 +116,100 @@ async def read(self) -> bytes:
...
+    to construct or use this class directly.
+
+    This class can be useful if you want to configure a Request callable
+    with a custom ``aiohttp.ClientSession`` in :class:`AuthorizedSession` or if
+    you want to manually refresh a :class:`~google.auth.aio.credentials.Credentials` instance::
+
+        import aiohttp
+        import google.auth.aio.transport.aiohttp
+
+        # Use Case 1:
```
clundin25, line 136:
```suggestion
        # Default use:
```

google/auth/aio/transport/aiohttp.py
```diff
@@ -113,4 +116,100 @@ async def read(self) -> bytes:
...
+        request = google.auth.aio.transport.aiohttp.Request()
+        await credentials.refresh(request)
+
+        # Use Case 2:
```
clundin25, line 140:
```suggestion
        # Custom aiohttp Session Example
```

clundin25 (approved):

google/auth/aio/transport/aiohttp.py
ohmayr: I'd like to correct myself here. The default value for `trust_env` is False.

will punt adding this as a parameter until required. Thanks!

google/auth/aio/transport/__init__.py
ohmayr: If required, will be addressed in the cleanup PR towards the end.

google/auth/aio/transport/aiohttp.py
ohmayr: addressed.

PR #1586, opened 2024-08-19
docs: Improve AWS supplier samples

aeitzman:


docs/user-guide.rst
```diff
@@ -552,36 +552,45 @@ whether the credential retrieval is retryable.
 Any call to the supplier from the Identity Pool credential will send a :class:`google.auth.external_account.SupplierContext`
 object, which contains the requested audience and subject type. Additionally, the credential will
 send the :class:`google.auth.transport.requests.Request` passed in the credential refresh call which
-can be used to make HTTP requests.::
+can be used to make HTTP requests. Using a custom supplier allows workload or workforce identity federation to be used
```
lsirac, line 555: I would reword this and maybe add a new line to:

"Currently, using ADC with your AWS workloads is only supported with EC2. An example of a good use case for using a custom credential suppliers is when your workloads are running in other AWS environments, such as ECS, EKS, Fargate, etc."

Also we should remove the mention of workforce here since it does not support AWS

lsirac (dismissed):

sai-sunder-s (dismissed):

samples/cloud-client/snippets/authenticate_with_aws_supplier.py
```diff
@@ -0,0 +1,65 @@
...
+        try:
+            return aws.AwsSecurityCredentials(aws_credentials.access_key, aws_credentials.secret_key, aws_credentials.token)
+        except Exception as e:
+            raise exceptions.RefreshError(e, retryable=True)
+
+    def get_aws_region(self, context, request):
+        return self._region
+
+
+def authenticate_with_aws_supplier(project_id="your-google-cloud-project-id", aws_region="your_aws_region", audience="your_federation_audience"):
```
parthea, line 41: `samples/cloud-client/snippets/snippets_test.py` needs to be updated to call `authenticate_with_aws_supplier` to add this to the existing samples presubmit tests

PR #1588, opened 2024-08-20
fix: add types to default and verify_token and Request __init__ based on comments in the source code.

wyattscarpenter:
Recently, I've been using this library with mypy strict mode, which doesn't like the fact that these functions are unannotated (they're the only functions I use, and thus the only ones mypy complains about to me). I'm happy to see that this project has py.typed, so these annotations should fix the problem!


arithmetic1728:
Could you squash your commits, and change the commit message to "fix: add types to verify_token and request __init__"? The `conventionalcommits.org` test requires the commit message to start with the "fix:" prefix. Thanks!

wyattscarpenter:
Happy to!

google/oauth2/id_token.py
```diff
@@ -105,12 +108,12 @@ def _fetch_certs(request, certs_url):
...
-    audience=None,
-    certs_url=_GOOGLE_OAUTH2_CERTS_URL,
-    clock_skew_in_seconds=0,
-):
+    id_token: Union[str, bytes],
+    request: transport.Request,
+    audience: str | list[str] | None = None,
+    certs_url: str = _GOOGLE_OAUTH2_CERTS_URL,
+    clock_skew_in_seconds: int = 0,
+) -> Mapping[str, Any]:
```
wyattscarpenter, line 115: It's actually bad to return something with Any in it, but, um, that's just what was in the comment. ¯\\\_(ツ)\_/¯

If the reviewer happens to have a better & more-accurate idea, let him change this line accordingly.

wyattscarpenter:
I'm hoping I can get this looked at in the near future. Also, it seems like (at least this time), it's the title of the PR that has to have fix: in it for conventionalcommits.org to pass, so  have done so.

google/auth/_default.py
```diff
@@ -532,7 +536,12 @@ def _apply_quota_project_id(credentials, quota_project_id):
     return credentials
 
 
-def default(scopes=None, request=None, quota_project_id=None, default_scopes=None):
+def default(
+        scopes: Sequence[str],
```
wyattscarpenter: The comment did not actually say this was nullable/optional, so I did not keep it so, even though the previous value here had an =None.

wyattscarpenter: Actually I was using the function without providing this parameter, so, you know, case in point. It's Optional now.

google/auth/_default.py
```diff
@@ -588,7 +592,12 @@ def _apply_quota_project_id(credentials, quota_project_id):
     return credentials
 
 
-def default(scopes=None, request=None, quota_project_id=None, default_scopes=None):
+def default(
+        scopes: Optional[Sequence[str]] = None,
+        request: Optional[Request] = None,
```
chalmerlowe:
```suggestion
        request: Optional["Request"] = None,
```

google/auth/_default.py
```diff
@@ -588,7 +592,12 @@ def _apply_quota_project_id(credentials, quota_project_id):
...
+        quota_project_id: Optional[str] = None,
+        default_scopes: Optional[Sequence[str]] = None,
+        ) -> tuple[Credentials, Optional[str]]:
```
chalmerlowe:
```suggestion
        ) -> tuple["Credentials", Optional[str]]:
```

google/auth/_default.py
```diff
@@ -21,10 +21,14 @@
...
 import logging
 import os
+from collections.abc import Sequence
+from typing import Optional
 import warnings
 
 from google.auth import environment_vars
 from google.auth import exceptions
+from google.auth.credentials import Credentials
+from google.auth.transport import Request
```
chalmerlowe:
```suggestion
if TYPE_CHECKING:
    from google.auth.credentials import Credentials
    from google.auth.transport import Request
```

google/auth/_default.py
```diff
@@ -21,10 +21,14 @@
 import json
 import logging
 import os
+from collections.abc import Sequence
+from typing import Optional
```
chalmerlowe:
```suggestion
from collections.abc import Sequence
from typing import TYPE_CHECKING, Optional
```

chalmerlowe (commented):
LGTM. Let's see how the tests turn out.
`Kokoro` and `Kokoro system-3.10` were failing due to circular imports. Hopefully we can nix that using the conditional.

google/oauth2/id_token.py
```diff
@@ -105,12 +108,12 @@ def _fetch_certs(request, certs_url):
 
 
 def verify_token(
-    id_token,
-    request,
-    audience=None,
-    certs_url=_GOOGLE_OAUTH2_CERTS_URL,
-    clock_skew_in_seconds=0,
-):
+    id_token: Union[str, bytes],
```
chalmerlowe, line 110:
```suggestion
    id_token: str | bytes,
```

google/auth/transport/requests.py
chalmerlowe, line 23:
```suggestion
import time
from typing import Optional
```

google/auth/transport/requests.py
```diff
@@ -137,7 +137,7 @@ class Request(transport.Request):
     .. automethod:: __call__
     """
 
-    def __init__(self, session=None):
+    def __init__(self, session: requests.Session | None = None) -> None:
```
chalmerlowe:
```suggestion
    def __init__(self, session: [Optional[requests.Session] = None) -> None:
```

google/oauth2/id_token.py
```diff
@@ -105,12 +108,12 @@ def _fetch_certs(request, certs_url):
 
 
 def verify_token(
-    id_token,
-    request,
-    audience=None,
-    certs_url=_GOOGLE_OAUTH2_CERTS_URL,
-    clock_skew_in_seconds=0,
-):
+    id_token: str | bytes,
```
chalmerlowe:
```suggestion
    id_token: Union[str, bytes],
```

google/oauth2/id_token.py
```diff
@@ -105,12 +108,12 @@ def _fetch_certs(request, certs_url):
...
+    request: transport.Request,
+    audience: str | list[str] | None = None,
```
chalmerlowe:
```suggestion
    audience: Union[str, list[str], None] = None,
```

google/auth/transport/requests.py
```diff
@@ -137,7 +138,7 @@ class Request(transport.Request):
     .. automethod:: __call__
     """
 
-    def __init__(self, session=None):
+    def __init__(self, session: [Optional[requests.Session] = None) -> None:
```
chalmerlowe:
```suggestion
    def __init__(self, session: [Optional[requests.Session]] = None) -> None:
```

google/auth/transport/requests.py
```diff
@@ -137,7 +138,7 @@ class Request(transport.Request):
     .. automethod:: __call__
     """
 
-    def __init__(self, session=None):
+    def __init__(self, session: [Optional[requests.Session]] = None) -> None:
```
chalmerlowe:
```suggestion
    def __init__(self, session: Optional[requests.Session] = None) -> None:
```

google/auth/_default.py
```diff
@@ -17,14 +17,20 @@
...
 import json
 import logging
 import os
+from typing import Optional, TYPE_CHECKING
 import warnings
 
 from google.auth import environment_vars
 from google.auth import exceptions
+
+if TYPE_CHECKING: # pragma: NO COVER
```
chalmerlowe:
```suggestion
if TYPE_CHECKING:  # pragma: NO COVER
```

google/auth/_default.py
```diff
@@ -17,14 +17,20 @@
...
 import os
+from typing import Optional, TYPE_CHECKING
 import warnings
 
 from google.auth import environment_vars
 from google.auth import exceptions
+
+if TYPE_CHECKING:  # pragma: NO COVER
+    from google.auth.credentials import Credentials
+    from google.auth.transport import Request
```
chalmerlowe:
```suggestion
if TYPE_CHECKING:  # pragma: NO COVER
    # flake8 raises an error because these lines are not used in the code,
    # they are only used for type-hinting, hence the F401 pragma. 
    from google.auth.credentials import Credentials  # noqa: F401
    from google.auth.transport import Request  # noqa: F401
```

google/auth/_default.py
```diff
@@ -17,14 +17,22 @@
...
 
 from google.auth import environment_vars
 from google.auth import exceptions
+
+if TYPE_CHECKING:  # pragma: NO COVER
+    # flake8 raises an error because these lines are not used in the code,
+    # they are only used for type-hinting, hence the F401 pragma. 
+    from google.auth.credentials import Credentials  # noqa: F401
+    from google.auth.transport import Request  # noqa: F401
 import google.auth.transport._http_client
```
chalmerlowe:
```suggestion
import google.auth.transport._http_client

if TYPE_CHECKING:  # pragma: NO COVER
    from google.auth.credentials import Credentials  # noqa: F401
    from google.auth.transport import Request  # noqa: F401
```

chalmerlowe:
This almost broke me.

For my own sake and anyone else who looks at all this commit hell:

* `blacken`, `flake8`, `mypy`, `coverage`, and `sphinx` all have rules for how things should be formatted, checked, and tested. Some of those rules differ by Python Runtime and some introduce subtle interactions that caused one OR more of these utilities to fail in ways that I did not expect.

I attempted to do the following:
* ensure that the type-hints were viable in under both Python 3.9 and 3.10+ since there was type hinting capabilities in 3.10 that was missing in 3.9
* avoid a circular import by adding an if condition related to type checking to control when/how specific libraries were imported i.e. only during type checking, not during runtime.
* comment the change so future me would remember why
* remove the comment because `flake8` has rigid rules about such type checking conditionals: it expects 3 lines between it and the last normal import statement: the import statement, a blank line, and the if statement. Comments added extra lines to the block and `flake8` was NOT happy.
* include full paths to the `Request` and `Credentials` type hints because sphinx saw multiple cross references to different `Request` and `Credential` types that come from different parts of the code and did not know how to cross reference them correctly.
* add `flake8` pragmas because `flake8` also didn't like the fact that the imports were not used in the code (`flake8` did not recognize that the if statement and code block were required for type checking... it just considered them unused)
* reorder the import statements to put the if conditional last, since `flake8` has rules about where such things can go
* add pragmas to ensure `coverage` won't flag missing unittest coverage for the if conditional (we chose not to unittest it since it is used solely by the type checker).
* ensure there are two spaces in front of the pragma because blacken requires two spaces:
  * `if TYPE_CHECKING:  # pragma: NO COVER`
  versus
  * `if TYPE_CHECKING: # pragma: NO COVER`
* add several miscellaneous commits and tests to try and discover what was happening with all of these intermingled rules, etc.

chalmerlowe (approved):
LGTM

Linchin (approved):

wyattscarpenter:
Thanks!

PR #1595, opened 2024-09-16
fix: remove token_info call from token refresh path

sai-sunder-s:
A call to `token_info` endpoint is being made in critical path when refreshing access tokens. Removing that call and instead going to provide the `account` from user credential file.


clundin25:
@sai-sunder-s could we leave this in as a fallback? I am worried this will break certain use cases.

google/oauth2/credentials.py
```diff
@@ -351,33 +350,6 @@ def with_universe_domain(self, universe_domain):
     def _metric_header_for_usage(self):
         return metrics.CRED_TYPE_USER
 
-    def _set_account_from_access_token(self, request):
```
clundin25, line 354: Could you alternatively perform this in the `account` property in case `_account` is not already set?

clundin25 (approved):

PR #1604, opened 2024-10-10
feat: making iam endpoint universe-aware

TimurSadykov:
This adds Impersonated credential support of universe domain. So they call endpoint in proper universes.


google/auth/impersonated_credentials.py
```diff
@@ -219,6 +219,8 @@ def __init__(
                 and self._source_credentials._always_use_jwt_access
             ):
                 self._source_credentials._create_self_signed_jwt(None)
+        if (self.universe_domain != self._source_credentials.universe_domain):
```
sai-sunder-s, line 222: maybe we should throw error if the universes are different?

google/auth/iam.py
```diff
@@ -35,11 +35,14 @@
     http_client.GATEWAY_TIMEOUT,
 }
 
+_DEFAULT_UNIVERSE_DOMAIN = ["googleapis.com"]
```
sai-sunder-s, line 38: There is already a constant in base class: https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/credentials.py#L28.

Use it in init like https://github.com/googleapis/google-auth-library-python/blob/484c8db151690a4ae7b6b0ae38db0a8ede88df69/google/auth/external_account_authorized_user.py#L78

google/auth/impersonated_credentials.py
TimurSadykov: it is by design, that we don't set universe explicitly for impersonated credentials

We might want to throw if source_cred is GDU and impersonated cred somehow is not, but I think it is not possible

google/auth/impersonated_credentials.py
```diff
@@ -227,6 +229,9 @@ def __init__(
         self.expiry = _helpers.utcnow()
         self._quota_project_id = quota_project_id
         self._iam_endpoint_override = iam_endpoint_override
+        
+        if (self._iam_endpoint_override == None):
+            self._iam_endpoint_override = iam._IAM_ENDPOINT.format(self.universe_domain, self._target_principal)
```
sai-sunder-s, line 234: Why do we have to create endpoint override always? What if target_principal is none?

sai-sunder-s: I think if override is not given by user, we should not do it. We should fallback to building the url the same way we are doing now, but with universe domain.

google/auth/impersonated_credentials.py
TimurSadykov: it gets set to GDU automatically in base credential

google/auth/iam.py
TimurSadykov: Yep that was not needed

google/auth/impersonated_credentials.py
TimurSadykov: This does not have to be a user-provided override. GCloud crates impersonated_credentials with iam_override without user explicitly asking for it. Here we following that path. 

We can do it differently but it will make code more complex

TimurSadykov: If target principal in None - it is a misconfiguration that will later fail. Nothing new for the change.

sai-sunder-s:
Use monkeypatch to test whether the correct urls are used. One place to do is test_sign_bytes in test_impersonated_credentials.py. 

For example, follow pattern in test__get_explicit_environ_credentials in test__default.py

google/auth/impersonated_credentials.py
```diff
@@ -173,6 +180,7 @@ def __init__(
         lifetime=_DEFAULT_TOKEN_LIFETIME_SECS,
         quota_project_id=None,
         iam_endpoint_override=None,
+        universe_domain=None,
```
sai-sunder-s, line 183: as per GUAC discussion, do not introduce a way to set universe_domain in top level cred, if not already there.

google/auth/impersonated_credentials.py
```diff
@@ -219,6 +227,12 @@ def __init__(
                 and self._source_credentials._always_use_jwt_access
             ):
                 self._source_credentials._create_self_signed_jwt(None)
+        if (
```
sai-sunder-s, line 230: can skip this check once the universe_domain in parameter is removed.

google/auth/impersonated_credentials.py
```diff
@@ -271,13 +285,19 @@ def _update_token(self, request):
             principal=self._target_principal,
             headers=headers,
             body=body,
+            universe_domain=self.universe_domain,
             iam_endpoint_override=self._iam_endpoint_override,
         )
 
+    def get_iam_sign_endpoint(self):
```
sai-sunder-s, line 292: If this method was introduced for testing only, please remove it and use monkeypatch for testing like rest of the repo.

tests/oauth2/test_service_account.py
```diff
@@ -811,18 +812,19 @@ def test_refresh_iam_flow_non_gdu(self, call_iam_generate_id_token_endpoint):
         )
         request = mock.Mock()
         credentials.refresh(request)
-        req, iam_endpoint, signer_email, target_audience, access_token = call_iam_generate_id_token_endpoint.call_args[
+        req, iam_endpoint, signer_email, target_audience, access_token, universe_domain = call_iam_generate_id_token_endpoint.call_args[
```
sai-sunder-s, line 815: This is another example of how you can test the iam endpoints that are built. You can use this pattern instead of creating a separate method get_iam_sign_endpoint and testing it in impersonated_credentials

tests/test_impersonated_credentials.py
```diff
@@ -124,6 +124,7 @@ def make_credentials(
         lifetime=LIFETIME,
         target_principal=TARGET_PRINCIPAL,
         iam_endpoint_override=None,
+        universe_domain=None,
```
sai-sunder-s, line 127: do not need this. can get from source_credentials.

tests/test_impersonated_credentials.py
```diff
@@ -391,6 +417,28 @@ def test_signer_email(self):
         credentials = self.make_credentials(target_principal=self.TARGET_PRINCIPAL)
         assert credentials.signer_email == self.TARGET_PRINCIPAL
 
+    def test_sign_endpoint(self):
```
sai-sunder-s, line 420: this does not test if sign_bytes method is really using the correct url.

sai-sunder-s (changes requested):

tests/test_impersonated_credentials.py
TimurSadykov: fixed

tests/test_impersonated_credentials.py
TimurSadykov: it was not before ) but added

google/auth/impersonated_credentials.py
TimurSadykov: removed

google/auth/impersonated_credentials.py
TimurSadykov: removed

google/auth/impersonated_credentials.py
TimurSadykov: Yep, that was older impelementation

tests/test_impersonated_credentials.py
```diff
@@ -460,11 +419,19 @@ def test_sign_bytes(self, mock_donor_credentials, mock_authorizedsession_sign):
         request.return_value = response
 
         credentials.refresh(request)
-
         assert credentials.valid
         assert not credentials.expired
 
         signature = credentials.sign_bytes(b"signed bytes")
+        mock_authorizedsession_sign.assert_called_with(
```
sai-sunder-s, line 426: Add similar checks for other endpoints also? idtoken and accesstoken

sai-sunder-s (changes requested):

google/auth/impersonated_credentials.py
```diff
@@ -219,6 +221,15 @@ def __init__(
                 and self._source_credentials._always_use_jwt_access
             ):
                 self._source_credentials._create_self_signed_jwt(None)
+<<<<<<< HEAD
```
sai-sunder-s, line 224: seems some extra stuff from merging

sai-sunder-s (changes requested):

tests/test_impersonated_credentials.py
TimurSadykov: We add as a follow up change, as we did not test it before and the fix is validated.

sai-sunder-s: Looks like there are 0 TPC tests for impersonated creds. I think it is better to add a few with this PR. Maybe TPC versions of test_id_token_from_credential, test_make_from_service_account_self_signed_jwt, test_refresh_success and atleast one with external account as source cred.

TimurSadykov: added those few

tests/test_impersonated_credentials.py
```diff
@@ -580,17 +698,19 @@ def test_id_token_from_credential(
         assert credentials.valid
         assert not credentials.expired
 
-        new_credentials = self.make_credentials(lifetime=None)
-
         id_creds = impersonated_credentials.IDTokenCredentials(
             credentials, target_audience=target_audience, include_email=True
         )
-        id_creds = id_creds.from_credentials(target_credentials=new_credentials)
+        id_creds = id_creds.from_credentials(target_credentials=target_credentials)
```
sai-sunder-s, line 704: why is this line existing? IIUC, removing this will have no impact.

sai-sunder-s: It seems to remove the target_audience?

tests/test_impersonated_credentials.py
```diff
@@ -563,6 +642,45 @@ def test_id_token_from_credential(
...
+    def test_id_token_from_credential_nonGdu(
+        self, mock_donor_credentials, mock_authorizedsession_idtoken
+    ):
+        source_credentials = service_account.Credentials(
+            SIGNER, "some@email.com", TOKEN_URI, universe_domain="foo.bar"
+        )
+        credentials = self.make_credentials(
+            lifetime=None, source_credentials=source_credentials
+        )
+        target_credentials = self.make_credentials(
```
sai-sunder-s, line 664: why is same cred created 2 times?

sai-sunder-s (approved):
Looks good mostly. Have that extra line in id token test. I think there is no need for the 2 creds created in same way. But that seems same way in existing test as well.

PR #1605, opened 2024-10-11
fix: add type hints to credentials

rinarakaki:
Enables https://github.com/googleapis/python-cloud-core/pull/302

I checked:
- [x] `pyright google/auth/_credentials_base.py`
- [ ] `pyright google/auth/_default.py`
- [x] `pyright google/auth/_refresh_worker.py`
- [ ] `pyright google/auth/aio/credentials.py`
- [ ] `pyright google/auth/credentials.py`
- [x] `pyright google/auth/metrics.py`
- [x] `pyright google/oauth2/service_account.py`

### DIFF
--- a/google/auth/_credentials_base.py
+++ b/google/auth/_credentials_base.py
@@ -19,6 +19,7 @@ import abc
 from typing import Optional
 
 from google.auth import _helpers
+from google.auth.transport.requests import Request
 
 
 class BaseCredentials(metaclass=abc.ABCMeta):
@@ -48,7 +49,7 @@ class BaseCredentials(metaclass=abc.ABCMeta):
         self.token: Optional[str] = None
 
     @abc.abstractmethod
-    def refresh(self, request):
+    def refresh(self, request: Request) -> None:
         """Refreshes the access token.
 
         Args:
--- a/google/auth/_credentials_base.py
+++ b/google/auth/_credentials_base.py
@@ -16,11 +16,12 @@
 """Interface for base credentials."""
 
 import abc
+from typing import Optional
 
 from google.auth import _helpers
 
 
-class _BaseCredentials(metaclass=abc.ABCMeta):
+class BaseCredentials(metaclass=abc.ABCMeta):
     """Base class for all credentials.
 
     All credentials have a :attr:`token` that is used for authentication and
@@ -44,7 +45,7 @@ class _BaseCredentials(metaclass=abc.ABCMeta):
     """
 
     def __init__(self):
-        self.token = None
+        self.token: Optional[str] = None
 
     @abc.abstractmethod
     def refresh(self, request):
@@ -62,7 +63,7 @@ class _BaseCredentials(metaclass=abc.ABCMeta):
         # (pylint doesn't recognize that this is abstract)
         raise NotImplementedError("Refresh must be implemented")
 
-    def _apply(self, headers, token=None):
+    def _apply(self, headers: dict[str, str], token: Optional[str] = None):
         """Apply the token to the authentication header.
 
         Args:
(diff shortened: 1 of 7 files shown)


google/auth/_credentials_base.py
```diff
@@ -62,7 +63,7 @@ def refresh(self, request):
         # (pylint doesn't recognize that this is abstract)
         raise NotImplementedError("Refresh must be implemented")
 
-    def _apply(self, headers, token=None):
+    def _apply(self, headers: dict[str, str], token: Optional[str] = None):
```
ohmayr, line 66: Shall we use `Mapping` from typing (here and in other places)? Also update the docstring to `Mapping[str, str]`.

google/auth/_refresh_worker.py
```diff
@@ -32,7 +34,7 @@ def __init__(self):
         self._worker = None
         self._lock = threading.Lock()  # protects access to worker threads.
 
-    def start_refresh(self, cred, request):
+    def start_refresh(self, cred: Credentials, request: Request):
```
ohmayr, line 37:
```suggestion
    def start_refresh(self, cred: Credentials, request: Request): -> bool
```

google/auth/_refresh_worker.py
```diff
@@ -61,8 +63,8 @@ def start_refresh(self, cred, request):
 
     def clear_error(self):
```
ohmayr, line 64:
```suggestion
    def clear_error(self): -> None
```

google/auth/credentials.py
```diff
@@ -18,17 +18,21 @@
...
 import os
+from typing import Optional, Self, Sequence
 
 from google.auth import _helpers, environment_vars
 from google.auth import exceptions
 from google.auth import metrics
-from google.auth._credentials_base import _BaseCredentials
+from google.auth._credentials_base import BaseCredentials
 from google.auth._refresh_worker import RefreshThreadManager
+from google.auth.credentials import Credentials
```
ohmayr, line 28: We're importing from the file itself? This doesn't seem right.

ohmayr:
```suggestion
```

google/auth/_credentials_base.py
rinarakaki: This operation requires the type of `headers` has `__setitem__` method that's why being a `Mapping` is not enough:

```py
headers["authorization"] = "Bearer {}".format(
    _helpers.from_bytes(token or self.token)
)
```

google/auth/_refresh_worker.py
rinarakaki: done

google/auth/_refresh_worker.py
rinarakaki: done

google/auth/credentials.py
rinarakaki: Ah, yeah, sorry you're right

google/auth/_credentials_base.py
```diff
@@ -16,11 +16,12 @@
 """Interface for base credentials."""
 
 import abc
+from typing import Optional
 
 from google.auth import _helpers
 
 
-class _BaseCredentials(metaclass=abc.ABCMeta):
+class BaseCredentials(metaclass=abc.ABCMeta):
```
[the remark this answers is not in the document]
rinarakaki: `_BaseCredentials` is imported from outside the `_credentials_base.py` file, so it's not private in that scope. You can see that it's captured as an error by running:

```sh
pyright google/auth/credentials.py
```

google/auth/_credentials_base.py
rinarakaki: Do you want me to modify the doctring instead?

PR #1488, opened 2024-03-02
feat: add configurable GCE Metadata Server retries

judahrand:
This PR aims to replicate the functionality of the Nodejs library implemented here: https://github.com/googleapis/gcp-metadata/pull/275

PR #1610, opened 2024-10-17
feat: honor `NO_GCE_CHECK` environment variable

0x2b3bfa0:
Currently, `google.auth.default()` always attempts to authenticate using the Google Compute Engine (GCE) metadata service as a fallback. There is no easy way to prevent this behavior, leading to unnecessary delays and unhelpful exceptions, particularly in cases where GCE-based authentication is pointless or undesired.

This pull request introduces a `NO_GCE_CHECK` environment variable, allowing users to explicitly skip GCE metadata service authentication. This implementation mirrors `NO_GCE_CHECK` in [googleapis/google-auth-library-java](https://github.com/googleapis/google-auth-library-java/blob/f154edb3d8503d29f0020b6904dfa40e034ded93/README.md?plain=1#L106-L119):

> _The following are searched (in order) to find the Application Default Credentials:_
>
> [...]
>
> 5. Google Compute Engine built-in credentials
>    - Skip this check by setting the environment variable `NO_GCE_CHECK=true`

> [!NOTE]
> Other clouds also provide similar environment variables, such as `AWS_EC2_METADATA_DISABLED` on AWS.

This change is also tangentially related to the following pull requests:
- https://github.com/googleapis/google-auth-library-python/pull/1488
- https://github.com/googleapis/google-auth-library-python/pull/1481


shcheklein:
Hey folks / maintainers, just to bump this up a bit. Would it be possible to merge / release this? What else do we need to do here - please let us know.

chalmerlowe (approved):
LGTM

daniel-sanche (approved):

PR #1612, opened 2024-10-23
chore: update staging bucket to dev

dandhlee:
The staging bucket is no longer used (b/374954111). Changing to dev instead.

Towards b/374991608 🦕


dandhlee:
This change will be updated once synthtool template is updated.

PR #1613, opened 2024-10-23
fix: changing universe_domain endpoint to "universe-domain"

TimurSadykov:
The endpoint format was updated on GCE side, details in b/368084639


google/auth/compute_engine/_metadata.py
```diff
@@ -294,7 +294,7 @@ def get_universe_domain(request):
             404 occurs while retrieving metadata.
     """
     universe_domain = get(
-        request, "universe/universe_domain", return_none_for_not_found_error=True
```
sai-sunder-s, line 297: Do we have to support both endpoints right now or we can just try only the new one?

sai-sunder-s (approved):

PR #1616, opened 2024-10-24
fix: add timeout for auth requests

harkamaljot:
expose a parameter in the request constructor to allow the end user to pass in the custom timeout value. Currently _DEFAULT_TIMEOUT is being used, which may not be suitable for all the users


google/auth/transport/requests.py
```diff
@@ -137,9 +137,10 @@ class Request(transport.Request):
     .. automethod:: __call__
     """
 
-    def __init__(self, session=None):
```
sai-sunder-s, line 140: Checkout how [AuthorizedSession](https://github.com/googleapis/google-auth-library-python/blob/16c728d30adb36e19aabe34f0ed5a95152d9a135/google/auth/transport/requests.py#L396C9-L396C30) is handling the timeouts and see if we can adapt same. I see something about TimeGuards there as well.

PR #1627, opened 2024-11-01
fix: improve user guide for Impersonation and SA

sai-sunder-s:


docs/user-guide.rst
```diff
@@ -66,7 +66,14 @@ Service account private key files
 +++++++++++++++++++++++++++++++++
 
 A service account private key file can be used to obtain credentials for a
-service account. You can create a private key using the `Credentials page of the
+service account. But it is **not recommended** to download a service account key. If you do need
```
westarle, line 69: I would move this to the bottom and avoid calling it "not recommended" in this context.

You would say something like:

"If you are not able to use any of the authentication methods listed above, a service account private key file can be used...."  This can reduce the need for indexing here.

If you need to make a recommendation, the safest thing is to link to our authentication docs.

westarle (commented):
I like the impersonation changes, one thing to fix in the service account change.

westarle (approved):

PR #1638, opened 2024-11-27
feat: allow users to use jwk keys for verifying ID token

harkamaljot:
allow users to pass in the certs url which provides keys in jwk format eg: https://www.googleapis.com/oauth2/v3/certs, for verifying id tokens.


google/oauth2/id_token.py
```diff
@@ -59,6 +59,13 @@
 import json
 import os
 
+try:
+    import jwt as jwt_lib  # type: ignore
+except ImportError as caught_exc:  # pragma: NO COVER
+    raise ImportError(
+        "The pyjwt library is not installed from please install the pyjwt package to use the jwk certs format."
```
sai-sunder-s, line 66: The message seems to have an extra "from"? Is something supposed to come after the "from"?

google/oauth2/id_token.py
```diff
@@ -59,6 +59,13 @@
 import json
 import os
 
+try:
+    import jwt as jwt_lib  # type: ignore
```
sai-sunder-s, line 63: We have to try importing only if it is required. If not we will be breaking existing users by trying to import and raise the exception.

google/oauth2/id_token.py
```diff
@@ -120,21 +126,30 @@ def verify_token(
...
     certs = _fetch_certs(request, certs_url)
-
-    return jwt.decode(
-        id_token,
-        certs=certs,
-        audience=audience,
-        clock_skew_in_seconds=clock_skew_in_seconds,
-    )
+    if "keys" in certs:
+        jwks_client = jwt_lib.PyJWKClient(certs_url)
```
sai-sunder-s, line 138: This could be a good place to import the library.

google/oauth2/id_token.py
```diff
@@ -120,21 +126,30 @@ def verify_token(
...
     """
     certs = _fetch_certs(request, certs_url)
-
-    return jwt.decode(
-        id_token,
-        certs=certs,
-        audience=audience,
-        clock_skew_in_seconds=clock_skew_in_seconds,
-    )
+    if "keys" in certs:
```
sai-sunder-s, line 137: is this doing a contains search on the entire json or checking if there is a key named "keys" at the top level?

If its the latter, then its ok.

python_wf_headless.txt
sai-sunder-s, line 1: why this empty file?

testing/constraints-3.7.txt
```diff
@@ -11,3 +11,4 @@ setuptools==40.3.0
 rsa==3.1.4
 aiohttp==3.6.2
 requests==2.20.0
+pyjwt==2.0
```
sai-sunder-s, line 14: Is there a way to perform testing with this pyjwt not added to the test? Want to make sure existing users are not broken

sai-sunder-s (changes requested):

testing/constraints-3.7.txt
harkamaljot: ran a manual test to verify existing customers won't be broken. In future we will update our testing so that automated testing can handle this.

sai-sunder-s (approved):

PR #1641, opened 2024-12-05
feat: allow users to use jwk keys for verifying ID token

harkamaljot:
allow users to pass in the certs url which provides keys in jwk format eg: https://www.googleapis.com/oauth2/v3/certs, for verifying id tokens.


docs/requirements-docs.txt
```diff
@@ -3,3 +3,4 @@ sphinx-docstring-typing
 urllib3
 requests
 requests-oauthlib
+pyjwt
```
sai-sunder-s, line 6: what happens if you do not add this dependency here?

harkamaljot: I think we need this dependency for documentation, so removing it also works, removed the dependency and updated the pull request

docs/requirements-docs.txt
```diff
@@ -2,4 +2,4 @@ cryptography
 sphinx-docstring-typing
 urllib3
 requests
-requests-oauthlib
+requests-oauthlib
```
sai-sunder-s, line 5: why is this line getting removed and added same way?

sai-sunder-s (approved):

PR #1677, opened 2025-02-27
feat: add functionality to hash data

ohmayr:
This PR adds a helper function to hash sensitive information such as `access_token`, `id_token` before logging it.

### DIFF
--- a/google/auth/_helpers.py
+++ b/google/auth/_helpers.py
@@ -18,6 +18,7 @@ import base64
 import calendar
 import datetime
 from email.message import Message
+import hashlib
 import sys
 import urllib
 
@@ -27,6 +28,7 @@ from google.auth import exceptions
 # expiry.
 REFRESH_THRESHOLD = datetime.timedelta(minutes=3, seconds=45)
 
+SENSITIVE_FIELDS = {"accessToken", "access_token", "id_token", "client_id", "refresh_token", "client_secret"}
 
 def copy_docstring(source_class):
     """Decorator that copies a method's docstring from another class.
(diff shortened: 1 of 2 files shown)


google/auth/_helpers.py
```diff
@@ -271,3 +273,33 @@ def is_python_3():
...
+            hashed_data[key] = value
+    return hashed_data
+
+
+def _hash_value(value, field_name: str) -> str:
+    """Hashes a value and returns a formatted hash string."""
+    if value is None:
+        return None
+    encoded_value = str(value).encode("utf-8")
+    hash_object = hashlib.sha256()
```
parthea, line 302:
```suggestion
    hash_object = hashlib.sha512()
```

Assuming that this will only be called when debugging, is there a strong reason to use `sha256()` instead of `sha512()` which is more secure?

google/auth/_helpers.py
```diff
@@ -271,3 +273,33 @@ def is_python_3():
...
+
+def hash_sensitive_info(data: dict) -> dict:
+    """
+    Hashes sensitive information within a dictionary.
+
+    Args:
+        data: The dictionary containing data to be processed.
+
+    Returns:
+        A new dictionary with sensitive values replaced by their SHA256 hashes.
```
parthea, line 286:
```suggestion
        A new dictionary with sensitive values replaced by their SHA512 hashes.
```

parthea (approved):

sai-sunder-s (approved):

google/auth/_helpers.py
ohmayr: addressed

PR #1678, opened 2025-02-28
feat: add request response logging to auth

ohmayr:
This PR adds logging support to the auth library.


google/auth/_helpers.py
```diff
@@ -271,3 +296,157 @@ def is_python_3():
...
+        body: The request body (can be None).
+        headers: The request headers (can be None).
+    """
+    # TODO(https://github.com/googleapis/google-auth-library-python/issues/1681): Hash sensitive information.
+    if is_logging_enabled(logger):
+        content_type = (
+            headers["Content-Type"] if headers and "Content-Type" in headers else ""
+        )
+        json_body = _parse_request_body(body, content_type=content_type)
+        logger.debug(
```
sai-sunder-s, line 367: For both request and response, please make sure that you either:
1. Hash sensitive fields and log or
2. If you cannot hash/remove the sensitive fields, then do not log the whole request/response

ohmayr: This has been addressed in https://github.com/googleapis/google-auth-library-python/pull/1700.

google/auth/_helpers.py
```diff
@@ -18,15 +18,40 @@
...
 import sys
+from typing import Any, Dict, Mapping, Optional, Union
 import urllib
 
 from google.auth import exceptions
 
+try:
+    # TODO(https://github.com/googleapis/python-api-core/issues/813): Remove `# type: ignore` when
+    # `google-api-core` type hints issue is resolved.
+    from google.api_core import client_logging  # type: ignore # noqa: F401
```
sai-sunder-s, line 33: doesn't google.api_core depend on auth libs? (Assuming that is gax)

google/auth/_helpers.py
```diff
@@ -18,15 +18,40 @@
...
+# logging is supported in minimum version of google-api-core.
+except ImportError:  # pragma: NO COVER
+    CLIENT_LOGGING_SUPPORTED = False
+
 # The smallest MDS cache used by this library stores tokens until 4 minutes from
 # expiry.
 REFRESH_THRESHOLD = datetime.timedelta(minutes=3, seconds=45)
 
+# TODO(https://github.com/googleapis/google-auth-library-python/issues/1684): Audit and update the list below.
+_SENSITIVE_FIELDS = {
```
sai-sunder-s, line 46: @ohmayr is there a complete analysis on all the possible fields that can contain a secret?

I can see that this does not include `assertion` which is the field used in [Service Account token exchange flow](https://github.com/googleapis/google-auth-library-python/blob/f24976452d741de6a49d9b7a85cdab47812f5312/google/oauth2/_client.py#L297).

google/auth/_helpers.py
ohmayr: `google.api_core` isn't a gax but it does depend on https://github.com/googleapis/python-api-core/blob/main/pyproject.toml#L51.

An alternative is for us to copy the logging code to the auth library.

google/auth/_helpers.py
ohmayr: An audit needs to be done to determine the complete list of sensitive fields. https://github.com/googleapis/google-auth-library-python/issues/1684 is filed to capture this.

parthea (approved):

google/auth/_helpers.py
ohmayr: Submitted https://github.com/googleapis/google-auth-library-python/pull/1748 so we don't ad `google.api_core` as a dependency to auth.

sai-sunder-s (approved):

keepdying:
https://github.com/googleapis/google-auth-library-python/issues/1762

PR #1687, opened 2025-03-04
feat: allow users to call default service account endpoint if no email is passed

harkamaljot:
Currently, when we always fetch the email from the mds server and use that email to make an rpc call to mds server. However, this doesn't allow the user to call the default endpoint. So removing the update of the email after the call is made.

This is a safe change because in case when a email was specified by the user when making a cred object then mds server, return back same email, hence the current change doesn't effect this code path.

In case no email was specified, we call default service account endpoint, similar to other languages rather that updating the default with the email fetched from mds server. Both of these endpoints should return similar tokens in a normal case.

Tested the code on a GCE VM and verified the default path works as intended.


sai-sunder-s (changes requested):

harkamaljot:
closing this pull request in favor of https://github.com/googleapis/google-auth-library-python/pull/1769

PR #1695, opened 2025-03-17
feat: adds GA support for X.509 workload identity federation

aeitzman:


google/auth/identity_pool.py
```diff
@@ -145,9 +148,69 @@ def get_subject_token(self, context, request):
...
+        )
+        trust_chain = self._read_trust_chain()
+        cert_chain = []
+
+        cert_chain.append(_X509Supplier._encode_cert(leaf_cert))
+        for i, cert in enumerate(trust_chain):
+            encoded = _X509Supplier._encode_cert(cert)
+            # Check if the current cert is the leaf cert and raise an exception if it is not the first
+            # cert in the chain.
+            if encoded == cert_chain[0]:
```
sai-sunder-s, line 168: it could be more efficient and easy to understand if this check is outside the loop? You can just chck whether first element of trust chain and leaf are same?

lsirac (approved):

sai-sunder-s (approved):

PR #1699, opened 2025-03-21
fix: add request timeout for MDS requests

sai-sunder-s:
Currently only MDS ping request has a default timeout.

Add same default timeout for other MDS requests as well.


google/auth/compute_engine/_metadata.py
```diff
@@ -178,6 +179,7 @@ def get(
         headers (Optional[Mapping[str, str]]): Headers for the request.
         return_none_for_not_found_error (Optional[bool]): If True, returns None
             for 404 error instead of throwing an exception.
+        timeout (int): How long to wait for the metadata server to respond.
```
parthea, line 182:
```suggestion
        timeout (int): How long to wait for the metadata server to respond in seconds.
```

parthea (approved):

parthea (approved):

harkamaljot (approved):

PR #1703, opened 2025-03-25
fix: fix systest failure

sai-sunder-s:
The latest version of nox seems to breaking. We are getting below error even though the env var is indeed being set by the setup script.
```
nox > Session compute_engine-3.7 raised exception KeyError('GOOGLE_APPLICATION_CREDENTIALS')
  File "/venv/lib/python3.10/site-packages/nox/sessions.py", line 1100, in execute
    self.func(session)
  File "/venv/lib/python3.10/site-packages/nox/_decorators.py", line 94, in __call__
    return self.func(*args, **kwargs)
  File "/tmpfs/src/github/google-auth-library-python/system_tests/noxfile.py", line 276, in compute_engine
    del session.virtualenv.env["GOOGLE_APPLICATION_CREDENTIALS"]
KeyError: 'GOOGLE_APPLICATION_CREDENTIALS'
```

Ideally each nox session should start with the env var. Maybe a bug in the nox session is not resetting the settings with the start of each session.

BEGIN_COMMIT_OVERRIDE
test: fix systest failure
END_COMMIT_OVERRIDE


harkamaljot:
Please update the commit title as fix is repeated multiple times. This is good to unblock the currently failing tests but should we add a tracking somewhere to track this issue so that we can revert back to fetching nox from head, once this issue is fixed by nox?

harkamaljot (approved):

Issue #1508, opened 2024-04-03
Support POST verb for URL-sourced credentials to enable Azure Pipelines WIF


laurensknoll:
**Is your feature request related to a problem? Please describe.**

External account URL-sourced credentials require a HTTP GET endpoint:

> An HTTP GET request should be sent to this local url while injecting the headers key/values (if provided in the configuration file) in the request header. The request should respond with the external credentials subject token to be passed to STS token endpoint.
Source: https://google.aip.dev/auth/4117#determining-the-subject-token-in-microsoft-azure-and-url-sourced-credentials

The Azure DevOps OIDC token endpoint, however, requires a HTTP POST request:

```
POST https://dev.azure.com/{organization}/{scopeIdentifier}/_apis/distributedtask/hubs/{hubName}/plans/{planId}/jobs/{jobId}/oidctoken?serviceConnectionId={serviceConnectionId}&api-version=7.1-preview.1
```
Source: https://learn.microsoft.com/en-us/rest/api/azure/devops/distributedtask/oidctoken/create?view=azure-devops-rest-7.1

Consequently, URL-sourced credentials can't yet be applied to Azure Pipelines. The process fails with the following error, when using the specified credential file:

```
There was a problem refreshing your current auth tokens: ('Unable to retrieve Identity Pool subject token', '{"count":1,"value":{"Message":"The requested resource does not support http method \'GET\'."}}')
```

```json
{
  "type": "external_account",
  "audience":"//iam.googleapis.com/projects/.../locations/global/workloadIdentityPools/example/providers/ado",
  "subject_token_type":"urn:ietf:params:oauth:token-type:jwt",
  "token_url":"https://sts.googleapis.com/v1/token",
  "credential_source":{
    "url":"https://dev.azure.com/.../.../_apis/distributedtask/hubs/build/plans/519d9925-3aa9-481d-ac99-8d10e3f105ec/jobs/12f1170f-54f2-53f3-20dd-22fc7dff55f9/oidctoken?serviceConnectionId=563d287e-a228-4e4e-9293-bbe6adb74605",
    "headers":{
      "Authorization":"Bearer ...",
      "Accept":"application/json;api-version=7.2-preview.1"
    },
    "format":{
      "type":"json",
      "subject_token_field_name":"oidcToken"
    }
  },
[...]

PR #1710, opened 2025-04-02
fix(types): add type annotations to exceptions module for Mypy strict mode (#1508)

cureprotocols:
This PR contributes to issue [#1508](https://github.com/googleapis/google-auth-library-python/issues/1508) by adding full type annotations to `google/auth/exceptions.py`.

All functions and properties now comply with Mypy's `--strict` mode.

### ✔️ Highlights:
- Added explicit return types for all functions and properties.
- Replaced keyword-only syntax with `**kwargs` for Python 3.7 compatibility while preserving clarity and behavior.
- Maintained full backward compatibility with existing exception usage patterns.
- Verified clean via: `mypy --strict google/auth/exceptions.py`.

This is part of the ongoing effort to enhance strict typing across the library.  
Happy to refine or expand this PR further as needed!


cureprotocols:
I signed it!

PR #1732, opened 2025-04-06
fix: add missing packaging dependency for feature requiring urllib3

harkamaljot:
Resolves an `ImportError: No module named 'packaging'` encountered when using functionality in the urllib3.py file.

The code paths relying on `urllib3` transport also have a runtime dependency on the `packaging` library. This dependency was introduced in PR #1390.

However, `packaging` was not explicitly listed as a requirement for the `urllib3`  in `setup.py`. This was masked during development because `nox` often installs `packaging` transitively. Users installing the library without having `nox` or its dependencies installed separately would face an `ImportError` when the `urllib3` dependent code executes.

This PR fixes the issue by explicitly adding `packaging` alongside `urllib3` in the extras group definition. This ensures that installing the optional feature via the extra correctly pulls in all its necessary runtime dependencies.


setup.py
```diff
@@ -34,6 +34,7 @@
     "reauth": "pyu2f>=0.1.5",
     "enterprise_cert": ["cryptography", "pyopenssl"],
     "pyjwt": ["pyjwt>=2.0", "cryptography>=38.0.3"],
+    "urllib3": ["urllib3", "packaging"],
```
sai-sunder-s, line 37: how does user know to install this extra?

Maybe update the error msg in the try catch and tell users about this extra.

Also move packaging into the try?

https://github.com/googleapis/google-auth-library-python/blob/9f7d3fa92c0e656a1c970182833abe2d0d3ad3ee/google/auth/transport/urllib3.py#L43

harkamaljot: installing dependency like pip install google-auth[urllib3] should install all the dependencies defined for urlllib3. I was thinking about updating error message, however I saw similar error in other modules, but this is definitely confusing. let me update the message to include google-auth[urllib3] command to avoid user installing directly from pip.

sai-sunder-s (approved):

PR #1740, opened 2025-04-08
fix: add impersonated SA via local ADC support for fetch_id_token

nolanleastin:
fetch_id_token was only supported for service accounts and fetch ID token from metadata server if it exists. 

This PR adds support for impersonated service accounts.

### DIFF
--- a/google/oauth2/id_token.py
+++ b/google/oauth2/id_token.py
@@ -284,6 +284,19 @@ def fetch_id_token_credentials(audience, request=None):
                     return service_account.IDTokenCredentials.from_service_account_info(
                         info, target_audience=audience
                     )
+                elif info.get("type") == "impersonated_service_account":
+                    from google.auth import impersonated_credentials
+
+                    target_credentials =  impersonated_credentials.Credentials.from_impersonated_account_info(
+                        info
+                    )
+
+                    id_creds = impersonated_credentials.IDTokenCredentials(
+                        target_credentials=target_credentials,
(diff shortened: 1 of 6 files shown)


google/oauth2/id_token.py
```diff
@@ -284,6 +284,18 @@ def fetch_id_token_credentials(audience, request=None):
...
+                    from google.auth import impersonated_credentials
+
+                    target_credentials = impersonated_credentials.Credentials.from_impersonated_service_account_info(
+                        info
+                    )
+
+                    return impersonated_credentials.IDTokenCredentials(
+                        target_credentials=target_credentials,
+                        target_audience=audience,
+                        include_email=True,
```
sai-sunder-s, line 297: do other credential types have the email option as well?

nolanleastin: Right now, it's a bit uneven. 

1. Service Accounts do have include_email=True. https://github.com/googleapis/google-auth-library-python/blob/221f4a82fa25c1ad453b85bc8b7f2fc304724879/google/oauth2/_client.py#L344
2. When running MDS  email is not included. 
3. By default, impersonated_credentials.IDTokenCredentials set include_email=False (https://github.com/googleapis/google-auth-library-python/blob/59e0b381e5d8861727b1dce989072fcb4cf1a7df/google/auth/impersonated_credentials.py#L423)

sai-sunder-s (approved):

PR #1771, opened 2025-05-23
fix: replace rsa dependency with cryptography

sai-sunder-s:
This commit removes the `rsa` library as a dependency and makes the `cryptography` library a required, core dependency.

Previously, `cryptography` was an optional dependency, and the library would fall back to a pure Python RSA implementation using the `rsa` library if `cryptography` was not installed.

Changes made:
- Modified `setup.py` to remove `rsa` from dependencies and add `cryptography` with version constraints.
- Updated `google/auth/crypt/rsa.py` to directly use the `cryptography`-based RSA implementation (`_cryptography_rsa.py`) and remove the fallback mechanism.
- Removed the pure Python RSA implementation file (`google/auth/crypt/_python_rsa.py`).
- Removed the corresponding tests for the pure Python RSA implementation (`tests/crypt/test__python_rsa.py`).

Core unit tests pass after these changes.


docs/conf.py
```diff
@@ -358,6 +358,47 @@
 # texinfo_no_detailmenu = False
 
 
+# -- Options for autodoc --------------------------------------------------
+
+
+def autodoc_skip_member_handler(app, what, name, obj, skip, options):
+    """
+    Skips members from internal modules (like _cryptography_rsa or base)
```
parthea, line 366: Please could you file a bug for issue that requires this docs workaround, even if it will be closed with the changes in this PR.  Add a link to the issue in this comment.

parthea (commented):
LGTM. We should mark this as `fix` instead of `refactor` so that the change to setup.py is visible in release notes.

```
fix: add dependency on cryptography
fix: drop dependency on rsa
```

parthea (commented):
Also remove rsa here: https://github.com/googleapis/google-auth-library-python/blob/25660bc57569b3f3e9a1f664516d10f4bc8744b8/testing/constraints-3.7.txt#L11

HiromuHota:
Any update here? rsa (https://stuvel.eu/software/rsa/) is officially archived.

daniel-sanche:
Closing in favor of https://github.com/googleapis/google-auth-library-python/pull/1929

PR #1776, opened 2025-06-02
fix: Fetch token from default endpoint for GCE

harkamaljot:
After unnessary call to mds was removed here: https://github.com/googleapis/google-auth-library-python/pull/1769/files we realized downstream customers might have created an inherent dependency on how this refresh method works. Therefore to make it a non breaking change for customers, we will only fetch the token using "default" service account email. For other purposes, the call is still available for customers.


google/auth/compute_engine/credentials.py
```diff
@@ -104,8 +121,10 @@ def refresh(self, request):
         """
         scopes = self._scopes if self._scopes is not None else self._default_scopes
         try:
+            self._retrieve_info(request)
+            # Always fetch token with default service account email.
             self.token, self.expiry = _metadata.get_service_account_token(
-                request, service_account=self._service_account_email, scopes=scopes
+                request, service_account="default", scopes=scopes
```
sai-sunder-s, line 127: If user provides email, we do have to use that?

sai-sunder-s (changes requested):

google/auth/compute_engine/credentials.py
harkamaljot: As only 1 email can be associated/active with a gce/gke instance so fetching the token using "default" will always work. Hence, it shouldn't make any difference here?

harkamaljot:
closing this PR out as it's cleaner to revert that change and send out just the token change here.

PR #1778, opened 2025-06-03
feat: Add trust boundary support for service accounts and impersonation.

nbayati:
design: [go/trust-boundaries-auth-sdk-v2](http://goto.google.com/trust-boundaries-auth-sdk-v2)


google/auth/credentials.py
```diff
@@ -299,6 +289,124 @@ def with_universe_domain(self, universe_domain):
...
+            return
+        new_trust_boundary = {}
+        try:
+            new_trust_boundary = self._lookup_trust_boundary(request)
+        except exceptions.RefreshError as error:
+            # If the call to the lookup API failed, check if there is a trust boundary
+            # already cached. If there is, do nothing. If not, then throw the error.
+            if self._trust_boundary is None:
+                raise (error)
+            return
```
sai-sunder-s, line 391: add a debug log in case we used cached value. If not there might be error received forever and when trying to debug something related to trust boundary, we will not know that the requests have been failing.

google/auth/credentials.py
```diff
@@ -299,6 +289,124 @@ def with_universe_domain(self, universe_domain):
...
+        specific URL based on the credential type and its properties.
+
+        Returns:
+            str: The URL for the trust boundary lookup endpoint, or None
+                 if lookup should be skipped (e.g., for non-applicable universe domains).
+        """
+        raise NotImplementedError("_build_trust_boundary_lookup_url must be implemented")
+
+    @staticmethod
+    def _parse_trust_boundary(trust_boundary_string: str):
```
sai-sunder-s: where is this used?

google/auth/credentials.py
```diff
@@ -26,6 +28,9 @@
 from google.auth._refresh_worker import RefreshThreadManager
 
 DEFAULT_UNIVERSE_DOMAIN = "googleapis.com"
+NO_OP_TRUST_BOUNDARY_LOCATIONS: "typing.Tuple[str]" = ()
+NO_OP_TRUST_BOUNDARY_ENCODED_LOCATIONS = "0x0"
+TRUST_BOUNDARY_ENV_VAR = "GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED"
```
sai-sunder-s: this belongs in environment_vars.py

google/auth/impersonated_credentials.py
```diff
@@ -331,6 +341,28 @@ def _update_token(self, request):
...
+        and service account email.
+
+        Raises:
+            ValueError: If `self.service_account_email` is None or an empty
+                string, as it's required to form the URL.
+
+        Returns:
+            str: The URL for the trust boundary lookup endpoint.
+        """
+        if not self.service_account_email:
```
sai-sunder-s, line 353: where is self.service_account_email initialized? Are you looking for _target_principal?

google/oauth2/service_account.py
```diff
@@ -84,13 +84,17 @@
 
 _DEFAULT_TOKEN_LIFETIME_SECS = 3600  # 1 hour in seconds
 _GOOGLE_OAUTH2_TOKEN_ENDPOINT = "https://oauth2.googleapis.com/token"
+_TRUST_BOUNDARY_LOOKUP_ENDPOINT = (
+    "https://iamcredentials.{}/v1/projects/-/serviceAccounts/{}/allowedLocations"
```
sai-sunder-s, line 88: this url seems to be the same for impersonated cred as well. Will it be same for all supported cred types?

tests/test_impersonated_credentials.py
```diff
@@ -143,6 +148,7 @@ def make_credentials(
         target_principal=TARGET_PRINCIPAL,
         subject=None,
         iam_endpoint_override=None,
+        trust_boundary=NO_OP_TRUST_BOUNDARY,
```
sai-sunder-s: all tests are using NO_OP_TRUST_BOUNDARY? We have to test valid trust boundaries as well.

sai-sunder-s: same for service account tests as well

sai-sunder-s (changes requested):

google/oauth2/service_account.py
nbayati: No, Workforce and workload each will have their own url as well.

google/auth/credentials.py
nbayati: Thanks for catching this. It's no longer used so I went ahead and removed it.

google/auth/impersonated_credentials.py
nbayati: The `service_account_email` property returns the value of the internal attribute `self._target_principal`:

https://github.com/googleapis/google-auth-library-python/blob/ca94ead4035beea4741dc5384449032f8e6f75d8/google/auth/impersonated_credentials.py#L372-L374

lsirac (commented):
Can you handle feature gating the same way we do for pluggable auth? We should align this across all languages.

Also, please add the following test cases:

* Please add a test scenario for both service_account.Credentials and impersonated_credentials.Credentials where a valid (non-noop) trust boundary is successfully fetched and cached on the first refresh. A second refresh should then mock a RefreshError from the lookup endpoint. The test should assert that no exception is raised and that the initially cached trust boundary data is preserved on the credentials object. This would validate the fallback logic in _refresh_trust_boundary.

* Include explicit tests for both service_account.Credentials in tests/oauth2/test_service_account.py and impersonated_credentials.Credentials in tests/test_impersonated_credentials.py to confirm that the trust boundary lookup is skipped when the credential is instantiated with a non-default universe_domain. This ensures the check is correctly triggered by the concrete classes.

google/auth/compute_engine/credentials.py
```diff
@@ -111,6 +119,33 @@ def refresh(self, request):
...
+    def _build_trust_boundary_lookup_url(self):
+        """Builds and returns the URL for the trust boundary lookup API for GCE."""
+        # If the service account email is 'default', we need to get the
+        # actual email address from the metadata server.
+        if self._service_account_email == "default":
+            from google.auth.transport import requests as google_auth_requests
+
+            request = google_auth_requests.Request()
+            try:
+                info = _metadata.get_service_account_info(request, "default")
```
sai-sunder-s, line 158: @harkamaljot you had worked on optimizing this call somewhere else in the code recently. Can you PTAL at this method and see if there are any issues in doing this call or something can be optimized.

google/auth/compute_engine/credentials.py
```diff
@@ -179,9 +216,23 @@ def with_universe_domain(self, universe_domain):
...
+    @_helpers.copy_docstring(credentials.CredentialsWithTrustBoundary)
+    def with_trust_boundary(self, trust_boundary):
+        creds = self.__class__(
+            service_account_email=self._service_account_email,
+            quota_project_id=self._quota_project_id,
+            scopes=self._scopes,
+            default_scopes=self._default_scopes,
+            trust_boundary=trust_boundary,
+        )
+        creds._universe_domain = self._universe_domain
```
sai-sunder-s: why not initialize universe domain through the creds init

google/auth/credentials.py
```diff
@@ -299,6 +289,128 @@ def with_universe_domain(self, universe_domain):
...
+        """
+        raise NotImplementedError("This credential does not support trust boundaries.")
+
+    def apply(self, headers, token=None):
+        """Apply the token to the authentication header."""
+        super().apply(headers, token)
+        if self._trust_boundary is not None:
+            if (
+                self._trust_boundary["encodedLocations"]
+                == NO_OP_TRUST_BOUNDARY_ENCODED_LOCATIONS
```
sai-sunder-s: can use _has_no_op_trust_boundary here?

google/auth/compute_engine/credentials.py
nbayati: Thanks for the suggestion! That's a great point. I've updated the code to pass universe_domain to the constructor and applied the same pattern to the other with_* methods for consistency.

google/auth/_helpers.py
```diff
@@ -287,6 +288,45 @@ def unpadded_urlsafe_b64encode(value):
     return base64.urlsafe_b64encode(value).rstrip(b"=")
 
 
+def get_bool_from_env(variable_name, default=False):
+    """Gets a boolean value from an environment variable.
+
+    The environment variable is interpreted as a boolean with the following
+    (case-insensitive) rules:
+    - "true", "1" are considered true.
+    - "false", "0" are considered false.
```
lsirac, line 297: any other values -> exception is thrown

google/auth/external_account.py
lsirac, line 1: Lets revert these changes and keep them for a future PR. We should only add this with the changes for supporting the feature

google/auth/impersonated_credentials.py
```diff
@@ -251,13 +259,15 @@ def __init__(
...
         self._cred_file_path = None
+        self._trust_boundary = trust_boundary
 
     def _metric_header_for_usage(self):
         return metrics.CRED_TYPE_SA_IMPERSONATE
 
     @_helpers.copy_docstring(credentials.Credentials)
     def refresh(self, request):
         self._update_token(request)
+        self._refresh_trust_boundary(request)
```
lsirac: One suggestion for a potential refactoring to improve the design:

Currently, each class that inherits from `CredentialsWithTrustBoundary` (e.g., `compute_engine.Credentials`, `impersonated_credentials.Credentials`, `service_account.Credentials`) is required to explicitly call `self._refresh_trust_boundary(request)` within its own `refresh` method.

To enhance encapsulation and reduce the burden on subclasses, it would be preferable if this call was handled automatically by the base class. Could this be refactored?

google/oauth2/_client.py
```diff
@@ -506,3 +513,138 @@ def refresh_grant(
...
+        url (str): The trust boundary lookup url.
+        can_retry (bool): Enable or disable request retry behavior. Defaults to true.
+        headers (Optional[Mapping[str, str]]): The headers for the request.
+        kwargs: Additional arguments passed on to the request method. The
+            kwargs will be passed to `requests.request` method, see:
+            https://docs.python-requests.org/en/latest/api/#requests.request.
+            For example, you can use `cert=("cert_pem_path", "key_pem_path")`
+            to set up client side SSL certificate, and use
+            `verify="ca_bundle_path"` to set up the CA certificates for sever
+            side SSL certificate verification.
```
lsirac: I don't think we should include this. There isn't really anything that can be added here

google/oauth2/service_account.py
lsirac, line 1: How does this work for self signed JWTs?

google/oauth2/_client.py
```diff
@@ -510,7 +515,7 @@ def refresh_grant(
     return _handle_refresh_grant_response(response_data, refresh_token)
 
 
-def lookup_trust_boundary(request, url, access_token):
+def lookup_trust_boundary(request, url, headers=None):
```
lsirac: Shouldn't this be _lookup_trust_boundary?

Also, should this method handle adding auth to the call, instead of that burden being handled by callers?

lsirac (changes requested):
This is looking good! The main thing is to refactor this, if possible, so that the burden of refreshing trust boundaries is handled by the base class.

For better test coverage, we should add a few test cases for some edge scenarios.
- For the new `with_trust_boundary()` method, it would be great to have tests for each credential type confirming that it returns a new instance and correctly copies all other attributes, modifying only the trust boundary itself.

- What is the expected behavior if `impersonated_credentials` are created with `source_credentials` that don't have trust boundary support? We should probably add a test to ensure the `x-allowed-locations` header is not sent in that case. 

- For `service_account.Credentials` with `always_use_jwt_access=True`, does the trust boundary lookup still work as expected?

- A test case for when `_build_trust_boundary_lookup_url` is called when `service_account_email` is `None` would be valuable. The code seems to handle this by raising a `ValueError`, and it's worth having a test to confirm this behavior.

google/auth/compute_engine/credentials.py
harkamaljot: I optimized this call when getting token for the service account, however for getting the trust boundary looks like we need service account email. Based on the documentation(http://cloud/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/getAllowedLocations) and also trying it out manually as well, I get 400 error when querying this endpoint using default.

google/oauth2/_client.py
nbayati: Renamed the method to _lookup_trust_boundary.

Regarding where the auth header should be added, I believe it's best to keep that responsibility with the caller via its apply() method.

The current design, where the Credentials object calls self.apply(headers) and then passes the fully-formed headers to _lookup_trust_boundary, ensures that all necessary headers are applied consistently from a single place. This is important because apply() can add more than just the Authorization header (for example, it also adds the x-allowed-locations header for the trust boundary itself). This keeps the concerns cleanly separated and avoids duplicating header logic, or potentially missing adding a header.

google/auth/impersonated_credentials.py
nbayati: That's a great suggestion! 
I've refactored the CredentialsWithTrustBoundary base class to handle this. The base refresh() method now handles the entire refresh process. It calls a new abstract method, _refresh_token(), which subclasses must implement, and then it calls _refresh_trust_boundary() itself.

google/oauth2/service_account.py
nbayati: I've updated service_account.Credentials to handle self signed JWTs. When a credential that uses a self-signed JWT also needs to perform a trust boundary lookup, the refresh() method now follows a two-step process:

1. IAM-Specific JWT for Lookup: It first generates a temporary self-signed JWT where the audience is specifically set to the IAM Credentials API. This token is used exclusively to authenticate the request that fetches the trust boundary information.

2. Final JWT for Target API: After the trust boundary is successfully retrieved, the method then proceeds to generate the final self-signed JWT with the audience required by the application's target API (e.g., Pub/Sub, Storage).

This approach ensures that the trust boundary lookup is authenticated correctly using a token the IAM service expects, while the final token used by the application is properly configured for its intended service.

nbayati:
> * What is the expected behavior if `impersonated_credentials` are created with `source_credentials` that don't have trust boundary support? We should probably add a test to ensure the `x-allowed-locations` header is not sent in that case.
I agree that the header shouldn't be sent in those cases. Added a unit test to cover this scenario. 

> * For `service_account.Credentials` with `always_use_jwt_access=True`, does the trust boundary lookup still work as expected?
Added some logic to support trust boundaries for the self signed jwt flow.

Also implemented the suggested refactoring and test scenarios. 
Will update the design to get clarification on some of the credential types we're not sure about, like the idtoken.

google/auth/credentials.py
```diff
@@ -299,6 +289,161 @@ def with_universe_domain(self, universe_domain):
...
+            request (google.auth.transport.Request): The object used to make
+                HTTP requests.
+
+        Raises:
+            google.auth.exceptions.RefreshError: If the credentials could
+                not be refreshed.
+        """
+        raise NotImplementedError("_refresh_token must be implemented")
+
+    def with_trust_boundary(self, trust_boundary):
```
lsirac, line 307: Is this method necessary? When would it be used?

tests/test_external_account.py
```diff
@@ -920,7 +920,9 @@ def test_refresh_impersonation_without_client_auth_success(
             "Content-Type": "application/json",
             "authorization": "Bearer {}".format(token_response["access_token"]),
             "x-goog-api-client": IMPERSONATE_ACCESS_TOKEN_REQUEST_METRICS_HEADER_VALUE,
-            "x-allowed-locations": "0x0",
+            # TODO(negarb): Uncomment and update when trust boundary is supported
```
lsirac, line 923: Would prefer to not have commented placeholders here and just add in another PR

tests/test_aws.py
```diff
@@ -2057,7 +2059,9 @@ def test_refresh_success_with_impersonation_ignore_default_scopes(
             "authorization": "Bearer {}".format(self.SUCCESS_RESPONSE["access_token"]),
             "x-goog-user-project": QUOTA_PROJECT_ID,
             "x-goog-api-client": IMPERSONATE_ACCESS_TOKEN_REQUEST_METRICS_HEADER_VALUE,
-            "x-allowed-locations": "0x0",
+            # TODO(negarb): Uncomment and update when trust boundary is supported
```
lsirac, line 2060: Same here

lsirac (dismissed):
Looks good, great job!

tests/test_external_account.py
nbayati: Thanks for the feedback—I agree it's best to avoid placeholders.

The challenge here is that the commented-out value ("x-allowed-locations": "0x0") is from two years ago and is incorrect under the new design.. Since support for external accounts is out of scope for this particular PR, uncommenting this line would cause tests to fail with the wrong assertion.

I've kept it commented to preserve the test structure for the follow-up PR that will correctly implement this feature for external accounts.

Given this context, would you be okay with keeping these placeholders for this PR, or would you prefer I remove the them and create a separate tracking issue?

google/auth/credentials.py
nbayati: The with_* method is part of a design pattern used for credential objects throughout this library.

We treat credential objects as mostly immutable. You'll see similar factory methods like with_scopes(), with_quota_project(), and with_universe_domain().

google/auth/impersonated_credentials.py
```diff
@@ -570,6 +602,7 @@ def refresh(self, request):
             "Content-Type": "application/json",
             metrics.API_CLIENT_HEADER: metrics.token_request_id_token_impersonate(),
         }
+        headers.update(self._target_credentials._get_trust_boundary_header())
```
sai-sunder-s: Why do we have to pass trust boundary header for id token request? `_get_trust_boundary_header` might not give the headers because an access token is not available in this credential.

google/oauth2/service_account.py
```diff
@@ -451,6 +462,57 @@ def refresh(self, request):
...
+            # 1. Create a temporary self-signed JWT for the IAM API.
+            iam_audience = "https://iamcredentials.{}/".format(self._universe_domain)
+            iam_jwt_creds = jwt.Credentials.from_signing_credentials(self, iam_audience)
+            iam_jwt_creds.refresh(request)
+
+            # 2. Use this JWT to perform the trust boundary lookup.
+            # We temporarily set self.token for the base lookup method.
+            # The base lookup method will call self.apply() which adds the
+            # authorization header.
+            original_token = self.token
```
sai-sunder-s: This seems to be welcoming race conditions. What happens if one thread is updating trust boundary and set `self.token` to iam audience and another thread tries to read `self.token`

google/oauth2/service_account.py
```diff
@@ -806,13 +920,17 @@ def _refresh_with_iam_endpoint(self, request):
             additional_claims={"scope": "https://www.googleapis.com/auth/iam"},
         )
         jwt_credentials.refresh(request)
+
+        headers = self._get_trust_boundary_header()
```
sai-sunder-s: you will get NotImplementedError here, thrown from ln 863?

sai-sunder-s: Also, there is no access token in this cred to call trust boundary endpoint

sai-sunder-s (changes requested):

google/auth/impersonated_credentials.py
nbayati: You're right. Thank you for catching this. Initially we thought idtoken would need trust boundary too, but we determined that it's not a supported credential and I reverted these changes.

google/oauth2/service_account.py
nbayati: Reverted this change since we're still waiting to hear back from the backend team on whether or not the workaround is necessary. We'll address self signed jwt flow if necessary in a future PR.

google/oauth2/service_account.py
nbayati: I reverted the changes to idtoken as it turns out it doesn't need trust boundaries. Thanks for flagging it!

lsirac (dismissed):

sai-sunder-s (dismissed):

sai-sunder-s (approved):

PR #1779, opened 2025-06-03
fix: auth fetch token from default endpoint

harkamaljot:
Fetch the token using "default" service account email. For other purposes, the call is still available for customers. As sometimes if a customer has overridden the principal to something else then this call fails. Therefore, to avoid such cases fetching the token using "default" endpoint. The reason this change works is because mds is associated with only 1 service account, so we do not need to provide an email address for fetching the token.


sai-sunder-s:
Can you push this change to gce vm which has a non-default SA attached to it and verify the change works?

sai-sunder-s (approved):

PR #1798, opened 2025-07-30
fix: Use `user_verification=preferred` for ReAuth WebAuthn challenge

darkfeline:
Since ReAuth is a second factor credential, it is not necessary to require UV here.  This was discussed with ReAuth folks.

Also, in practice, downstream clients disregard this because the U2F protocol doesn't expose UV enforcement.


darkfeline:
@cpisunyer who added this originally.  Apologies if you're the wrong person to tag or if I missed some contributor task.

gkevinzheng (dismissed):

chalmerlowe (commented):
LGTM

parthea (approved):

chalmerlowe (approved):
LGTM

PR #1809, opened 2025-08-22
feat: Add trust boundary support for external accounts.

nbayati:
Implements the trust boundary feature for external accounts (workforce and workload pool identity, and also authorized user)
Note: The design has changed since the first PR that implemented trust boundary for service accounts, and we are no longer required to send the allowed locations header to IAM or STS requests, only to the google API calls. You can review the most up to date design here: [go/trust-boundaries-auth-sdk-v2](http://goto.google.com/trust-boundaries-auth-sdk-v2)


tests/test_external_account_authorized_user.py
```diff
@@ -557,3 +576,51 @@ def test_from_file_full_options(self, tmpdir):
...
+            credentials.refresh(request)
+
+        mock_lookup.assert_not_called()
+        headers = {}
+        credentials.apply(headers)
+        assert "x-allowed-locations" not in headers
+
+    def test_build_trust_boundary_lookup_url(self):
+        credentials = self.make_credentials()
+        expected_url = "https://iamcredentials.googleapis.com/v1/locations/global/workforcePools/POOL_ID/allowedLocations"
```
sai-sunder-s, line 620: Add few more cases:
- rep instead of global
- different universe

tests/test_external_account_authorized_user.py
```diff
@@ -557,3 +576,51 @@ def test_from_file_full_options(self, tmpdir):
...
+        assert credentials._build_trust_boundary_lookup_url() == expected_url
+
+    def test_build_trust_boundary_lookup_url_invalid_audience(self):
+        credentials = self.make_credentials(audience="invalid")
```
sai-sunder-s: add cases with real looking url but not have expected pattern

google/auth/impersonated_credentials.py
```diff
@@ -278,7 +278,7 @@ def _refresh_token(self, request):
             self._source_credentials.token_state == credentials.TokenState.STALE
             or self._source_credentials.token_state == credentials.TokenState.INVALID
         ):
-            self._source_credentials.refresh(request)
+            self._source_credentials._refresh_token(request)
```
sai-sunder-s, line 289: why do we need to get the turst boundary locations for source cred? It will be the outer (impersonated) cred that will be adding the headers for the API calls.

nbayati: `_refresh_token()` is just refreshing the token and won't fetch the trust boundary data. 
In the scenario where an external account is impersonating a service account, we only lookup the trust boundary of the target service account.

tests/test_external_account_authorized_user.py
nbayati: Added a test for a different universe. I'm waiting to hear from the backend team to see if they even accept a rep or if the url has to be global. Depending on the conversation, we can address it in a feature PR to make this feature more future-proof.

sai-sunder-s (dismissed):

lsirac (dismissed):

sai-sunder-s (approved):

PR #1811, opened 2025-09-15
feat: Introduce a way to provide custom headers

sai-sunder-s:
For certain scenarios, custom headers are required to be sent to the backend along with usual headers like `authorization`, `x-goog-user-project`. Exposing a way for users to specify these custom headers.


google/auth/credentials.py
```diff
@@ -233,6 +235,38 @@ def before_request(self, request, method, url, headers):
...
+
+        Returns:
+            google.auth.credentials.Credentials: A new credentials instance.
+
+        Raises:
+            ValueError: If a protected header is included in the input headers.
+        """
+        import copy
+
+        PROTECTED_HEADERS = {
```
nolanleastin, line 252: Is there a way to capture this in a more central way? I worry that we are missing some now or may miss some in the future if new ones get added

for example https://github.com/googleapis/google-auth-library-python/blob/c30a6a781d3e385598a0ac28a370a7f4800010cc/google/auth/compute_engine/_metadata.py#L54

tests/test_credentials.py
```diff
@@ -71,6 +71,48 @@ def test_with_non_blocking_refresh():
...
+    credentials = CredentialsImpl()
+    request = mock.Mock()
+
+    # 1. Add a new custom header
+    creds_with_header = credentials.with_headers({"X-Custom-Header": "value1"})
+    headers = {}
+    creds_with_header.before_request(request, "http://example.com", "GET", headers)
+    assert headers["X-Custom-Header"] == "value1"
+    assert "authorization" in headers  # Ensure base apply logic ran
+    assert creds_with_header is not credentials
```
nolanleastin, line 85: Not sure what you're trying to test here but as I understand python, this will never be true.

i.e. if you did:
```
credentials_a = CredentialsImpl()
credentials_b = CredentialsImpl()
assert credentials_a is not credentials_b # success 
```

tests/test_credentials.py
```diff
@@ -71,6 +71,48 @@ def test_with_non_blocking_refresh():
     assert c._use_non_blocking_refresh
 
 
+def test_with_headers():
```
nolanleastin, line 75: nit: this test is testing multiple scenarios. could you break it up so that we are testing ~1 thing per test? 

https://engdoc.corp.google.com/eng/doc/tott/episodes/726.md?cl=head

google/auth/credentials.py
```diff
@@ -69,6 +69,7 @@ def __init__(self):
 
         self._use_non_blocking_refresh = False
         self._refresh_worker = RefreshThreadManager()
+        self._custom_headers = {}
```
nolanleastin, line 72: This is not used?

chalmerlowe:
Discussed with @parthea.
Will close this PR for now, as we are attempting to clear the books so that we can successfully migrate this repo to the google-cloud-python mono-repo.

Have filed bug #1911 (that will be carried with this repo into the mono-repo) to revisit this PR after the migration.

PR #1812, opened 2025-09-23
fix: Deprecating generic load methods and adding warnings on few cred types

sai-sunder-s:
This PR deprecates methods that load any credential type from a given json or file, specifically `load_credentials_from_file` and `load_credentials_from_dict`. Instead, it is recommended to use the load methods specific to a credential type. It is recommended to perform validation when a credential config is obtained from external sources. Please read the deprecation warning message and the warnings on appropriate credential types for the full context.

BEGIN_COMMIT_OVERRIDE
fix: Deprecating [load_credentials_from_file](https://googleapis.dev/python/google-auth/latest/reference/google.auth.html#google.auth.load_credentials_from_file)
fix: Deprecating [load_credentials_from_dict](https://googleapis.dev/python/google-auth/latest/reference/google.auth.html#google.auth.load_credentials_from_dict)
END_COMMIT_OVERRIDE


google/auth/_default.py
```diff
@@ -59,10 +59,50 @@
...
+
+Refer to https://cloud.google.com/docs/authentication/external/externally-sourced-credentials
+for more details.
+"""
+
 # The subject token type used for AWS external_account credentials.
 _AWS_SUBJECT_TOKEN_TYPE = "urn:ietf:params:aws:token-type:aws4_request"
 
 
+class GenericLoadMethodWarning(DeprecationWarning):  # pragma: NO COVER
```
sai-sunder-s: @parthea is it required to create a class extending from DeprecationWarning or is it ok to use DeprecationWarningDirectly?

google/auth/_default.py
```diff
@@ -59,6 +59,38 @@
 https://cloud.google.com/docs/authentication/adc-troubleshooting/user-creds. \
 """
 
+_GENERIC_LOAD_METHOD_WARNING = """\
+The {} method is being deprecated because of a potential security risk.
```
chalmerlowe: #QUESTION:

Are we using the right wording OR does the current wording provide the right connotation?

When I first saw the words `is being deprecated` I felt like we were insinuating that this deprecation would take place in the future.

Would this wording be a more assertive explanation?

```suggestion
The {} method is deprecated because of a potential security risk.
```

google/auth/_default.py
```diff
@@ -76,6 +108,20 @@ def _warn_about_problematic_credentials(credentials):
...
+
+    This is to discourage use of the generic load methods in favor of
+    more specific methods. The generic methods are more likely to lead to
+    security issues if the input is not validated.
+
+    Args:
+        method_name (str): The name of the method being used.
+    """
+
+    warnings.warn(_GENERIC_LOAD_METHOD_WARNING.format(method_name), DeprecationWarning)
```
chalmerlowe, line 122: QUESTION:

There are multiple [types of deprecation warnings](https://docs.python.org/3/library/warnings.html#warning-categories). They cause different behaviors in terms of who sees them. Are our warnings intended for Devs or for End Users? If end users, we may want to consider a different warning class.

| Warning Type | Description | Audience |
|---|---|---|
| DeprecationWarning | Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__). | Devs |
| FutureWarning | Base category for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python. | End user |

chalmerlowe (dismissed):
UPDATE: sigh. turns out I am not in the list of allowed approvers. sorry. i tried...

~~Pending changes, if needed, I am Approving.
You end up change something and the Approval gets clobbered, ping me. I will gladly pop back in and re-approve.~~

google/auth/_default.py
sai-sunder-s: Thank you for bringing this up. The intended audience is the developer but I do not like the "ignored by default" part.  `DeprecationWarning` seems to be used in other parts of the codebase though, so maybe ok? Open to suggestions @chalmerlowe and @parthea

google/auth/external_account.py
```diff
@@ -89,6 +89,14 @@ class Credentials(
...
     The base class implements the common logic for exchanging external account
     credentials for Google access tokens.
+
+    **IMPORTANT**:
+    This method does not validate the credential configuration. A security
+    risk occurs when a credential configuration configured with malicious urls
+    is used.
+    When the credential configuration is accepted from an
+    untrusted source, you should validate it before using with this method.
+    Refer https://cloud.google.com/docs/authentication/external/externally-sourced-credentials for more details.
```
parthea: Will this entire class be removed? The text should be updated since it says `This method` rather than `This class`

sai-sunder-s: Class/method not going to be removed. This is just a wanring.

Deprecation is in _default.py only

parthea (dismissed):

parthea (approved):

PR #1821, opened 2025-10-08
feat: Add support for Agent Identity bound tokens

nbayati:
This change introduces support for requesting certificate-bound access tokens for Agent Identities  on GKE and Cloud Run. The design doc: [go/sdk-agent-identity](http://goto.google.com/sdk-agent-identity)


google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    Args:
+        cert_bytes (bytes): The PEM-encoded certificate bytes.
+
+    Returns:
+        str: The base64-encoded SHA256 fingerprint.
+    """
+    from cryptography import x509
+    from cryptography.hazmat.primitives import serialization
+
+    cert = x509.load_pem_x509_certificate(cert_bytes)
```
sai-sunder-s: this operation is done in _is_agent_identity_certificate as well. Is loading the cert bytes an expensive operation? if so, better to load once and pass the object around

google/auth/compute_engine/credentials.py
```diff
@@ -135,9 +135,9 @@ def _refresh_token(self, request):
                 service can't be reached if if the instance has not
                 credentials.
         """
-        scopes = self._scopes if self._scopes is not None else self._default_scopes
         try:
             self._retrieve_info(request)
+            scopes = self._scopes if self._scopes is not None else self._default_scopes
```
sai-sunder-s, line 140: why is this line being moved?

Did Gemini do it ;)

nbayati: Haha no, this change was intentional.
My new unit test `test_refresh_with_agent_identity` in `tests/compute_engine/test_credentials.py` discovered an existing bug.  The issue is that the code was assigning the scopes before calling _retrieve_info(). However, _retrieve_info() is what actually fetches the scopes from the metadata server if they aren't already set. This would be problematic in cases where scopes are not specified during credential creation. By moving this line we ensure that the scopes are fetched and then assigned.

sai-sunder-s: Makes sense. Thank you for fixing it!

google/auth/_agent_identity_utils.py
nbayati: Done!

sai-sunder-s (dismissed):

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,165 @@
...
+                )
+                if cert_path and os.path.exists(cert_path):
+                    return cert_path
+        if i < 50:
+            time.sleep(0.1)
+        else:
+            time.sleep(0.5)
+
+    raise exceptions.RefreshError(
+        "Certificate config or certificate file not found after multiple retries. "
```
lsirac, line 123: I thought we would point to some URL here?

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,165 @@
...
+
+    cert_config_path = os.environ.get(environment_vars.GOOGLE_API_CERTIFICATE_CONFIG)
+    if not cert_config_path:
+        return None
+
+    # Poll for the config file and the certificate file to be available.
+    # Phase 1: Poll rapidly for 5 seconds (50 * 0.1s).
+    # Phase 2: Slow down polling for the next 25 seconds (50 * 0.5s).
+    for i in range(100):
+        if os.path.exists(cert_config_path):
```
lsirac: If the platform is in the middle of writing this file when the poll happens, it might be empty or malformed JSON, causing json.load to raise an exception and crash the entire retry loop prematurely.

I don't know if atomic writes are guaranteed (?).. if it opens the target file directly for writing, os.path.exists returns True (?) the moment the file is created, even if it currently has 0 bytes or incomplete data.

I think we either need to clarify or assume the worst-case scenario (non-atomic writes) and defend against it. crashing the entire 30-second wait because we caught the file during the 10ms it was being written would be unfortunate!

tests/test_agent_identity_utils.py
```diff
@@ -0,0 +1,212 @@
...
+    b"odJMlWrJBtynn/TA1oJlYu3yV5clc/71Vr/AxuX5xGP+IXL32YDF9lTUJXG/uUGk\n"
+    b"+JETpKmQviPbRsvzYhz4pf6ZIOZMc3/GIcNq92ECbseGO+yAgyWUVKMmZM0HqXC9\n"
+    b"ovNslqe0M8C1sLm1zAR5z/h/litE7/8O2ietija3Q/qtl2TOXJdCA6sgjJX2WUql\n"
+    b"ybrC55ct18NKf3qhpcEkGQvFU40rVYApJpi98DiZPYFdx1oBDp/f4uZ3ojpxRVFT\n"
+    b"cDwcJLfNRCPUhormsY7fDS9xSyThiHsW9mjJYdcaKQkwYZ0F11yB\n"
+    b"-----END CERTIFICATE-----\n"
+)
+
+
+class TestAgentIdentityUtils:
```
lsirac, line 53: We should add a test that asserts that the raised RefreshError contains the the expected text re "GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES" (if thats the direction we're going)

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,165 @@
...
+    import json
+
+    cert_config_path = os.environ.get(environment_vars.GOOGLE_API_CERTIFICATE_CONFIG)
+    if not cert_config_path:
+        return None
+
+    # Poll for the config file and the certificate file to be available.
+    # Phase 1: Poll rapidly for 5 seconds (50 * 0.1s).
+    # Phase 2: Slow down polling for the next 25 seconds (50 * 0.5s).
+    for i in range(100):
```
lsirac: lets define these:

Fast poll: wait for typical startup (e.g., 5 seconds)
_FAST_POLL_INTERVAL = 0.1
_FAST_POLL_CYCLES = 50  # 50 * 0.1 = 5 seconds

Slow poll: wait for stragglers up to total 30s timeout (e.g., remaining 25s)
_SLOW_POLL_INTERVAL = 0.5
_SLOW_POLL_CYCLES = 50  # 50 * 0.5 = 25 seconds

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,165 @@
...
+        if os.path.exists(cert_config_path):
+            with open(cert_config_path, "r") as f:
```
lsirac, line 90: Should use binary mode: open(cert_path, "rb").

Maybe this should be defined as a utility somewhere

lsirac (changes requested):

google/auth/_agent_identity_utils.py
nbayati: The URL would be for the 401 error code, pointing to a troubleshooting guide that lists the possible reasons why they are getting an error since we can't determine if it's agent identity or something else (Scenario 2 and 3 from the [go/client-side-error-handling-for-agentic](http://goto.google.com/client-side-error-handling-for-agentic) doc)

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,165 @@
...
+                    cert_config.get("cert_configs", {})
+                    .get("workload", {})
+                    .get("cert_path")
+                )
+                if cert_path and os.path.exists(cert_path):
+                    return cert_path
+        if i < 50:
+            time.sleep(0.1)
+        else:
+            time.sleep(0.5)
```
daniel-sanche: nit: maybe instead of using conditional sleeps here, you can generate the sleep values directly? Something like:

```
BACKOFF_VALUES = [0.1] * 50 + [0.5] * 50

for sleep_value in BACKOFF_VALUES:
   ...
   time.sleep(sleep_value)
```

google/auth/_agent_identity_utils.py
daniel-sanche: +1. It might be better to do a try/except block inside the loop, instead of the `os.path.exists` checks. Then you can catch FileNotFound, and any other transient error that may come up when reading

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,165 @@
...
+            ExtensionOID.SUBJECT_ALTERNATIVE_NAME
+        )
+    except x509.ExtensionNotFound:
+        return False
+    uris = ext.value.get_values_for_type(x509.UniformResourceIdentifier)
+
+    for uri in uris:
+        if uri.startswith("spiffe://"):
+            spiffe_id = uri[len("spiffe://") :]
+            trust_domain = spiffe_id.split("/", 1)[0]
```
daniel-sanche: nit: consider using `urllib.parse.urlparse` instead of hand-parsing the raw string directly

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Args:
+        cert (cryptography.x509.Certificate): The parsed certificate object.
+
+    Returns:
+        bool: True if the certificate is an Agent Identity certificate,
+            False otherwise.
+    """
+    from cryptography import x509
+    from cryptography.x509.oid import ExtensionOID
```
daniel-sanche: IIUC it looks like cryptography is an optional dependency. Did you handle the case where this import fails?

Is there a chance existing workflows start calling any of these new methods, and encountering exceptions?

google/auth/_agent_identity_utils.py
nbayati: Done!

google/auth/_agent_identity_utils.py
nbayati: We're not reading the certificate files here. This line is reading the json file and loading it with `json.load()`, which will expect a text file. So I think opening it in the read mode ("r") is the right approach here.

google/auth/_agent_identity_utils.py
nbayati: Thanks for catching this. You are correct, cryptography is an optional dependency. 
To address this, I've made a few changes:
 1.  I've added a new [cryptography] extra to setup.py.
 2.  All imports of the cryptography library are now wrapped in a try...except ImportError block. If the import fails, we raise a new ImportError with a helpful message instructing the user to install the dependency via pip install 
  google-auth[cryptography].
  
Regarding your second question about existing workflows, they are safe. The new code paths that require cryptography are only entered if the env has been configured to use a certificate based authentication by setting  the GOOGLE_API_CERTIFICATE_CONFIG env var.

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,211 @@
...
+                    "Certificate config file not found at %s (from %s environment "
+                    "variable). Retrying for up to %s seconds.",
+                    cert_config_path,
+                    environment_vars.GOOGLE_API_CERTIFICATE_CONFIG,
+                    _TOTAL_TIMEOUT,
+                )
+                has_logged_warning = True
+            pass
+
+        time.sleep(interval)
```
daniel-sanche, line 120: This doesn't matter at all, but I think the sleep could be inside the except, to remove that pass

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,211 @@
...
+    Returns:
+        cryptography.x509.Certificate: The parsed certificate object.
+    """
+    try:
+        from cryptography import x509
+    except ImportError:
+        raise ImportError(
+            "The cryptography library is required for certificate-based authentication."
+            "Please install it with `pip install google-auth[cryptography]`."
+        )
```
daniel-sanche: a `raise from` would be useful here, so the user can still see the original exception message as well

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,211 @@
...
+            False otherwise.
+    """
+    try:
+        from cryptography import x509
+        from cryptography.x509.oid import ExtensionOID
+    except ImportError:
+        raise ImportError(
+            "The cryptography library is required for certificate-based authentication."
+            "Please install it with `pip install google-auth[cryptography]`."
+        )
```
daniel-sanche: Should the error string be a constant, since it's re-used in a couple places?

google/auth/_agent_identity_utils.py
```diff
@@ -0,0 +1,211 @@
...
+        cert (cryptography.x509.Certificate): The parsed certificate object.
+
+    Returns:
+        bool: True if the certificate is an Agent Identity certificate,
+            False otherwise.
+    """
+    try:
+        from cryptography import x509
+        from cryptography.x509.oid import ExtensionOID
+    except ImportError:
```
daniel-sanche: Do we have any tests that try running these functions without cryptography imported?

I don't really have enough context to know if that would be necessary, but IIRC I think it's pretty easy to un-import for specific test functions, so maybe consider if it would be

google/auth/_agent_identity_utils.py
nbayati: Thanks for the suggestion! I've kept the time.sleep() outside the except block to handle two
  distinct scenarios while polling for the certificate:

   1. The certificate configuration file (GOOGLE_API_CERTIFICATE_CONFIG) is not yet available. This
      triggers the except block.
   2. The configuration file is available, but the certificate file path specified inside it is not
      yet on the filesystem. This case does not trigger an exception.

  In both scenarios, we need to pause and retry. If we move the sleep into the except block, there won't be any pause and sleep for the second scenario would .

  I've added a comment to the code to make this polling logic clearer for future developers.

google/auth/_agent_identity_utils.py
nbayati: Done!

tests/test_agent_identity_utils.py
```diff
@@ -0,0 +1,287 @@
...
+        assert not _agent_identity_utils._is_agent_identity_certificate(mock_cert)
+
+    def test__is_agent_identity_certificate_no_san(self):
+        mock_cert = mock.MagicMock()
+        mock_cert.extensions.get_extension_for_oid.side_effect = x509.ExtensionNotFound(
+            "Test extension not found", None
+        )
+        assert not _agent_identity_utils._is_agent_identity_certificate(mock_cert)
+
+    def test__is_agent_identity_certificate_not_spiffe_uri(self):
```
lsirac, line 93:
```suggestion
    def test_is_agent_identity_certificate_not_spiffe_uri(self):
```

tests/test_agent_identity_utils.py
```diff
@@ -0,0 +1,287 @@
...
+            "true",
+        )
+        mock_get_path.return_value = None
+        result = _agent_identity_utils.get_and_parse_agent_identity_certificate()
+        assert result is None
+        mock_get_path.assert_called_once()
+
+    @mock.patch("google.auth._agent_identity_utils.parse_certificate")
+    @mock.patch("google.auth._agent_identity_utils.get_agent_identity_certificate_path")
+    def test_get_and_parse_agent_identity_certificate_success(
```
lsirac, line 246: Lets add a test case to simulate a sequence of failures followed by success:

Iteration 1: os.path.exists(config) -> False
Iteration 2: os.path.exists(config) -> True, but open() raises IOError (incomplete write).
Iteration 3: open() succeeds, but json.load raises ValueError (malformed JSON).
Iteration 4: Everything succeeds.
Assert that get_agent_identity_certificate_path returns successfully after these 4 attempts

lsirac (dismissed):

daniel-sanche (dismissed):
LGTM

sai-sunder-s (dismissed):

PR #1822, opened 2025-10-08
feat: support Python 3.14

Linchin:


samples/cloud-client/snippets/requirements.txt
```diff
@@ -1,4 +1,5 @@
 google-cloud-compute==1.5.1
 google-cloud-storage==3.1.0
 google-auth==2.38.0
-pytest==7.1.2
+pytest===7.1.2; python_version <= '3.13'
+pytest==8.4.2; python_version == '3.14'
```
daniel-sanche: is this needed? Can we use 8.4.2 for all?

tests_async/transport/test_aiohttp_requests.py
```diff
@@ -115,10 +115,11 @@ def make_with_parameter_request(self):
         http = aiohttp.ClientSession(auto_decompress=False)
         return aiohttp_requests.Request(http)
 
-    def test_unsupported_session(self):
+    @pytest.mark.asyncio
+    async def test_unsupported_session(self):
         http = aiohttp.ClientSession(auto_decompress=True)
         with pytest.raises(ValueError):
-            aiohttp_requests.Request(http)
+            await aiohttp_requests.Request(http)
```
daniel-sanche, line 122: why were these tests changed to async?

samples/cloud-client/snippets/requirements.txt
parthea: 8.4.2 can be used for all. Resolving stale comment

tests_async/transport/test_aiohttp_requests.py
parthea: This change looks correct to me. My initial thought is that this should have been `async` based on this code

https://github.com/googleapis/google-auth-library-python/blob/9be69189061babf06abf356c82d23e4a2bdaf04b/google/auth/transport/_aiohttp_requests.py#L122-L153

vchudnov-g (approved):

PR #1842, opened 2025-10-20
fix: catch ValueError for json.loads()

Linchin:
see internal bug 448976223

TODO: 
 - [x] add test
 - [x] match the exception string so we don't catch unexpected cases


google/auth/_helpers.py
```diff
@@ -489,7 +489,7 @@ def _parse_request_body(body: Optional[bytes], content_type: str = "") -> Any:
     if not content_type or "application/json" in content_type:
         try:
             return json.loads(body_str)
-        except (json.JSONDecodeError, TypeError):
+        except (json.JSONDecodeError, TypeError, ValueError):
```
ohmayr: `ValueError` will catch `json.JSONDecodeError` as well. We can remove it since we're adding a base exception.

ohmayr (dismissed):

ohmayr (approved):

PR #1848, opened 2025-10-27
feat: Enable mTLS if GOOGLE_API_USE_CLIENT_CERTIFICATE  is not set, if the MWID/X.509 cert sources detected

agrawalradhika-cell:
The Python SDK will use a hybrid approach for mTLS enablement:

- If the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable is set (either true or false), the SDK will respect that setting. This is necessary for test scenarios and users who need to explicitly control mTLS behavior.
- If the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable is not set, the SDK will automatically enable mTLS only if it detects Managed Workload Identity (MWID) or X.509 Workforce Identity Federation (WIF) certificate sources. In other cases where the variable is not set, mTLS will remain disabled.

** This change also adds the helper method `check_use_client_cert`  and it's unit test,  which will be used for checking the criteria for setting the mTLS to true 
** This change is only for Auth-Library, other changes will be created for Client-Library use-cases.


google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,43 @@ def client_cert_callback():
 
     # Then dump the decrypted key bytes
     return crypto.dump_privatekey(crypto.FILETYPE_PEM, pkey)
+
+def check_use_client_cert():
+  """Returns the effective value of use_client_cert to be used.
+
+  Returns:
+      str:
+          A boolean indicating if client certificate should be used.
```
sai-sunder-s: Convert to a boolean instead of checking strings everywhere?

Will be a bit cleaner.

sai-sunder-s (dismissed):

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,43 @@ def client_cert_callback():
...
+          "true" if the certificate config file contains "workload" section
+          and "false" otherwise.
+
+  Raises:
+      ValueError: if GOOGLE_API_USE_CLIENT_CERTIFICATE is set to unsupported
+      value, which is not "true" or "false".
+  """
+  use_client_cert = os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE")
+  ### Check if the value of GOOGLE_API_USE_CLIENT_CERTIFICATE is unset.
+  if use_client_cert == "" or use_client_cert is None:
```
daniel-sanche: IIUC this could be simplified to `if not use_client_cert:`

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,43 @@ def client_cert_callback():
...
+    cert_path = os.getenv("GOOGLE_API_CERTIFICATE_CONFIG")
+    if cert_path:
+      with open(cert_path, "r") as f:
+        content = f.read()
+        if "workload" in content:
+          return "true"
+    return "false"
```
daniel-sanche: why return strings instead of bools?

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,43 @@ def client_cert_callback():
...
+  else:
+    ### Check if the value of GOOGLE_API_USE_CLIENT_CERTIFICATE is set but to an
+    ### invalid value.
+    use_client_cert = use_client_cert.lower()
+    if use_client_cert not in ("true", "false"):
+      raise ValueError(
+          "Environment variable `GOOGLE_API_USE_CLIENT_CERTIFICATE` must be"
+          " either `true` or `false`"
+      )
```
daniel-sanche: Are you sure raising an exception is needed here? We can't just default to false?

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,43 @@ def client_cert_callback():
...
+      ValueError: if GOOGLE_API_USE_CLIENT_CERTIFICATE is set to unsupported
+      value, which is not "true" or "false".
+  """
+  use_client_cert = os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE")
+  ### Check if the value of GOOGLE_API_USE_CLIENT_CERTIFICATE is unset.
+  if use_client_cert == "" or use_client_cert is None:
+    cert_path = os.getenv("GOOGLE_API_CERTIFICATE_CONFIG")
+    if cert_path:
+      with open(cert_path, "r") as f:
+        content = f.read()
```
daniel-sanche: reading the entire arbitrary file into memory here makes me a bit nervous. I assume we expect the file to be quite small. Can we add any constraints to this read?

I made a test config file, and it was 0.5kb. Maybe we can read up to n kb max, otherwise raise an error or something?

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,43 @@ def client_cert_callback():
...
+        if "workload" in content:
+          return "true"
```
daniel-sanche: is this robust? Isn't it likely for the word "workload" to appear in a file path, or another place other than where you expect?

If the file is kept small, maybe json.load could help parse this. If that's too heavy, you could use regex

google/auth/transport/grpc.py
```diff
@@ -256,9 +256,7 @@ def my_client_cert_callback():
 
     # If SSL credentials are not explicitly set, try client_cert_callback and ADC.
```
daniel-sanche, line 255: The docstrings for this method only mention raising MutualTLSChannelError, but check_use_client_cert raises a ValueError if the env var is wrong. A couple options:

1. add ValueError to the `Raises: ` section in the docstring
2. make the env var default to false, so it doesn't raise an exception
3. change the ValueError to MutualTLSChannelError or a subclass
4. catch the Value error, and use `raise ... from` to chain it into a MutualTLSChannelError or subclass


If you keep the exception in `_mtls_helper.check_use_client_cert`, you should mention that it can raise ValueError it in the `Raises:` section in the docstring. 

(Or, catch and chain it with MutualTLSChannelError, or even change ValueError MutualTLSChannelError, since that looks like it's meant to be an umbrella error)

daniel-sanche (changes requested):

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,43 @@ def client_cert_callback():
 
     # Then dump the decrypted key bytes
     return crypto.dump_privatekey(crypto.FILETYPE_PEM, pkey)
+
+def check_use_client_cert():
+  """Returns the effective value of use_client_cert to be used.
```
daniel-sanche: This docstring isn't very helpful. What is use_client_cert? What should its value be?

google/auth/transport/_mtls_helper.py
andyrzhao: +1 let's default to false for everything besides "true"

google/auth/transport/_mtls_helper.py
andyrzhao: +1 we should use json to parse this. The spec is well documented.

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Use cases of use_client_cert in this repo and other repos, which will be using this helper, have been checking for values true/false. Also, the env var is set in string format. Thus, to maintain continuity used strings instead of bools.

Can track in a separate action item to update all such occurrences to bool

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Actually, I did some testing on the terminal,``` unset GOOGLE_API_USE_CLIENT_CERTIFICATE
 if [[ "${GOOGLE_API_USE_CLIENT_CERTIFICATE}" == "" ]]; then    echo "empty"; fi
empty``` and found value of `GOOGLE_API_USE_CLIENT_CERTIFICATE` is "" after unsetting.

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Use cases of `use_client_cert` in this repo and other repos, which will be using this helper,  have been checking for values `true/false`. Also, the env var is set in string format. Thus, to maintain continuity used strings instead of bools.

Can track in a separate action item to update all such occurrences to bool

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: We have been in favor of when the user is setting value of env var, we need to respect that. If you see, this is when `GOOGLE_API_USE_CLIENT_CERTIFICATE` is set to an unsupported value and not `unset`, in that case, the existing logic has been to create this exception, see for all the APIs on the client side - https://github.com/googleapis/google-cloud-python/blob/c0baa3211766aec17b25450bbbafff61cb9e236c/packages/google-cloud-compute/google/cloud/compute_v1/services/routers/client.py#L317.

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Moved away from file.read now

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Updated to use json to parse, `workload`  is not expected anywhere else IIUC.

google/auth/transport/grpc.py
agrawalradhika-cell: Updated docstring, to add ValueError to the Raises:  section in the docstring

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Updated docstring

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Also, updated the check to make it more robust using json

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Now, I have removed case `when user sets unsupported value`, so that code is backward compatabile with the rest of auth repo. On the client side - https://github.com/googleapis/google-cloud-python/blob/c0baa3211766aec17b25450bbbafff61cb9e236c/packages/google-cloud-compute/google/cloud/compute_v1/services/routers/client.py#L317 will keep the check as is for value from `check_use_client_cert()` helper to make sure the client repo is backward compatible too.

google/auth/transport/_mtls_helper.py
daniel-sanche: an empty string is falsy though, so `if not use_client_cert` covers both cases

agrawalradhika-cell: Fixed

sai-sunder-s (dismissed):

google/auth/transport/_mtls_helper.py
```diff
@@ -17,6 +17,7 @@
 import json
 import logging
 from os import environ, path
+import os
```
nbayati: Consider adding getenv to the import above so we're not importing os in two lines.

```suggestion
from os import environ, path, getenv
```

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,30 @@ def client_cert_callback():
...
     # Then dump the decrypted key bytes
     return crypto.dump_privatekey(crypto.FILETYPE_PEM, pkey)
+
+def check_use_client_cert():
+  """Returns whether the client certificate should to be used for mTLS.
+
+  Returns:
+      str:
+          A boolean indicating if client certificate should be used.
+          The value is "true" or "false" or unset.
```
nbayati: The docstring is a bit ambiguous. It seems to be describing the possible states of the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable ("true", "false", or unset)   under the Returns section, which could be misinterpreted as the function's return value. 

To improve clarity, I suggest rephrasing it to first describe the logic based on the   environment variable's state, and then clearly state what the function returns ("true" or "false")

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,30 @@ def client_cert_callback():
...
+          "true" if the certificate config file contains "workload" section
+          and "false" otherwise.
+  """
+  use_client_cert = os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE")
+  ### Check if the value of GOOGLE_API_USE_CLIENT_CERTIFICATE is unset.
+  if not use_client_cert:
+    cert_path = os.getenv("GOOGLE_API_CERTIFICATE_CONFIG")
+    if cert_path:
+      with open(cert_path, "r") as f:
+        content = json.load(f)
```
nbayati: The open() call can raise a FileNotFoundError, and json.load() can raise a json.JSONDecodeError. These exceptions are not currently handled. It would be more robust to wrap the file reading and parsing in a try...except block. In the case of an exception, you can log a warning and return "false".

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +406,30 @@ def client_cert_callback():
...
+      str:
+          A boolean indicating if client certificate should be used.
+          The value is "true" or "false" or unset.
+          If unset, the function checks if the GOOGLE_API_CERTIFICATE_CONFIG
+          environment variable is set. If it is set, the function returns
+          "true" if the certificate config file contains "workload" section
+          and "false" otherwise.
+  """
+  use_client_cert = os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE")
+  ### Check if the value of GOOGLE_API_USE_CLIENT_CERTIFICATE is unset.
```
nbayati: nit: The two comments in this method that start with ### are unnecessary as the code is self-explanatory. These can be removed to improve code clarity.
If you'd like to keep the comments, I'd recommend changing the prefix to "#" since "###" is not a common practice in this repo.

tests/transport/test__mtls_helper.py
```diff
@@ -638,3 +639,29 @@ def test_crypto_error(self):
             _mtls_helper.decrypt_private_key(
                 ENCRYPTED_EC_PRIVATE_KEY, b"wrong_password"
             )
+
+    def test_check_use_client_cert(self):
+        os.environ["GOOGLE_API_USE_CLIENT_CERTIFICATE"] = "true"
```
nbayati: Directly modifying os.environ can cause state to leak between tests. It's safer to use a fixture that automatically handles setup and teardown. Since this project uses pytest, I'd recommend using the monkeypatch fixture to set and unset the environment variables.

tests/transport/test__mtls_helper.py
```diff
@@ -638,3 +639,29 @@ def test_crypto_error(self):
...
+                }
+            },
+        }
+        config_filename = "mock_certificate_config.json"
+        config_file_content = json.dumps(config_data)
+        # Use mock_open to simulate the file in memory
+        m = mock.mock_open(read_data=config_file_content)
+        with mock.patch("builtins.open", m):
+            os.environ["GOOGLE_API_CERTIFICATE_CONFIG"] = config_filename
+            os.environ["GOOGLE_API_USE_CLIENT_CERTIFICATE"] = ""
```
nbayati: Please apply the monkeypatch solution for environment variable management here as well.

tests/transport/test__mtls_helper.py
```diff
@@ -638,3 +639,29 @@ def test_crypto_error(self):
...
+                "workload": {
+                    "cert_path": "path/to/cert/file",
+                    "key_path": "path/to/key/file",
+                }
+            },
+        }
+        config_filename = "mock_certificate_config.json"
+        config_file_content = json.dumps(config_data)
+        # Use mock_open to simulate the file in memory
+        m = mock.mock_open(read_data=config_file_content)
```
nbayati: nit: The variable m for the mock.mock_open object is not very descriptive. Consider renaming it to something more explicit, like mock_file_handle or mock_open_func, for better readability.

tests/transport/test__mtls_helper.py
nbayati, line 1: The test coverage for check_use_client_cert could be improved. Please consider adding tests for the following cases:
       * GOOGLE_API_USE_CLIENT_CERTIFICATE is explicitly set to "false".
       * The config file is present but does not contain the "workload" key.
       * The config file specified by GOOGLE_API_CERTIFICATE_CONFIG does not exist.
       * The config file is not valid JSON.
       * GOOGLE_API_CERTIFICATE_CONFIG is not set at all.

nbayati (changes requested):

google/auth/transport/_mtls_helper.py
daniel-sanche: +1. If you do this, I'd also suggest just doing `content["cert_configs"]["workload"]` below, and catch KeyError if the field don't exist. That way all the error handling can be in one place

google/auth/transport/_mtls_helper.py
daniel-sanche: I still think this function should return a bool instead of strings. 

If callers truly need to use "true" and "false" strings, they can build the string themselves using an expression like `use_cert = "true" if check_use_client_cert() else "false"`. But the code would be much cleaner if we could avoid that kind of thing everywhere possible

This is a new function, so I'm not really convinced by a backwards compatibility argument, unless I'm missing some context

google/auth/transport/_mtls_helper.py
daniel-sanche: +1, I was wondering what "unset" meant here. If it's talking about the GOOGLE_API_USE_CLIENT_CERTIFICATE env var, that should be moved outside the Returns block

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Done

tests/transport/test__mtls_helper.py
agrawalradhika-cell: Done

tests/transport/test__mtls_helper.py
agrawalradhika-cell: Done

tests/transport/test__mtls_helper.py
agrawalradhika-cell: Done

tests/transport/test__mtls_helper.py
agrawalradhika-cell: Done

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Done

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +405,53 @@ def client_cert_callback():
...
+            except json.JSONDecodeError:
+                _LOGGER.debug("JSON decode error.")
+                return "false"
+            except FileNotFoundError:
+                _LOGGER.debug("Certificate config file not found.")
+                return "false"
+            except OSError:
+                _LOGGER.debug("OS error.")
+                return "false"
+            try:
```
daniel-sanche: Why do we need two try/catch blocks here? Can't they be combined?

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +405,53 @@ def client_cert_callback():
...
+                if content["cert_configs"]["workload"]:
+                    return "true"
```
daniel-sanche, line 442: the "if" seems redundant here. If it's not True, it will raise a key error

I was imagining something like:
```
            try:
                with open(cert_path, "r") as f:
                    content = json.load(f)
                    # verify json has workload key
                    content["cert_configs"]["workload"]
                    return "true"
            except (FileNotFoundError, OSError, KeyError, TypeError) as e:
                   _LOGGER.debug(f"error decoding certificate: {e.message}")
                   return "false"
 
``

**Review by andyrzhao — DISMISSED** [review]
LGTM overall. Please address other reviewer's comments. Thanks!

**Review by andyrzhao — DISMISSED** [review]
LGTM. Thanks!

File: google/auth/transport/_mtls_helper.py [hunk 18]
```diff
@@ -405,3 +405,46 @@ def client_cert_callback():
...
+    """Returns the value of the GOOGLE_API_USE_CLIENT_CERTIFICATE variable,
+    or an inferred 'true' or 'false' value if unset.
+
+    The function checks the value of GOOGLE_API_USE_CLIENT_CERTIFICATE
+    environment variable, and GOOGLE_API_CERTIFICATE_CONFIG environment variable
+    if the former is not set. If GOOGLE_API_USE_CLIENT_CERTIFICATE is set,
+    this helper function returns the value of the former. If it is unset, this
+    helper function checks if GOOGLE_API_CERTIFICATE_CONFIG is set. If it is set,
+    this helper function returns "true" if the certificate config file contains
+    "workload" section and "false" otherwise.
```
daniel-sanche: nit: I think this second paragraph could be re-worded to be a bit clearer. Especially since the first line seems to say something similar

I don't have full context on this change, but maybe something like:

```
    Returns the value of the GOOGLE_API_USE_CLIENT_CERTIFICATE variable,
    or an inferre value if unset.

    This value is meant to be interpreted as a "true" or "false" value representing whether 
    the client certificate should be used, but could be any arbitrary string.

    If GOOGLE_API_USE_CLIENT_CERTIFICATE is unset, the value value will be
    inferred by reading a file pointed at by GOOGLE_API_CERTIFICATE_CONFIG, and 
    verifying it contains a "workload" section. If so, the function will return "true", 
   otherwise "false".
```

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +405,46 @@ def client_cert_callback():
...
+                    return "true"
+            except (
+                FileNotFoundError,
+                OSError,
+                KeyError,
+                TypeError,
+                json.JSONDecodeError,
+            ) as e:
+                _LOGGER.debug("error decoding certificate: %s", e)
+                return "false"
```
daniel-sanche: nit: this return is a bit redundant, since it will just fall through to next line, which does the same thing

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +405,46 @@ def client_cert_callback():
...
+    if the former is not set. If GOOGLE_API_USE_CLIENT_CERTIFICATE is set,
+    this helper function returns the value of the former. If it is unset, this
+    helper function checks if GOOGLE_API_CERTIFICATE_CONFIG is set. If it is set,
+    this helper function returns "true" if the certificate config file contains
+    "workload" section and "false" otherwise.
+
+    Returns:
+        str: A string("true" or "false" or value of the 
+         GOOGLE_API_USE_CLIENT_CERTIFICATE variable set) indicating if client
+         certificate should be used.
```
daniel-sanche: This could also be reworded. maybe something like:

```
str: The value of GOOGLE_API_USE_CLIENT_CERTIFICATE, or an inferred value if unset. 
    This string should contain a "true" or "false" value, but may be an any arbitrary string
    read from the user's environment
```

daniel-sanche (dismissed):
Left a couple suggestions to consider, but LGTM

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Done

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Done

google/auth/transport/_mtls_helper.py
```diff
@@ -405,3 +405,47 @@ def client_cert_callback():
 
     # Then dump the decrypted key bytes
     return crypto.dump_privatekey(crypto.FILETYPE_PEM, pkey)
+
+
+def check_use_client_cert():
```
agrawalradhika-cell, line 410: Please note: That here the function is emitting str instead of boolean because we want to also consider the value set by user for GOOGLE_API_USE_CLIENT_CERTIFICATE

nbayati (approved):

daniel-sanche (approved):

sai-sunder-s (approved):

PR #1849, opened 2025-10-29
feat: Implement token revocation in STS client and add revoke() metho…

kdeniz-git:
…d to ExternalAccountAuthorizedUser credentials

* Add support for OAuth 2.0 token revocation to the STS client, aligning with the specification in RFC7009.

* A new revoke_token method is introduced, which makes a POST request to a revocation endpoint. The underlying request handler has also been updated to correctly process successful but empty HTTP responses, as specified by the standard for revocation.

* Building on the STS client's new capabilities, this change exposes a public revoke() method on the ExternalAccountAuthorizedUser credentials class.

* This method encapsulates the logic for revoking the refresh token by calling the underlying STS client's revoke_token function. It simplifies the process for client applications, like gcloud, to revoke these specific credentials without needing to interact directly with the STS client.

* Unit tests are included to verify successful revocation and to ensure appropriate errors are raised if required fields (like revoke_url) are missing.


google/auth/external_account_authorized_user.py
```diff
@@ -321,6 +321,31 @@ def _build_trust_boundary_lookup_url(self):
             universe_domain=self._universe_domain, pool_id=pool_id
         )
 
+    def revoke(self, request):
```
nbayati, line 324: This looks great. One suggestion for improving the robustness :

After a successful revocation call to the STS client, should we also clear the local state of the credential? For example, by setting self._refresh_token = None and self.token = None.

This would make the credential object's state (valid and can_refresh properties) immediately consistent with the revoked status on the server and would prevent any future attempts to use or refresh this credential object. What do you think?

kdeniz-git: Makes sense. Updated both google/auth/external_account_authorized_user.py and tests/test_external_account_authorized_user.py. Could you please re-review them?

daniel-sanche (dismissed):
LGTM

nbayati (dismissed):

nbayati (approved):

google/oauth2/sts.py
```diff
@@ -174,3 +179,23 @@ def refresh_token(self, request, refresh_token):
             None,
             {"grant_type": "refresh_token", "refresh_token": refresh_token},
         )
+
+    def revoke_token(self, request, token, token_type_hint, revoke_url):
```
daniel-sanche, line 183: nit: It seems like token_type_hint could be optional. It might be good to mention that in the docstrings, and in the signature

daniel-sanche (approved):
LGTM

PR #1856, opened 2025-11-03
feat: MDS connections use mTLS

nolanleastin:
**Feature Gating**
The `GCE_METADATA_MTLS_MODE` environment variable is introduced, which can be set to strict, none, or default.

The `should_use_mds_mtls` function determines whether to use mTLS based on the environment variable and the existence of the certificate files.

**Using MDS mTLS**
A custom `MdsMtlsAdapter` for requests is implemented to handle the SSL context for mTLS.

The `create_session` function mounts an MdsMtlsAdapter into the request.Session

**MdsMtlsAdapter**
Loads MDS mTLS certificates from well-known location (https://docs.cloud.google.com/compute/docs/metadata/overview#https-mds-certificates).

If HTTPS request fails, MdsMtlsAdapter will fallback to HTTP. 

**Integrating with existing code**
compute_engine/_metadata.py:
- The metadata server URL construction is now dynamic, supporting both http and https schemes based on whether mTLS is enabled.
- ping and get functions are updated to use mTLS when it's enabled.

### DIFF
--- a/google/auth/compute_engine/_metadata.py
+++ b/google/auth/compute_engine/_metadata.py
@@ -142,7 +142,7 @@ def detect_gce_residency_linux():
 def _prepare_request_for_mds(request, use_mtls=False):
     """Prepares a request for the metadata server.
 
-    This will check if mTLS should be used and return a new request object if so.
+    This will check if mTLS should be used and mount the mTLS adapter if needed.
 
     Args:
         request (google.auth.transport.Request): A callable used to make
@@ -151,8 +151,8 @@ def _prepare_request_for_mds(request, use_mtls=False):
 
     Returns:
         google.auth.transport.Request: A request object to use.
-            If mTLS is enabled, this will be a new request object with mTLS session configured.
-            Otherwise, it will be the same as the input request.
+            If mTLS is enabled, the request will have the mTLS adapter mounted.
+            Otherwise, the original request will be returned unchanged.
     """
     if not use_mtls:
         return request
--- a/google/auth/compute_engine/_metadata.py
+++ b/google/auth/compute_engine/_metadata.py
@@ -24,6 +24,8 @@ import logging
 import os
 from urllib.parse import urljoin
 
+import requests
+
 from google.auth import _helpers
 from google.auth import environment_vars
 from google.auth import exceptions
@@ -31,9 +33,13 @@ from google.auth import metrics
 from google.auth import transport
 from google.auth._exponential_backoff import ExponentialBackoff
 from google.auth.compute_engine import _mtls
-from google.auth.transport import requests
+
 
 _LOGGER = logging.getLogger(__name__)
+
+_GCE_DEFAULT_MDS_IP = "169.254.169.254"
+_GCE_DEFAULT_HOST = "metadata.google.internal"
+_GCE_DEFAULT_MDS_HOSTS = [_GCE_DEFAULT_HOST, _GCE_DEFAULT_MDS_IP]
 
 # Environment variable GCE_METADATA_HOST is originally named
 # GCE_METADATA_ROOT. For compatibility reasons, here it checks
@@ -42,20 +48,37 @@ _LOGGER = logging.getLogger(__name__)
 _GCE_METADATA_HOST = os.getenv(environment_vars.GCE_METADATA_HOST, None)
 if not _GCE_METADATA_HOST:
     _GCE_METADATA_HOST = os.getenv(
-        environment_vars.GCE_METADATA_ROOT, "metadata.google.internal"
+        environment_vars.GCE_METADATA_ROOT, _GCE_DEFAULT_HOST
     )
 
-_GCE_DEFAULT_MDS_IP = "169.254.169.254"
-_GCE_MDS_HOSTS = ["metadata.google.internal", _GCE_DEFAULT_MDS_IP]
+
+def _validate_gce_mds_configured_environment():
+    """Validates the GCE metadata server environment configuration for mTLS.
+
+    Raises:
+        google.auth.exceptions.MutualTLSChannelError: if the environment
+            configuration is invalid for mTLS.
+    """
+    mode = _mtls._parse_mds_mode()
+    if mode == _mtls.MdsMtlsMode.STRICT:
+        if _GCE_METADATA_HOST != _GCE_DEFAULT_HOST:
+            # mTLS is only supported when connecting to the default metadata host.
+            # Raise an exception if we are in strict mode (which requires mTLS)
+            # but the metadata host has been overridden. (which means mTLS will fail)
+            raise exceptions.MutualTLSChannelError(
+                "Mutual TLS is required, but the metadata host has been overridden. "
+                "mTLS is only supported when connecting to the default metadata host."
+            )
 
 
-def _get_metadata_root(use_mtls):
+def _get_metadata_root(use_mtls: bool):
     """Returns the metadata server root URL."""
+
     scheme = "https" if use_mtls else "http"
     return "{}://{}/computeMetadata/v1/".format(scheme, _GCE_METADATA_HOST)
 
 
-def _get_metadata_ip_root(use_mtls):
+def _get_metadata_ip_root(use_mtls: bool):
     """Returns the metadata server IP root URL."""
     scheme = "https" if use_mtls else "http"
     return "{}://{}".format(
@@ -131,8 +154,14 @@ def _prepare_request_for_mds(request, use_mtls=False):
             If mTLS is enabled, this will be a new request object with mTLS session configured.
             Otherwise, it will be the same as the input request.
     """
-    if use_mtls:
-        request = requests.Request(_mtls.create_session())
+    if not use_mtls:
+        return request
+
+    adapter = _mtls.MdsMtlsAdapter()
+    if not request.session:
+        request.session = requests.Session()
+    for host in _GCE_DEFAULT_MDS_HOSTS:
+        request.session.mount(f"https://{host}/", adapter)
     return request
 
 
@@ -236,6 +265,7 @@ def get(
 
     if root is None:
         root = _get_metadata_root(use_mtls)
+    _validate_gce_mds_configured_environment()
 
     base_url = urljoin(root, path)
     query_params = {} if params is None else params
--- a/google/auth/compute_engine/_metadata.py
+++ b/google/auth/compute_engine/_metadata.py
@@ -263,7 +263,7 @@ def get(
     Raises:
         google.auth.exceptions.TransportError: if an error occurred while
             retrieving metadata.
-        google.auth.exceptions.MutualTLSChannelError: if the environment
+        google.auth.exceptions.MutualTLSChannelError: if using mtls and the environment
             configuration is invalid for mTLS (for example, the metadata host
             has been overridden in strict mTLS mode).
 
--- a/google/auth/compute_engine/_metadata.py
+++ b/google/auth/compute_engine/_metadata.py
@@ -30,6 +30,8 @@ from google.auth import exceptions
 from google.auth import metrics
 from google.auth import transport
 from google.auth._exponential_backoff import ExponentialBackoff
+from google.auth.compute_engine import _mtls
+from google.auth.transport import requests
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -42,13 +44,24 @@ if not _GCE_METADATA_HOST:
     _GCE_METADATA_HOST = os.getenv(
         environment_vars.GCE_METADATA_ROOT, "metadata.google.internal"
     )
-_METADATA_ROOT = "http://{}/computeMetadata/v1/".format(_GCE_METADATA_HOST)
 
-# This is used to ping the metadata server, it avoids the cost of a DNS
-# lookup.
-_METADATA_IP_ROOT = "http://{}".format(
-    os.getenv(environment_vars.GCE_METADATA_IP, "169.254.169.254")
-)
+GCE_MDS_HOSTS = ["metadata.google.internal", "169.254.169.254"]
+
+
+def _get_metadata_root(use_mtls):
+    """Returns the metadata server root URL."""
+    scheme = "https" if use_mtls else "http"
+    return "{}://{}/computeMetadata/v1/".format(scheme, _GCE_METADATA_HOST)
+
+
+def _get_metadata_ip_root(use_mtls):
+    """Returns the metadata server IP root URL."""
+    scheme = "https" if use_mtls else "http"
+    return "{}://{}".format(
+        scheme, os.getenv(environment_vars.GCE_METADATA_IP, "169.254.169.254")
+    )
+
+
 _METADATA_FLAVOR_HEADER = "metadata-flavor"
 _METADATA_FLAVOR_VALUE = "Google"
 _METADATA_HEADERS = {_METADATA_FLAVOR_HEADER: _METADATA_FLAVOR_VALUE}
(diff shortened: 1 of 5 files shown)


google/auth/compute_engine/_mtls.py
```diff
@@ -0,0 +1,109 @@
...
+    )  # path to CA certificate
+    client_combined_cert_path: str = os.path.join(
+        Path.home(), "mtls_mds_certificates", "client_creds.key"
+    )  # path to file containing client certificate and key
+
+
+class MdsMtlsMode(enum.Enum):
+    """MDS mTLS mode."""
+
+    STRICT = "strict"
```
sai-sunder-s, line 83: can you add a comment describing what each mode means

google/auth/compute_engine/_mtls.py
```diff
@@ -0,0 +1,109 @@
...
+        mds_mtls_config.client_combined_cert_path
+    )
+
+
+class MdsMtlsAdapter(HTTPAdapter):
+    """An HTTP adapter that uses mTLS for the metadata server."""
+
+    def __init__(self, mds_mtls_config: MdsMtlsConfig, *args, **kwargs):
+        self.ssl_context = ssl.create_default_context()
+        self.ssl_context.load_verify_locations(cafile=mds_mtls_config.ca_cert_path)
```
sai-sunder-s, line 123: what are `ca_cert_path` and `client_combined_cert_path`

nolanleastin: ca_cert_path points to the `root.crt` -- this is the self-signed CA certificate, used to verify the server identity (i.e. MDS)

client_combined_cert_path points to a file called `client_creds.key` -- this has  the client certificate and private key.

google/auth/compute_engine/_mtls.py
nolanleastin: Done!

sai-sunder-s (dismissed):

sai-sunder-s (dismissed):

google/auth/compute_engine/_metadata.py
```diff
@@ -42,13 +44,24 @@
...
+    """Returns the metadata server root URL."""
+    scheme = "https" if use_mtls else "http"
+    return "{}://{}/computeMetadata/v1/".format(scheme, _GCE_METADATA_HOST)
+
+
+def _get_metadata_ip_root(use_mtls):
+    """Returns the metadata server IP root URL."""
+    scheme = "https" if use_mtls else "http"
+    return "{}://{}".format(
+        scheme, os.getenv(environment_vars.GCE_METADATA_IP, "169.254.169.254")
```
daniel-sanche: It seems stange that this IP is hardcoded in multiple places. Can this be GCE_MDS_HOSTS[-1]? Or can we make a DEFAULT_GCE_METADATA_IP constant?

google/auth/compute_engine/_metadata.py
```diff
@@ -102,6 +115,24 @@ def detect_gce_residency_linux():
...
+
+    This will check if mTLS should be used and return a new request object if so.
+
+    Args:
+        request (google.auth.transport.Request): A callable used to make
+            HTTP requests.
+
+    Returns:
+        google.auth.transport.Request: Request
+            object to use.
```
daniel-sanche: nit: the line break seems unnecessary here

google/auth/compute_engine/_metadata.py
```diff
@@ -102,6 +115,24 @@ def detect_gce_residency_linux():
...
 
 
+def _prepare_request_for_mds(request, use_mtls=False):
+    """Prepares a request for the metadata server.
+
+    This will check if mTLS should be used and return a new request object if so.
+
+    Args:
+        request (google.auth.transport.Request): A callable used to make
+            HTTP requests.
```
daniel-sanche, line 153: use_mtls isn't mentioned here

google/auth/compute_engine/_metadata.py
```diff
@@ -115,6 +146,8 @@ def ping(request, timeout=_METADATA_DEFAULT_TIMEOUT, retry_count=3):
     Returns:
         bool: True if the metadata server is reachable, False otherwise.
     """
+    use_mtls = _mtls.should_use_mds_mtls()
+    request = _prepare_request_for_mds(request, use_mtls=use_mtls)
```
daniel-sanche: I'm not familiar with how requests work in this library, but I find it surprising the request passed in is completely replaced with a new one when mtls is enabled. Wouldn't it be cleaner to create a different request at the caller side, instead of accepting one and then overriding it here?

If you do keep this behaviour, do the docstrings need to be updated to make it clear that the input request might not be used?
> request (google.auth.transport.Request): A callable used to make HTTP requests.

 Or am I thinking about all this wrong?

daniel-sanche: Also, is _prepare_request_for_mds really needed? It seems like this could just be a one-liner: 

`request = requests.Request(_mtls.create_session()) if use_mtls else request`

google/auth/compute_engine/_mtls.py
```diff
@@ -0,0 +1,135 @@
...
+# https://cloud.google.com/compute/docs/metadata/overview#https-mds-certificates
+
+
+def _get_mds_root_crt_path():
+    if os.name == "nt":
+        return os.path.join(
+            "C:\\", "ProgramData", "Google", "ComputeEngine", "mds-mtls-root.crt"
+        )
+    else:
+        return os.path.join("/", "run", "google-mds-mtls", "root.crt")
```
daniel-sanche: These paths are a bit unweildy, especially since the directory is referenced multiple times in this file. I'd suggest doing something like:

```
WIN_BASE_PATH_COMPONENTS = ("C:\\", "ProgramData", "Google", "ComputeEngine")
BASE_PATH_COMPONENTS = ("/", "run", "google-mds-mtls")

def _get_mds_root_crt_path():
    if os.name == "nt":
        return os.path.join(*WIN_BASE_PATH_COMPONENTS, "mds-mtls-root.crt")
    else:
        return os.path.join(*BASE_PATH_COMPONENTS, "root.crt")
```

Or, preferably, consider using Path objects instead of strings. It could simplify some of the weirdness of paths on different platforms

```
from pathlib import Path

WIN_BASE_PATH = PATH("C:/ProgramData/Google/ComputeEngine")
BASE_PATH = PATH("/run/google-mds-mtls")

def _get_mds_root_crt_path():
    if os.name == "nt":
        return WIN_BASE_PATH / "mds-mtls-root.crt"
    else:
        return BASE_PATH / "root.crt"
```

google/auth/compute_engine/_mtls.py
```diff
@@ -0,0 +1,135 @@
...
+
+from google.auth import environment_vars, exceptions
+
+# MDS mTLS certificate paths based on OS.
+# Documentation to well known locations can be found at:
+# https://cloud.google.com/compute/docs/metadata/overview#https-mds-certificates
+
+
+def _get_mds_root_crt_path():
+    if os.name == "nt":
```
daniel-sanche: nit: The "nt" string is used in a few places, and could be confusing if you're not familiar

Maybe consider making this a constant. Something like `WINDOWS_OS_NAME="nt"`

daniel-sanche (commented):
Nothing major, but left a few suggestions to consider

google/auth/compute_engine/_metadata.py
nolanleastin: I actually did initially implement it the way you suggested here (which is: the caller passes in a different request). 

The problem @sai-sunder-s pointed out is that people using this library will create their own request object and then pass it in.

google/auth/compute_engine/_metadata.py
nolanleastin: I can add a _DEFAULT_GCE_METADATA_IP constant

google/auth/compute_engine/_metadata.py
daniel-sanche: Isn't this a private file? Do users call these functions directly?

daniel-sanche (changes requested):
Looking at this a bit closer, it looks like this change would [contradict the public API](https://github.com/googleapis/google-auth-library-python/blob/89a883824b437e9be0c97309ed36f35e5c071f11/google/oauth2/id_token.py#L255) as it currently stands. Custom request objects constructed and passed in by users would be bypassed after the next release, which could break existing code. I think we need to consider that more before merging 

Or let me know if I'm misunterstanding

google/auth/compute_engine/_metadata.py
nolanleastin: I have modified this. I no longer create a new request object!

nolanleastin: I modified how this works.

google/auth/compute_engine/_mtls.py
nolanleastin: Done

google/auth/compute_engine/_mtls.py
nolanleastin: Done

google/auth/compute_engine/_metadata.py
nolanleastin: done

google/auth/compute_engine/_metadata.py
```diff
@@ -190,6 +258,15 @@ def get(
...
             retrieving metadata.
     """
+    use_mtls = _mtls.should_use_mds_mtls()
+    # Prepare the request object for mTLS if needed.
+    # This will create a new request object with the mTLS session.
+    request = _prepare_request_for_mds(request, use_mtls=use_mtls)
+
+    if root is None:
+        root = _get_metadata_root(use_mtls)
+    _validate_gce_mds_configured_environment()
```
daniel-sanche, line 282: nit: maybe add a comment explaining what this does

google/auth/compute_engine/_metadata.py
```diff
@@ -190,6 +258,15 @@ def get(
         google.auth.exceptions.TransportError: if an error occurred while
             retrieving metadata.
     """
```
daniel-sanche, line 270: Raises doesn't mention MutualTLSChannelError

google/auth/compute_engine/_metadata.py
```diff
@@ -153,7 +221,7 @@ def ping(request, timeout=_METADATA_DEFAULT_TIMEOUT, retry_count=3):
 def get(
     request,
     path,
-    root=_METADATA_ROOT,
+    root=None,
```
daniel-sanche, line 229: the docstring for this says `root (str):`. It looks like it should be `root (Optional[str]):`. And the docstring should mention what happens if its left as None

google/auth/compute_engine/_mtls.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+@dataclass
+class MdsMtlsConfig:
+    ca_cert_path: str = field(
+        default_factory=_get_mds_root_crt_path
+    )  # path to CA certificate
+    client_combined_cert_path: str = field(
+        default_factory=_get_mds_client_combined_cert_path
+    )  # path to file containing client certificate and key
```
daniel-sanche, line 65: these are marked as str, but it looks like they are paths now?

Are you running `nox -s mypy` checks? I'm not sure if they're configured to run in CI

google/auth/compute_engine/_metadata.py
nolanleastin: Done

google/auth/compute_engine/_metadata.py
nolanleastin: Done

google/auth/compute_engine/_metadata.py
nolanleastin: Done

google/auth/compute_engine/_mtls.py
nolanleastin: Yes I have been running mypy checks! Not sure why this isn't flagged though.

nolanleastin: Either way, updated

google/auth/compute_engine/_metadata.py
```diff
@@ -24,31 +24,72 @@
...
+    Raises:
+        google.auth.exceptions.MutualTLSChannelError: if the environment
+            configuration is invalid for mTLS.
+    """
+    mode = _mtls._parse_mds_mode()
+    if mode == _mtls.MdsMtlsMode.STRICT:
+        if _GCE_METADATA_HOST != _GCE_DEFAULT_HOST:
+            # mTLS is only supported when connecting to the default metadata host.
+            # Raise an exception if we are in strict mode (which requires mTLS)
+            # but the metadata host has been overridden. (which means mTLS will fail)
```
sai-sunder-s: This validation is good. For extra robustness, you could consider checking if `_GCE_METADATA_HOST` is in `_GCE_DEFAULT_MDS_HOSTS` instead of just comparing it to `_GCE_DEFAULT_HOST`. This would also cover the case where the IP address is used as the host. However, the current implementation is likely sufficient as it's improbable for the IP to be set as the host via the environment variable.

sai-sunder-s (commented):
This is a great addition to the library. The implementation of mTLS for the GCE metadata server is well-structured and thoroughly tested.

Here's a summary of my review:
- **Good separation of concerns:** The mTLS logic is nicely encapsulated in the new `_mtls.py` module.
- **Clear configuration:** The use of the `GCE_METADATA_MTLS_MODE` environment variable provides a clear and flexible way to configure the mTLS behavior.
- **Robustness:** The fallback mechanism in `default` mode is a good feature to ensure that the library can still function even if mTLS is not properly configured.
- **Thorough testing:** The addition of new test files and the extension of existing tests provide good coverage for the new functionality.

I have one minor suggestion for improvement in `google/auth/compute_engine/_metadata.py`, which I've added as a line comment.

Overall, this looks ready to merge.

google/auth/compute_engine/_metadata.py
```diff
@@ -102,6 +143,32 @@ def detect_gce_residency_linux():
...
+    """
+    if not use_mtls:
+        return request
+
+    adapter = _mtls.MdsMtlsAdapter()
+    if not request.session:
+        request.session = requests.Session()
+    for host in _GCE_DEFAULT_MDS_HOSTS:
+        request.session.mount(f"https://{host}/", adapter)
+    return request
```
sai-sunder-s: you are returning the same request object as the input. All changes are inside the object.

Do we really have to return?

google/auth/compute_engine/_metadata.py
```diff
@@ -189,7 +262,24 @@ def get(
     Raises:
         google.auth.exceptions.TransportError: if an error occurred while
             retrieving metadata.
+        google.auth.exceptions.MutualTLSChannelError: if the environment
+            configuration is invalid for mTLS (for example, the metadata host
+            has been overridden in strict mTLS mode).
+
     """
+    use_mtls = _mtls.should_use_mds_mtls()
```
sai-sunder-s, line 271: can we determine whether to use mtls or not at cred initialization time and put it in an attribute instead of determining each time?

Do we expect the value to change after cred initilization?

sai-sunder-s (dismissed):
Overall LGTM. Left couple of comments.

google/auth/compute_engine/_metadata.py
nolanleastin: This is a good point. I will make this edit.

google/auth/compute_engine/_metadata.py
nolanleastin: Ok, this is a good point. Changed.

google/auth/compute_engine/_metadata.py
```diff
@@ -189,7 +263,24 @@ def get(
     Raises:
         google.auth.exceptions.TransportError: if an error occurred while
             retrieving metadata.
+        google.auth.exceptions.MutualTLSChannelError: if the environment
+            configuration is invalid for mTLS (for example, the metadata host
+            has been overridden in strict mTLS mode).
```
daniel-sanche, line 268: nit: this is only if mtls is requested and the configuration is invalid, right?

daniel-sanche (dismissed):
LGTM

google/auth/compute_engine/_metadata.py
nolanleastin: yes that's right. updated that.

sai-sunder-s (approved):

daniel-sanche (approved):

PR #1859, opened 2025-11-04
feat: Add public wrapper for _mtls_helper.check_use_client_cert which enables mTLS if GOOGLE_API_USE_CLIENT_CERTIFICATE is not set, when the MWID/X.509 cert sources detected

agrawalradhika-cell:
Add public wrapper for check_use_client_cert which enables mTLS if GOOGLE_API_USE_CLIENT_CERTIFICATE is not set, when the MWID/X.509 cert sources detected. Also, fix check_use_client_cert to return boolean value.

Change #1848 added the check_use_client_cert method that helps know if client cert should be used for mTLS connection. However, that was in a private class, thus, created a public wrapper of the same function so that it can be used by python Client Libraries. Also, updated check_use_client_cert to return a boolean value instead of existing string value for better readability and future scope.


google/auth/transport/_mtls_helper.py
```diff
@@ -408,28 +408,27 @@ def client_cert_callback():
...
-
-    If GOOGLE_API_USE_CLIENT_CERTIFICATE is unset, the value value will be
-    inferred by reading a file pointed at by GOOGLE_API_CERTIFICATE_CONFIG, and
-    verifying it contains a "workload" section. If so, the function will return
-    "true", otherwise "false".
+    This value is meant to be interpreted as a boolean representing whether
+    the client certificate should be used. If GOOGLE_API_USE_CLIENT_CERTIFICATE
+    is unset, the value will be inferred by reading a file pointed at by
+    GOOGLE_API_CERTIFICATE_CONFIG, and verifying it contains a "workload"
+    section. If so, the function will return True, otherwise False.
```
daniel-sanche: > This value is meant to be interpreted as a boolean representing whether the client certificate should be used.

This sentence seems unnecessary. It made more sense to explain string format, but I don't think it's needed anymore as a bool. No interpretation needed.

daniel-sanche: The docstring should mention how GOOGLE_API_USE_CLIENT_CERTIFICATE is interpreted into a bool, and how it handles unexpected non-true/false values

Maybe something like:

```
If GOOGLE_API_USE_CLIENT_CERTIFICATE is set to true or false, a corresponding 
bool value will be returned

If GOOGLE_API_USE_CLIENT_CERTIFICATE is unset or set to an invalid string, the value will 
be determined by reading a file pointed  at by GOOGLE_API_CERTIFICATE_CONFIG, and
verifying it contains a "workload" section. If so, the function will return True, otherwise 
False.
```

google/auth/transport/_mtls_helper.py
```diff
@@ -408,28 +408,27 @@ def client_cert_callback():
...
     """
     use_client_cert = getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE")
     # Check if the value of GOOGLE_API_USE_CLIENT_CERTIFICATE is set.
     if use_client_cert:
-        return use_client_cert.lower()
+        if use_client_cert.lower() == "true":
+            return True
+        # Check if GOOGLE_API_USE_CLIENT_CERTIFICATE is set to false explicitly.
+        # Invalid values for GOOGLE_API_USE_CLIENT_CERTIFICATE are not handled here.
+        # That will be handled by the code calling this function.
```
daniel-sanche: > Invalid values for GOOGLE_API_USE_CLIENT_CERTIFICATE are not handled here.
> That will be handled by the code calling this function.

I know that this is the plan for the code you're implementing, but this function should be able to stand alone, it shouldn't make declarations about what external code will do.

It looks like you're choosing to ignore values that aren't true or false and falling back to inferring the value from GOOGLE_API_CERTIFICATE_CONFIG instead, so that's what the comments and docstrings should say (even though you plan on having your callers do something different)

google/auth/transport/grpc.py
```diff
@@ -292,7 +292,7 @@ class SslCredentials:
 
     def __init__(self):
         use_client_cert = _mtls_helper.check_use_client_cert()
-        if use_client_cert != "true":
+        if use_client_cert:
```
daniel-sanche: shouldn't this be `if not use_client_cert:`? (I see a few more of these too)

google/auth/transport/mtls.py
```diff
@@ -110,3 +110,22 @@ def callback():
...
+    GOOGLE_API_CERTIFICATE_CONFIG, and verifying it contains a "workload"
+    section. If so, the function will return True, otherwise False. Also, note
+    that if GOOGLE_API_USE_CLIENT_CERTIFICATE is set but is not 'true' or 'false'
+    (case-insensitive), this check is inconclusive, that case should be handled
+    by the caller.
+
+    Returns:
+       bool: indicating whether the client certificate should be used for mTLS.
+    """
+    return _mtls_helper.check_use_client_cert()
```
daniel-sanche, line 129: Do you think you need both this public function and the old helper? Maybe you should just delete the helper and move it over here

tests/transport/test__mtls_helper.py
```diff
@@ -643,7 +643,7 @@ def test_crypto_error(self):
     def test_check_use_client_cert(self, monkeypatch):
         monkeypatch.setenv("GOOGLE_API_USE_CLIENT_CERTIFICATE", "true")
         use_client_cert = _mtls_helper.check_use_client_cert()
-        assert use_client_cert == "true"
+        assert use_client_cert
```
daniel-sanche: nit: `assert use_client_cert is True` and `assert use_client_cert is False` are preferred for checking against bool values

google/auth/transport/mtls.py
andyrzhao: Synced with Radhika offline on this - I don't think it hurts to keep the core logic check in the old _mtls_helper for the sake of code organization and encapsulation. should_use_client_cert() can potentially incorporate other decision factors as well in the future.

nbayati (commented):
Overall LGTM, I just have a minor suggestion: Add a test case to verify the behavior when GOOGLE_API_USE_CLIENT_CERTIFICATE is set to a string other than "true" or "false". The current implementation correctly evaluates such cases to False, and adding an explicit test for this would make the test suite more comprehensive.

google/auth/transport/_mtls_helper.py
```diff
@@ -408,28 +408,22 @@ def client_cert_callback():
...
-    This value is meant to be interpreted as a "true" or "false" value
-    representing whether the client certificate should be used, but could be any
-    arbitrary string.
-
-    If GOOGLE_API_USE_CLIENT_CERTIFICATE is unset, the value value will be
-    inferred by reading a file pointed at by GOOGLE_API_CERTIFICATE_CONFIG, and
-    verifying it contains a "workload" section. If so, the function will return
-    "true", otherwise "false".
+    If GOOGLE_API_USE_CLIENT_CERTIFICATE is set to true or false, a corresponding
+    bool value will be returned.
```
daniel-sanche: It looks like the behaviour changed since I suggested this wording. Maybe add this:

```
If the value is set to an unexpected string, it will default to False
```

daniel-sanche (dismissed):
LGTM, but one more small comment

google/auth/transport/_mtls_helper.py
agrawalradhika-cell: Done

nbayati (approved):

yaoliugg (approved):

andyrzhao (approved):
lgtm, thanks for working on this!

daniel-sanche (approved):

PR #1869, opened 2025-11-10
test: create and delete temp folder in scripts

Linchin:


scripts/encrypt-secrets.sh
```diff
@@ -29,4 +29,6 @@ gcloud kms encrypt \
   --plaintext-file=system_tests/secrets.tar \
   --ciphertext-file=system_tests/secrets.tar.enc
 
-rm system_tests/secrets.tar
\ No newline at end of file
+rm system_tests/secrets.tar
+
+rm system_tests/data
```
sai-sunder-s, line 34: this means user has to create the authorized_user.json file every time?

Linchin: When I ran the script, it would create all three json files in the empty folder.

sai-sunder-s (approved):

PR #1872, opened 2025-11-11
feat: add ecdsa p-384 support

shengjiang3:
GDC (Google Distributed Cloud) needs to support ECDSA-P384 keys for compliance. This change creates an EsSigner and EsVerifier class that is capable of supporting both ECDSA-P256 and ECDSA-P384 keys for backwards compatibility. The EsSigner and EsVerifier classes are plumbed through to the GDC service accounts and are used to both sign and verify JWTs.

This implementation was successfully tested against a GDC instance using both ECDSA-P256 and ECDSA-P384 keys.


google/auth/crypt/es256.py
```diff
@@ -15,161 +15,13 @@
...
-from cryptography.hazmat.primitives.asymmetric.utils import encode_dss_signature
-import cryptography.x509
+from google.auth.crypt.es import EsSigner
+from google.auth.crypt.es import EsVerifier
 
-from google.auth import _helpers
-from google.auth.crypt import base
 
+class ES256Verifier(EsVerifier):
+    pass
```
daniel-sanche, line 30: class docstrings would be useful here. 

It seems like `ES256Verifier` is essentially an alias for `EsVerifier`?

google/auth/crypt/__init__.py
```diff
@@ -65,9 +67,11 @@
...
 RSAVerifier = rsa.RSAVerifier
 
-if es256 is not None:  # pragma: NO COVER
-    ES256Signer = es256.ES256Signer
-    ES256Verifier = es256.ES256Verifier
+if es is not None:  # pragma: NO COVER
+    ES256Signer = es.EsSigner
+    ES256Verifier = es.EsVerifier
+    EsSigner = es.EsSigner
+    EsVerifier = es.EsVerifier
```
daniel-sanche: Couldn't this be combined with the other `if es is not None:` block above?

google/auth/crypt/__init__.py
```diff
@@ -65,9 +67,11 @@
 RSASigner = rsa.RSASigner
 RSAVerifier = rsa.RSAVerifier
 
-if es256 is not None:  # pragma: NO COVER
-    ES256Signer = es256.ES256Signer
-    ES256Verifier = es256.ES256Verifier
+if es is not None:  # pragma: NO COVER
+    ES256Signer = es.EsSigner
+    ES256Verifier = es.EsVerifier
```
daniel-sanche: We have `ES256Verifier`/`ES256Signer` as separate classes in the the es256 module, but the import file is shadowing them with a `EsSigner`/`EsVerifier` aliases

I understand they are both set up as aliases, so this is essentially equivalent for now. But I could see it getting very confusing in the future, if we try to make changes to the classes

google/auth/crypt/__init__.py
```diff
@@ -41,13 +41,15 @@
 from google.auth.crypt import rsa
 
 try:
-    from google.auth.crypt import es256
+    from google.auth.crypt import es
 except ImportError:  # pragma: NO COVER
```
daniel-sanche, line 48: - In what case would the import fail (I assume importing cryptography?)? That should probably be documented as a comment
- this is marked as NO COVER. But it seems like we should have some coverage for this scenerio. Thoughts?

google/auth/crypt/es.py
```diff
@@ -0,0 +1,208 @@
...
+_BACKEND = backends.default_backend()
+_PADDING = padding.PKCS1v15()
+
+
+class EsVerifier(base.Verifier):
+    """Verifies ECDSA cryptographic signatures using public keys.
+
+    Args:
+        public_key (
+                cryptography.hazmat.primitives.asymmetric.ec.ECDSAPublicKey):
```
daniel-sanche: Is this type accurate? 

I wanted to check if `SECP384R1` is a subclass of `ECDSAPublicKey`, but I'm having trouble finding `ECDSAPublicKey`

type annotations could be valuable in this class. It looks like there's a [mypy check set up](https://github.com/googleapis/google-auth-library-python/blob/d5638986ca03ee95bfffa9ad821124ed7e903e63/noxfile.py#L96) for this repo

google/auth/crypt/es.py
```diff
@@ -0,0 +1,208 @@
...
+
+    @_helpers.copy_docstring(base.Verifier)
+    def verify(self, message, signature):
+        # First convert (r||s) raw signature to ASN1 encoded signature.
+        sig_bytes = _helpers.to_bytes(signature)
+        if len(sig_bytes) != self._r_s_size * 2:
+            return False
+        r = (
+            int.from_bytes(sig_bytes[: self._r_s_size], byteorder="big")
+            if _helpers.is_python_3()
```
daniel-sanche: Can we remove these python2 code paths?

google/auth/crypt/es.py
```diff
@@ -0,0 +1,208 @@
+# Copyright 2017 Google Inc.
```
daniel-sanche, line 1: Note to self (and other reviewers): most of this file is copied over from `es256.py`, so I only focused on diff for this review

google/auth/crypt/es.py
```diff
@@ -0,0 +1,208 @@
...
+            self._sha_algo = hashes.SHA384()
+            self._alg = "ES384"
+        else:
+            self._r_s_size = 32
+            self._sha_algo = hashes.SHA256()
+            self._alg = "ES256"
+
+    @property  # type: ignore
+    def alg(self):
+        """Optional[str]: The algorithm used to sign the messages."""
```
daniel-sanche: - why alg? Can this be algorithm?
- the docstring says Optional, but it looks like it's always set
- is returning a hard-coded string here really helpful? Maybe I'm misunderstanding the use case, but consider:
  - return `self._sha_algo` or `self._sha_algo.name`
  - return `ec.ECDSA(self._sha_algo)`
  - declare an enum to hold the supported/expected values

google/auth/crypt/es.py
```diff
@@ -0,0 +1,208 @@
...
+        # for P-384 curve. For P-256 curve, we use SHA256 hash algo,
+        # and for P-384 curve we use SHA384 algo.
+        if isinstance(private_key.curve, ec.SECP384R1):
+            self._r_s_size = 48
+            self._sha_algo = hashes.SHA384()
+            self._alg = "ES384"
+        else:
+            self._r_s_size = 32
+            self._sha_algo = hashes.SHA256()
+            self._alg = "ES256"
```
daniel-sanche: These attributes are tightly coupled, and hard-coded in multiple places. Maybe it would be good to create a dataclass or named tuple to hold these? Something like:

```
@dataclass
class _ESAttrs:
    rs_size: int
    sha_algo: hashes.HashAlgorithm
    name: str

    @classmethod
    def from_key(cls, key):
        if isinstance(private_key.curve, ec.SECP384R1):
            return cls(48, hashes.SHA384(), "ES384")
        else:
            # default to ES256
            return cls(32, hashes.SHA256(), "ES256")

class EsSigner:
    def __init__(self, private_key, key_id=None):
        self._es_attrs = _ESAttrs.from_key(private_key)
```

google/auth/crypt/__init__.py
shengjiang3: I kept these as aliases in case there are users that import `ES256Signer`/`ES256Verifier` to not break them. What is the deprecation policy for this library? 

I agree that it is confusing though. Thoughts?

google/auth/crypt/__init__.py
shengjiang3: Added a comment.

I tried to add a unit test to test this, but found that it is not trivial to simulate an import error. The logic here is straight forward enough that it might be ok not to have code coverage for this part. What do you think?

google/auth/crypt/es256.py
shengjiang3: Done.

google/auth/crypt/__init__.py
shengjiang3: Done.

google/auth/crypt/es.py
shengjiang3: Done.

google/auth/crypt/es.py
shengjiang3: Updated the type and added type annotations.

google/auth/crypt/es.py
shengjiang3: Thanks! Added.

google/auth/crypt/es.py
shengjiang3: This is used to set the algorithm on the JWT. The value is different from the value held in self._sha_algo. Updated the name to algorithm and the doc string as well.

google/auth/crypt/__init__.py
daniel-sanche: Yes, we need to keep them around to avoid breaking existing code. Can we replace these with imports from the `es256.py` file, so we are defining the alias in one place?

google/auth/crypt/__init__.py
daniel-sanche: IIRC, you can unimport a package using something like `del cryptography`. That should let you define a test method that doesn't have cryptography active. Did you try doing something like that?

google/auth/crypt/es.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    def __init__(
+        self, private_key: ec.EllipticCurvePrivateKey, key_id: Optional[str] = None
+    ) -> None:
+        self._key = private_key
+        self._key_id = key_id
+        self._attributes = ESAttributes.from_private_key(private_key)
+
+    @property  # type: ignore
```
daniel-sanche: why do we need the type: ignore here?

google/auth/crypt/es.py
```diff
@@ -0,0 +1,226 @@
...
+    algorithm: str
+
+    @classmethod
+    def from_public_key(cls, key: ec.EllipticCurvePublicKey):
+        return cls.from_curve(key.curve)
+
+
+    @classmethod
+    def from_private_key(cls, key: ec.EllipticCurvePrivateKey):
+        return cls.from_curve(key.curve)
```
daniel-sanche, line 58: nit: these constructors seem a bit excessive. Maybe we should just have `from_curve`, instead of the extra helpers?

Or at least combine `from_public_key` and `from_private_key` to a single `from_key` with a union type, since the implementation is the same?

If you think this is useful, I'm fine with it though

google/auth/crypt/es.py
```diff
@@ -0,0 +1,226 @@
...
+from google.auth.crypt import base
+
+
+_CERTIFICATE_MARKER = b"-----BEGIN CERTIFICATE-----"
+_BACKEND = backends.default_backend()
+_PADDING = padding.PKCS1v15()
+
+
+@dataclass
+class ESAttributes:
```
daniel-sanche: Mark this with a leading _ to mark it as private, so we can make changes easily in the future if needed (assuming it's not meant to be exposed publicly)

google/auth/crypt/__init__.py
shengjiang3: Yeah, I tried using monkey patch to remove the library and then to reload imports but it wasn't working. I think spending additional time to figure out how to simulate an import error just to check that the ES* types are not defined has low ROI.

google/auth/crypt/es.py
shengjiang3: Good call. Done.

google/auth/crypt/es.py
shengjiang3: Removed.

google/auth/crypt/es.py
shengjiang3: Done.

daniel-sanche (dismissed):
LGTM

Linchin (dismissed):

daniel-sanche (approved):

Linchin (approved):

PR #1873, opened 2025-11-19
tests: make the TLS tests skip when pyopenssl isn't available

kraj:
In some distributions e.g. yocto
pyOpenSSL is a packageconfig and disabled by default, and it means pyopenssl.py module is not installed as a result.

This test fails plainly when it can not find urllib3.contrib.pyopenssl instead skip the test if it does not find the module.


gkevinzheng (dismissed):

parthea (dismissed):

chalmerlowe (approved):
LGTM

ohmayr (approved):

PR #1878, opened 2025-12-01
feat: Bring TB up to date with design changes

nbayati:
Make the fetching async and non blocking.
Provide a method for manual override.
Implement proactive refresh every 6 hours.
Implement automatic recovery if api request fails due to stale regional boundary. Remove no-op signal and checks.
Refactor to Regional Access Boundary name.


parthea:
Switching to draft since there are merge conflicts.

herbertjrjohnd-dev (approved):

PR #1880, opened 2025-12-02
fix(auth): Add temporary patch to workload cert logic to accomodate Cloud Run mis-configuration

andyrzhao:
This patch adds a fallback logic to look for Cloud Run cert/keys in the well-known location if the cert config contains the exact incorrect cert/key paths AND the incorrect cert/key paths point to non-existent files. 

Note: This patch will be reverted sometime in Jan 2026, after Cloud Run environment is updated with the correct cert configs. The revert will be tracked by googleapis/google-cloud-python#15137


nbayati (dismissed):
LGTM

google/auth/transport/_mtls_helper.py
```diff
@@ -183,6 +197,24 @@ def _get_workload_cert_and_key_paths(config_path):
         )
     key_path = workload["key_path"]
 
+    # == BEGIN Temporary Cloud Run PATCH ==
```
daniel-sanche, line 200: Please add a `TODO:` string somewhere giving some context, and letting us know how and when the patch can be removed. Maybe add a link to this PR or a bug too (is there a tracking bug that will be resolved when this can be removed?

andyrzhao: I created a github issue to track this at https://github.com/googleapis/google-auth-library-python/issues/1881 and assigned it to myself. Included this in the comments.

daniel-sanche (approved):

PR #1900, opened 2025-12-19
chore: rename _refresh_token to _perform_refresh_token

daniel-sanche:
[A recent change](https://github.com/googleapis/google-auth-library-python/commit/36ecb1d65883477d27faf9c2281fc289659b9903#diff-4762e5cf4088fa77af8bf63829f842548a4cad6e76be593786876e314807cd8f) added an internal _refresh_token method to some credentials files.

Parts of the codebase were already using `_refresh_token` as an internal variable, holding a token value

This causes some inconsistencies in the codebase, where it's not always clear if `_refresh_token` is referring to an action or an object

It was also causing complications in internal code that depended on _refresh_token objects

This PR renames the methods to `_perform_refresh_token`, to avoid the duplicate naming

### DIFF
--- a/google/auth/external_account_authorized_user.py
+++ b/google/auth/external_account_authorized_user.py
@@ -124,7 +124,7 @@ class Credentials(
         self.token = token
         self.expiry = expiry
         self._audience = audience
-        self._refresh_token_val = refresh_token
+        self._refresh_token = refresh_token
         self._token_url = token_url
         self._token_info_url = token_info_url
         self._client_id = client_id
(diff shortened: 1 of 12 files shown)


google/auth/external_account_authorized_user.py
```diff
@@ -333,15 +333,15 @@ def revoke(self, request):
...
                 "revoke the refresh token. You must specify revoke_url and "
                 "refresh_token."
             )
 
         self._sts_client.revoke_token(
-            request, self._refresh_token_val, "refresh_token", self._revoke_url
+            request, self._refresh_token, "refresh_token", self._revoke_url
         )
         self.token = None
         self._refresh_token = None
```
daniel-sanche, line 347: This appears to be a bug in the previous code. I think this was supposed to be `self._refresh_token_val`.

This PR should fix any issue here though

chalmerlowe (approved):
LGTM

PR #1918, opened 2026-01-06
fix: Python 3.8 support

daniel-sanche:
There were recently some type annotation changes that weren't compatible with python 3.8. This PR fixes those, and adds back 3.8 tests using github actions


cashpilotthrive-hue (approved):
Confirming all

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,58 @@
+on:
```
parthea, line 15: Add copyright header

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,58 @@
...
+  pull_request:
+    branches:
+      - main
+name: unittest
```
parthea, line 23: Set read only permissions

```suggestion

permissions:
  contents: read
  
name: unittest
```

parthea (dismissed):
LGTM assuming that 3.7 will be removed prior to releasing

.github/workflows/unittest.yml
daniel-sanche: good catch

parthea (approved):

PR #1919, opened 2026-01-06
chore: remove Python 3.7 support

daniel-sanche:
Removing support for EOL Python 3.7. 

A warning was present in the README that support would be removed after January 2024

Also added a warning that Python 3.8 and 3.9 are also EOL, and will be removed in future updates

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,15 +35,16 @@ Note that the extras pyopenssl and enterprise_cert should not be used together b
 
 Supported Python Versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^
-Python >= 3.7
+Python >= 3.8
 
 **NOTE**:
-Python 3.7 was marked as `unsupported`_ by the python community in June 2023.
-We recommend that all developers upgrade to Python 3.8 and newer as soon as
-they can. Support for Python 3.7 will be removed from this library after
-January 1 2024. Previous releases that support Python 3.7 will continue to be available
-for download, but releases after January 1 2024 will only target Python 3.8 and
-newer.
+Python 3.8 and Python 3.9 were marked as `unsupported`_ by the python community in 
+October 2024 and October 2025 respectively.
+We recommend that all developers upgrade to Python 3.10 and newer as soon as
+they can. Support for end-of-life Python runtimes will be removed from this
+library in future updates.
+Previous releases that support end-of-life Python versions will continue to be available
+for download, but future releases will only target supported versions.
 
 .. _unsupported: https://devguide.python.org/versions/#unsupported-versions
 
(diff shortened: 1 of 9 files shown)


google/oauth2/__init__.py
```diff
@@ -27,10 +27,14 @@ class Python37DeprecationWarning(DeprecationWarning):  # pragma: NO COVER
...
-        "After January 1, 2024, new releases of this library will drop support "
-        "for Python 3.7."
-    )
-    warnings.warn(message, Python37DeprecationWarning)
+# Raise warnings for deprecated versions
+eol_message = """
+    You are using a Python version {} past its end of life. Google will update
+    google-auth with critical bug fixes on a best-effort basis, but not
+    with any other fixes or features. Please upgrade your Python version,
+    and then update google-auth.
```
vchudnov-g, line 35: Do google/auth and google/oauth2 get released as part of the same PyPI package? In that case, could we re-use this check rather than duplicating it? If not, should this message say "google-oauth2"?

README.rst
```diff
@@ -35,15 +35,16 @@ Note that the extras pyopenssl and enterprise_cert should not be used together b
...
 
 **NOTE**:
-Python 3.7 was marked as `unsupported`_ by the python community in June 2023.
-We recommend that all developers upgrade to Python 3.8 and newer as soon as
-they can. Support for Python 3.7 will be removed from this library after
-January 1 2024. Previous releases that support Python 3.7 will continue to be available
-for download, but releases after January 1 2024 will only target Python 3.8 and
-newer.
+Python 3.8 and Python 3.9 were marked as `unsupported`_ by the python community in 
+October 2024 and October 2025 respectively.
```
vchudnov-g: nit
```suggestion
October 2024 and October 2025, respectively.
```

vchudnov-g (dismissed):
One substantive comment and one nit.

google/oauth2/__init__.py
daniel-sanche: I believe it's all released as a single package

I worry a bit about complications importing from one package to the other, or adding shared logic to the root `/google` directory. I'm not entirely sure if that would be a problem, but I can imagine a few ways that could get messy. I thought it would be safer to use the existing logic, which was duplicated.

I can try to look into it a bit more, but we do need something working today, since this is blocking a release

daniel-sanche: Yeah, it looks like:
- adding an __init__.py to /google isn't an option, since it's a shared namespace, so that could interfere with other packages
- if we kept the shared logic in auth and imported it in oauth2, there would be circular dependencies
- we may be able to get away with importing oauth2 from oauth. But I worry that will still cause import headaches, and import bloat

These lines are meant to be temporary, and will be replaced by your more robust version check system soon. I [opened a bug](https://github.com/googleapis/google-auth-library-python/issues/1921) to track that work. Let me know what you think

vchudnov-g: sgtm. (I'd make the bug a p2 chore, but we can discuss that separately.)

vchudnov-g (approved):

PR #1925, opened 2026-01-09
chore: add warning for rsa library

daniel-sanche:
The `rsa` library is archived, and scheduled to be removed. `google-auth` already supports an alternate implementation, using `cryptography`. This PR adds a warning to users still relying on the old library, and adds `rsa` as an optional dependency to allow users to continue to opt-in to rsa during deprecation persion

After release, a follow-up version will remove rsa as a required dependency, leaving it only for opt-in users

### DIFF
--- a/google/auth/crypt/_python_rsa.py
+++ b/google/auth/crypt/_python_rsa.py
@@ -22,6 +22,7 @@ certificates. There is no support for p12 files.
 from __future__ import absolute_import
 
 import io
+import warnings
 
 from pyasn1.codec.der import decoder  # type: ignore
 from pyasn1_modules import pem  # type: ignore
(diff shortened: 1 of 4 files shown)


google/auth/crypt/_python_rsa.py
```diff
@@ -39,6 +40,15 @@
 _PKCS8_MARKER = ("-----BEGIN PRIVATE KEY-----", "-----END PRIVATE KEY-----")
 _PKCS8_SPEC = PrivateKeyInfo()
 
+warnings.warn(
+    (
+        "The 'rsa' library is deprecated and will be removed in a future release. "
+        "Please migrate to 'cryptography'."
+    ),
+    category=FutureWarning,
```
vchudnov-g: Should this be a `DeprecationWarning`? I think that might be more appropriate if it doesn't introduce complications with existing tests, etc.

(not a blocker)

vchudnov-g (dismissed):

google/auth/crypt/_python_rsa.py
daniel-sanche: I was debating that too. I don't think the user vs developer divide described [in the docs](https://docs.python.org/3/library/warnings.html#warning-categories) is super clear for me in this context. 

Thinking again, maybe DeprecationWarning is a bit more accurate here. I made the change

vchudnov-g (approved):

PR #1928, opened 2026-01-10
[DRAFT] fix: remove dependency on rsa

daniel-sanche:
The rsa library has been deprecated and archived. This PR will remove it as a required dependency. Users are advised to use the optional cryptography dependency (`google-auth[cryptography]`) instead.

In the short term, `rsa` will still be supported as an optional dependency. In the long term, it will eventually be removed

As a part of this change, I introduced new RSASigner and RSAVerifier wrapper classes, that can use either cryptography or rsa implementations. Previously, the library would only use one, depending on if cryptography was installed. This simplifies the import structure, and puts `rsa` and `cryptography` on equal footing as optional dependencies

Note: Do not merge yet. We will do one [release giving a warning](https://github.com/googleapis/google-auth-library-python/pull/1925), and then this change will come in a future release. We plan on doing a pre-release on pypi of this branch, so users can opt-in to the `rsa`-free build early


daniel-sanche:
After some discussion, we will make `cryptography` a required dependency, instead of leaving them both as optional. Closing this in favor of https://github.com/googleapis/google-auth-library-python/compare/add_cryptography_dependency

Issue #1738, opened 2025-04-08
Add bounds to the dependencies in `enterprise_cert_extra_require`


ohmayr:
Determine if `cryptography_base_require` can be re-used which allows `"cryptography>=38.0.3"` instead of allowing all version of `cryptography` in `enterprise_cert_extra_require` in setup.py.

Do the same for the `pyopenssl` dependency within `enterprise_cert_extra_require` instead of allowing all version.

PR #1929, opened 2026-01-12
feat: add cryptography as required dependency

### LINKED ISSUE #912: Push `cryptography` more strongly
At the moment, many-many users of `google-auth` end up using the slower (and with security holes for some) `rsa`, rather than `cryptography`, not because they *can't* install `cryptography`, but because they don't know they are supposed to.  We should at a minimum have an `extra` for it in `setup.py`, and mention it more frequently and forcefully in the docs (including `README.md`).

Unlike Debian / Red Hat's systems, Python's dependency mechanism isn't featureful enough to let us require an "abstract" dependency with preferences for one over another, which would be the only "correct" fix for us.

I don't know the business case for continuing to rely on `rsa`:  perhaps someone can follow up?

daniel-sanche:
The `rsa` library has been deprecated and archived. This PR adds `cryptography` as a the new preferred backend for RSA operations

In the short term, both `rsa` and `cryptography` will be listed as dependencies. Soon, `rsa` will be removed, but still supported as an optional dependency. Eventually, it will be completely removed from the codebase.

As a part of this change, I introduced new RSASigner and RSAVerifier wrapper classes, that can use either cryptography or rsa implementations. Previously, the library would only import one or the other, depending on if cryptography was installed. This simplifies the import structure, and puts rsa and cryptography on equal footing

Fixes https://github.com/googleapis/google-auth-library-python/issues/912
Towards https://github.com/googleapis/google-cloud-python/issues/15600
Towards https://github.com/googleapis/google-auth-library-python/issues/941


setup.py
```diff
@@ -21,27 +21,28 @@
 
 DEPENDENCIES = (
     "pyasn1-modules>=0.2.1",
-    # rsa==4.5 is the last version to support 2.7
-    # https://github.com/sybrenstuvel/python-rsa/issues/152#issuecomment-643470233
+    "cryptography >= 38.0.3",
+    # TODO: remove rsa from dependencies in next release (replaced with cryptography)
```
parthea: Update TODO comment with a link to a tracking bug to follow up on this

google/auth/crypt/rsa.py
```diff
@@ -12,19 +12,128 @@
...
+
+
+class RSAVerifier(base.Verifier):
+    """Verifies RSA cryptographic signatures using public keys.
+
+    Requires installation of `cryptography` optional dependency.
+
+    .. deprecated::
+        The `rsa` library has been archived. Please migrate to
+        `cryptography` for public keys.
```
parthea:
```suggestion
    """Verifies RSA cryptographic signatures using public keys.

```

parthea: We're using the cryptography module by default so we don't need these notes

tests/crypt/test_rsa.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+DATA_DIR = os.path.join(os.path.dirname(__file__), "..", "data")
+
+with open(os.path.join(DATA_DIR, "privatekey.pem"), "rb") as fh:
+    PRIVATE_KEY_BYTES = fh.read()
+    CRYPTOGRAPHY_PRIVATE_KEY = serialization.load_pem_private_key(
+        PRIVATE_KEY_BYTES, password=None, backend=backends.default_backend()
+    )
+    RSA_PRIVATE_KEY = rsa_lib.PrivateKey.load_pkcs1(PRIVATE_KEY_BYTES)
```
parthea: nit: We could switch these to test fixtures

https://docs.pytest.org/en/stable/explanation/fixtures.html

setup.py
```diff
@@ -21,27 +21,28 @@
...
     "rsa>=3.1.4,<5",
 )
 
-cryptography_base_require = [
+# Note: cryptography was made into a required dependency. Extra is kept for backwards compatibility
+cryptography_extra_require = [
     "cryptography >= 38.0.3",
 ]
```
parthea: We could rename this back to `cryptography_base_require` and use it directly in `DEPENDENCIES` to avoid code duplication for ` "cryptography >= 38.0.3",`

google/auth/crypt/rsa.py
daniel-sanche: good catch

setup.py
daniel-sanche: Done

tests/crypt/test_rsa.py
daniel-sanche: Good idea, added

setup.py
```diff
@@ -18,30 +18,30 @@
...
 
 aiohttp_extra_require = ["aiohttp >= 3.6.2, < 4.0.0", *requests_extra_require]
 
-pyjwt_extra_require = ["pyjwt>=2.0", *cryptography_base_require]
+pyjwt_extra_require = ["pyjwt>=2.0"]
 
 reauth_extra_require = ["pyu2f>=0.1.5"]
 
 # TODO(https://github.com/googleapis/google-auth-library-python/issues/1738): Add bounds for cryptography and pyopenssl dependencies.
 enterprise_cert_extra_require = ["cryptography", "pyopenssl"]
```
parthea: Do we need `cryptography` here? Can https://github.com/googleapis/google-auth-library-python/issues/1738 be closed?

setup.py
parthea, line 72: non-blocking: For testing, we're using `pyopenssl < 24.3.0`, we should revisit this in the future so that we're able to test the latest version of `pyopenssl`. Same with `aiohttp` below

tests/crypt/test_rsa.py
```diff
@@ -0,0 +1,175 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from unittest import mock
+
+from cryptography.hazmat import backends
+from cryptography.hazmat.primitives import serialization
+import pytest
+import rsa as rsa_lib
```
parthea, line 21: We have `import rsa` here . Did you intend to add more tests that depend on the deprecated library?

daniel-sanche: Yes, I think we should still support using `rsa` keys when the user explicitly installs it, at least for the short term

This PR changed the interface a bit, where there's now one class that handles both key types. So I added a few extra tests to make sure both key types are supported

setup.py
daniel-sanche: I think we should be able to remove this. I'll make the change

parthea (approved):

PR #1930, opened 2026-01-13
fix: remove deprecated rsa dependency

daniel-sanche:
This PR removes `rsa` as a required dependency, now that it has been replaced with `cryptography` (with warnings issued)

This PR is currently blocked until at least `v2.49.0`. Cryptography was [just recently added](https://github.com/googleapis/google-auth-library-python/pull/1929) as the default, so we want to give time for that release to propagate before removing rsa

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,27 +4,27 @@
 # For syntax help see:
 # https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
-# The @googleapis/googleapis-auth and @googleapis/python-core-client-libraries is the default owner for changes in this repo
-*                                                           @googleapis/googleapis-auth @googleapis/python-core-client-libraries
-google/auth/_default.py                                     @googleapis/googleapis-auth @googleapis/aion-sdk @googleapis/python-core-client-libraries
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -73,6 +70,9 @@
     # TODO(https://github.com/googleapis/google-auth-library-python/issues/1722): `test_aiohttp_requests` depend on
     # aiohttp < 3.10.0 which is a bug. Investigate and remove the pinned aiohttp version.
     "aiohttp < 3.10.0",
+    # rsa library was removed as a dependency, but we still have some code paths that support it
+    # TODO: remove dependency when google.auth.crypt._python_rsa is removed
+    "rsa>=3.1.4,<5",
 ]
 
 extras = {
```
parthea, line 79: nit: add an extra for `rsa` since we still use it in the code

https://github.com/googleapis/google-auth-library-python/blob/6a982be06e513a3182069cd910466e743fa7db6a/google/auth/crypt/_python_rsa.py#L31

Otherwise, the changes look good.

daniel-sanche: I was a bit on the fence about it, since rsa is a dead library we don't want to "endorse". But I guess it makes sense to add so if people do choose to use rsa, they can find a version that's compatible

Added

PR #1933, opened 2026-01-13
chore: remove deprecated deprecation warning

akx:
I looked through [GitHub search](https://github.com/search?q=Python37DeprecationWarning&type=code), and it looks like no external code is referencing this deprecation warning, so I suppose it's good to go the way of the dodo post #1919.

### DIFF
--- a/google/auth/__init__.py
+++ b/google/auth/__init__.py
@@ -32,15 +32,6 @@ __version__ = google_auth_version.__version__
 __all__ = ["default", "load_credentials_from_file", "load_credentials_from_dict"]
 
 
-class Python37DeprecationWarning(DeprecationWarning):  # pragma: NO COVER
-    """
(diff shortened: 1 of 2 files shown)


daniel-sanche:
>  The only thing I could think of is someone doing with pytest.warns(Python37DeprecationWarning) (or similar), but that seems very unlikely.

This is the kind of thing I had in mind (or filterwarnings). There could be customers doing this kind of thing in private repositories, and removing this class would would cause their code to crash in the next release. It's GCP policy that once something is released as part of the public API, it is only ever removed in a major version upgrade, to avoid that. Even if the use-case seems uncommon

> It is also not de facto part of the public API that is declared in __all__

I just double-checked, and it is part of the \_\_init\_\_.py, and I can import it with `import google.auth.Python37DeprecationWarning`. So unfortunately, we're stuck with this for now

Thanks for bringing this up though

PR #1489, opened 2024-03-05
fix: Remove content-type header from IDMS GET requests

peter-volkov:
Detailed description is here
https://issuetracker.google.com/issues/328089077

PR #1934, opened 2026-01-13
fix: removes `content-header` from AWS IMDS get request

chalmerlowe:
When performing a IMDS request, the code incorrectly adds a content-type header to the request:

`content-type: application/json` to AWS metadata (IMDS) GET requests.`

Some services at AWS (such as AWS SageMaker Jupyter notebook) have a stricter than normal metadata server (IMDS, both v1 and v2) when it comes to handling incoming http requests.

This PR removes the default content-header and replaces it with `None`.

NOTE: initializing headers to `None` (instead of an empty `dict`) when no session token is present matches the existing behavior in `_get_metadata_role_name` and allows the transport adapter to handle default headers cleanly.

This PR updates existing unit tests (`tests/test_aws.py`) to match the new behavior.

NOTE: closing PR #1489 due to inactivity as we make the push to migrate this library to the `google-cloud-python` monorepo

For more information about the genesis of this, see the following issue: https://issuetracker.google.com/issues/328089077

### DIFF
--- a/google/auth/aws.py
+++ b/google/auth/aws.py
@@ -530,7 +530,7 @@ class _DefaultAwsSecurityCredentialsSupplier(AwsSecurityCredentialsSupplier):
             google.auth.exceptions.RefreshError: If an error occurs while
                 retrieving the AWS security credentials.
         """
-        headers = {"Content-Type": "application/json"}
+        headers = None
         if imdsv2_session_token is not None:
             headers["X-aws-ec2-metadata-token"] = imdsv2_session_token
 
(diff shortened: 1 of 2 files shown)


google/auth/aws.py
```diff
@@ -530,9 +530,9 @@ def _get_metadata_security_credentials(
             google.auth.exceptions.RefreshError: If an error occurs while
                 retrieving the AWS security credentials.
         """
-        headers = {"Content-Type": "application/json"}
+        headers = None
         if imdsv2_session_token is not None:
-            headers["X-aws-ec2-metadata-token"] = imdsv2_session_token
+            headers = {"X-aws-ec2-metadata-token": imdsv2_session_token}
```
akx, line 534:
```suggestion
        if imdsv2_session_token is not None:
            headers = {"X-aws-ec2-metadata-token": imdsv2_session_token}
        else:
            headers = None
```
would probably make Gemini happier.

parthea (approved):

PR #1942, opened 2026-01-16
fix: resolve circular imports

daniel-sanche:
We have a couple circular imports in the library:
- _mtls_helper and _agent_identity_utils both require each other
- credentials and _default form a cycle

This makes it difficult to import the library into google3

This PR resolved the import issues


chalmerlowe (approved):
LGTM

PR #1955, opened 2026-02-07
feat: implement updated design for regional access boundary

nbayati:
Make the fetching async and non blocking.
Provide a method for manual override.
Implement proactive refresh every 6 hours.
Implement automatic recovery if api request fails due to stale regional boundary. Remove no-op signal and checks.
Refactor to Regional Access Boundary name.

### DIFF
--- a/google/auth/_constants.py
+++ b/google/auth/_constants.py
@@ -1,5 +1,5 @@
 """Shared constants."""
 
-_SERVICE_ACCOUNT_TRUST_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/projects/-/serviceAccounts/{service_account_email}/allowedLocations"
-_WORKFORCE_POOL_TRUST_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/locations/global/workforcePools/{pool_id}/allowedLocations"
-_WORKLOAD_IDENTITY_POOL_TRUST_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/projects/{project_number}/locations/global/workloadIdentityPools/{pool_id}/allowedLocations"
+_SERVICE_ACCOUNT_REGIONAL_ACCESS_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/projects/-/serviceAccounts/{service_account_email}/allowedLocations"
+_WORKFORCE_POOL_REGIONAL_ACCESS_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/locations/global/workforcePools/{pool_id}/allowedLocations"
+_WORKLOAD_IDENTITY_POOL_REGIONAL_ACCESS_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/projects/{project_number}/locations/global/workloadIdentityPools/{pool_id}/allowedLocations"
--- a/google/auth/_regional_access_boundary_utils.py
+++ b/google/auth/_regional_access_boundary_utils.py
@@ -1,0 +1,87 @@
+"""Utilities for Regional Access Boundary management."""
+
+import threading
+import datetime
+
+from google.auth import _helpers
+from google.auth import exceptions
+from google.auth._default import _LOGGER
+
+
+# The default lifetime for a cached Regional Access Boundary.
+DEFAULT_REGIONAL_ACCESS_BOUNDARY_TTL = datetime.timedelta(hours=6)
+
+# The initial cooldown period for a failed Regional Access Boundary lookup.
+DEFAULT_REGIONAL_ACCESS_BOUNDARY_COOLDOWN = datetime.timedelta(minutes=15)
+
+
+class _RegionalAccessBoundaryRefreshThread(threading.Thread):
+    """Thread for background refreshing of the Regional Access Boundary."""
+
+    def __init__(self, credentials, request):
+        super(_RegionalAccessBoundaryRefreshThread, self).__init__()
+        self._credentials = credentials
+        self._request = request
+
+    def run(self):
+        """
+        Performs the Regional Access Boundary lookup. This method is run in a separate thread.
+
+        It includes a short-term retry loop for transient server errors. If the
+        lookup fails completely, it sets a longer-term cooldown period on the
+        credential to avoid overwhelming the lookup service.
+        """
+        regional_access_boundary_info = self._credentials._lookup_regional_access_boundary_with_retry(
+            self._request
+        )
+
+        if regional_access_boundary_info:
+            # On success, update the boundary and its expiry, and clear any cooldown.
+            self._credentials._regional_access_boundary = regional_access_boundary_info
+            self._credentials._regional_access_boundary_expiry = (
+                _helpers.utcnow() + DEFAULT_REGIONAL_ACCESS_BOUNDARY_TTL
+            )
+            self._credentials._regional_access_boundary_cooldown_expiry = None
+            if _helpers.is_logging_enabled(_LOGGER):
+                _LOGGER.debug(
+                    "Asynchronous Regional Access Boundary lookup successful."
+                )
+        else:
+            # On complete failure, set a cooldown period. The existing
+            # _regional_access_boundary and _regional_access_boundary_expiry
+            # will be kept as they are considered safe to use until explicitly
+            # invalidated by a "stale Regional Access Boundary" API error.
+            if _helpers.is_logging_enabled(_LOGGER):
+                _LOGGER.warning(
+                    "Asynchronous Regional Access Boundary lookup failed. Entering cooldown."
+                )
+
+            self._credentials._regional_access_boundary_cooldown_expiry = (
+                _helpers.utcnow() + DEFAULT_REGIONAL_ACCESS_BOUNDARY_COOLDOWN
+            )
+
+
+class _RegionalAccessBoundaryRefreshManager(object):
+    """Manages a thread for background refreshing of the Regional Access Boundary."""
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._worker = None
+
+    def start_refresh(self, credentials, request):
+        """
+        Starts a background thread to refresh the Regional Access Boundary if one is not already running.
+
+        Args:
+            credentials (CredentialsWithRegionalAccessBoundary): The credentials
+                to refresh.
+            request (google.auth.transport.Request): The object used to make
+                HTTP requests.
+        """
+        with self._lock:
+            if self._worker and self._worker.is_alive():
+                # A refresh is already in progress.
+                return
+
+            self._worker = _RegionalAccessBoundaryRefreshThread(credentials, request)
+            self._worker.start()
--- a/google/auth/_regional_access_boundary_utils.py
+++ b/google/auth/_regional_access_boundary_utils.py
@@ -3,8 +3,11 @@
 import datetime
 import threading
 
+import logging
+
 from google.auth import _helpers
-from google.auth._default import _LOGGER
+
+_LOGGER = logging.getLogger(__name__)
 
 
 # The default lifetime for a cached Regional Access Boundary.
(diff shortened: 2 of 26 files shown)


google/auth/compute_engine/credentials.py
```diff
@@ -66,7 +66,6 @@ def __init__(
         scopes=None,
         default_scopes=None,
         universe_domain=None,
-        trust_boundary=None,
```
daniel-sanche, line 69: We can't remove an argument for a public class

We could add a comment to mark it as deprecated though, if it's no longer needed. But what's the context here?

google/auth/compute_engine/credentials.py
```diff
@@ -211,57 +210,39 @@ def get_cred_info(self):
...
-            scopes=self._scopes,
-            default_scopes=self._default_scopes,
-            service_account_email=self._service_account_email,
-            quota_project_id=self._quota_project_id,
-            trust_boundary=self._trust_boundary,
-            universe_domain=universe_domain,
-        )
-
-    @_helpers.copy_docstring(credentials.CredentialsWithTrustBoundary)
-    def with_trust_boundary(self, trust_boundary):
```
daniel-sanche, line 255: this is also part of the public API surface

google/auth/credentials.py
```diff
@@ -290,8 +286,21 @@ def with_universe_domain(self, universe_domain):
         )
 
 
-class CredentialsWithTrustBoundary(Credentials):
```
daniel-sanche, line 293: We have to keep the old class around too

google/auth/_constants.py
```diff
@@ -1,5 +1,5 @@
 """Shared constants."""
 
-_SERVICE_ACCOUNT_TRUST_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/projects/-/serviceAccounts/{service_account_email}/allowedLocations"
-_WORKFORCE_POOL_TRUST_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/locations/global/workforcePools/{pool_id}/allowedLocations"
-_WORKLOAD_IDENTITY_POOL_TRUST_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.{universe_domain}/v1/projects/{project_number}/locations/global/workloadIdentityPools/{pool_id}/allowedLocations"
+_SERVICE_ACCOUNT_REGIONAL_ACCESS_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{service_account_email}/allowedLocations"
+_WORKFORCE_POOL_REGIONAL_ACCESS_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.googleapis.com/v1/locations/global/workforcePools/{pool_id}/allowedLocations"
+_WORKLOAD_IDENTITY_POOL_REGIONAL_ACCESS_BOUNDARY_LOOKUP_ENDPOINT = "https://iamcredentials.googleapis.com/v1/projects/{project_number}/locations/global/workloadIdentityPools/{pool_id}/allowedLocations"
```
daniel-sanche, line 5: do we not need to worry about universe domains now?

google/auth/_helpers.py
```diff
@@ -338,14 +333,8 @@ def get_bool_from_env(variable_name, default=False):
...
         return True
-    elif value in ("false", "0"):
-        return False
     else:
-        raise exceptions.InvalidValue(
-            'Environment variable "{}" must be one of "true", "false", "1", or "0".'.format(
-                variable_name
-            )
-        )
+        return False
```
daniel-sanche, line 337: nit: this could be `return value.lower() in ("true", "1")`

google/auth/_regional_access_boundary_utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class _RegionalAccessBoundaryRefreshManager(object):
+    """Manages a thread for background refreshing of the Regional Access Boundary."""
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._worker = None
+
+    def start_refresh(self, credentials, request):
```
daniel-sanche, line 107: I'm trying to get my head around the ownership here. It looks like each credentials object has a _RegionalAccessBoundaryRefreshManager to manage refresh? But we also pass in a credentials object at refresh time? Would start_refresh ever be called with a different credential object?

Does it make sense to return early from the manager when a worker is active, if the worker is working on a different credentials object? That seems like it could lead to misunderstandings

I just want to make sure I understand this, since there are locks in play.

google/auth/_regional_access_boundary_utils.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        Starts a background thread to refresh the Regional Access Boundary if one is not already running.
+
+        Args:
+            credentials (CredentialsWithRegionalAccessBoundary): The credentials
+                to refresh.
+            request (google.auth.transport.Request): The object used to make
+                HTTP requests.
+        """
+        with self._lock:
```
daniel-sanche, line 117: is this lock needed? Is the manager ever entered from a background threads?

google/auth/credentials.py
```diff
@@ -19,17 +19,17 @@
...
 from google.auth import _helpers, environment_vars
+from google.auth import _regional_access_boundary_utils
 from google.auth import exceptions
 from google.auth import metrics
 from google.auth._credentials_base import _BaseCredentials
 from google.auth._refresh_worker import RefreshThreadManager
 
 DEFAULT_UNIVERSE_DOMAIN = "googleapis.com"
-NO_OP_TRUST_BOUNDARY_LOCATIONS: List[str] = []
-NO_OP_TRUST_BOUNDARY_ENCODED_LOCATIONS = "0x0"
```
daniel-sanche, line 32: These might be considered part of the public API too

daniel-sanche (changes requested):
I'll try to do another pass on this soon, but my main feedback so far is the we need to keep the public API consistent, even if we're not using parts of it anymore. We can add comments to the docstrings mentioning deprecated arguments and their replacements

google/auth/compute_engine/credentials.py
```diff
@@ -157,24 +154,26 @@ def _build_trust_boundary_lookup_url(self):
             try:
                 info = _metadata.get_service_account_info(request, "default")
```
lsirac, line 155: There's no request in this scope, how does this work?

tests/compute_engine/test_credentials.py
```diff
@@ -350,301 +349,68 @@ def test_refresh_trust_boundary_lookup_skipped_if_env_var_missing(
...
             "email": "resolved-email@example.com"
         }
-        mock_get_universe_domain.return_value = "googleapis.com"
-
-        url = creds._build_trust_boundary_lookup_url()
+        creds = self.credentials
+        creds._universe_domain_cached = True
+        url = creds._build_regional_access_boundary_lookup_url()
 
         mock_get_service_account_info.assert_called_once_with(mock.ANY, "default")
```
lsirac, line 368: Probably shouldn't use mock.ANY, as I think this masked a bug. This should pass a mock request object to the call above, and then validate that it was passed here.

google/auth/credentials.py
```diff
@@ -307,141 +316,198 @@ def _perform_refresh_token(self, request):
...
+        return True
 
-    def _get_trust_boundary_header(self):
-        if self._trust_boundary is not None:
-            if self._has_no_op_trust_boundary():
-                # STS expects an empty string if the trust boundary value is no-op.
-                return {"x-allowed-locations": ""}
-            else:
-                return {"x-allowed-locations": self._trust_boundary["encodedLocations"]}
+    def _get_regional_access_boundary_header(self):
```
lsirac, line 465: With the current implementation, if the background refresh fails repeatedly and the 6-hour TTL passes, we will continue to blindly attach the expired header?

google/auth/_helpers.py
```diff
@@ -314,8 +314,7 @@ def get_bool_from_env(variable_name, default=False):
     The environment variable is interpreted as a boolean with the following
     (case-insensitive) rules:
     - "true", "1" are considered true.
-    - "false", "0" are considered false.
-    Any other values will raise an exception.
+    - Any other value (or unset) is considered false.
```
lsirac, line 317: Documentation says this is case insensitive, implementation doesn't match.

google/oauth2/_service_account_async.py
```diff
@@ -75,6 +75,13 @@ async def refresh(self, request):
         self.token = access_token
         self.expiry = expiry
 
+    @_helpers.copy_docstring(credentials_async.Credentials)
+    async def before_request(self, request, method, url, headers):
+        # Explicit override to bypass synchronous CredentialsWithRegionalAccessBoundary.
+        await credentials_async.Credentials.before_request(
+            self, request, method, url, headers
+        )
+
```
lsirac, line 84: Can you explain this bit?

google/auth/credentials.py
```diff
@@ -307,141 +316,198 @@ def _perform_refresh_token(self, request):
...
+        if (
+            self._regional_access_boundary_cooldown_expiry
+            and _helpers.utcnow() < self._regional_access_boundary_cooldown_expiry
+        ):
+            return
+
+        # If all checks pass, start the background refresh.
+        self._regional_access_boundary_refresh_manager.start_refresh(self, request)
+
+    def _is_regional_access_boundary_lookup_required(self):
```
lsirac, line 444: os.environ is polled on every request. Should we cache it?

google/auth/_regional_access_boundary_utils.py
lsirac: +1

google/auth/credentials.py
```diff
@@ -307,141 +316,198 @@ def _perform_refresh_token(self, request):
...
+    def _copy_regional_access_boundary_state(self, target):
+        """Copies the regional access boundary state to another instance."""
+        target._regional_access_boundary = self._regional_access_boundary
+        target._regional_access_boundary_expiry = self._regional_access_boundary_expiry
+        target._regional_access_boundary_cooldown_expiry = (
+            self._regional_access_boundary_cooldown_expiry
+        )
+        target._current_rab_cooldown_duration = self._current_rab_cooldown_duration
+        # Create a new lock for the target instance to ensure independent thread-safety.
+        target._stale_boundary_lock = threading.Lock()
```
lsirac, line 391: We should not be creating a new threading.Lock() here, nor should we let the copied credential use a new _RegionalAccessBoundaryRefreshManager.

When credentials are copied via with_quota_project() or with_scopes(), they share the exact same expiry time. If a user has 5 scoped copies of a credential and the soft expiry threshold is reached, all 5 copies will try to refresh concurrently. Because they have different locks and managers, they will spawn 5 identical background threads and to call the lookup endpoint simultaneously.

To prevent these redundant network requests, cloned credentials should share the same _stale_boundary_lock and _regional_access_boundary_refresh_manager by reference so the manager can successfully deduplicate the refresh threads.

lsirac:
Great work with the PR! Adding some of Gemini's thoughts:

# Test Coverage Gaps

**A. Masking the missing `request` bug with `mock.ANY`**
In `tests/compute_engine/test_credentials.py`, we are using `mock.ANY` to validate the request object being passed:
`mock_get_service_account_info.assert_called_once_with(mock.ANY, "default")`

This is dangerous because it is actively masking a `NameError` bug. The variable `request` isn't actually being passed down to `_build_regional_access_boundary_lookup_url()`, but `mock.ANY` blindly accepts the garbage/leaked variable from the test environment. We should instantiate a `mock_req = mock.Mock()`, pass it explicitly, and assert against that exact mock object rather than using `mock.ANY`. This will force us to fix the missing `request` arguments in the URL builder methods.

**B. Missing coverage for the background thread's "Happy Path"**
It looks like many of the old synchronous `refresh()` tests for the boundary lookups were deleted because the architecture shifted to a background thread (e.g., the massive `-271` line deletions in Compute Engine tests). However, we currently have no tests verifying that the new background thread successfully executes, updates the credential state, and properly handles TTL/cooldowns. We need to add tests that explicitly validate `_RegionalAccessBoundaryRefreshThread.run()` to ensure we aren't silently failing in the background.

**C. Missing coverage for Reactive Refresh (Stale RAB Error)**
The design document mentions that if a developer provides a manual override that goes stale, the library should catch the `"stale regional access boundary" error` (406 Not Acceptable), clear the cache, retry, and trigger a background refresh. I noticed we don't have any transport-layer tests covering this reactive refresh logic. We need both the implementation and the tests to verify the cache clears and the retry succeeds.

PR #1956, opened 2026-02-09
feat: Add helper methods for asynchronous x.509 certificate discovery

agrawalradhika-cell:
This PR introduces the google.auth.aio.transport.mtls module, providing asynchronous helper methods for mTLS certificate discovery [go/caa:x509-async-support](http://goto.google.com/caa:x509-async-support). These helpers are designed to be for x.509 certs discovery and non-blocking, ensuring that disk I/O operations are async during mTLS handshake process. Plus, added unit tests respectively for the helper functions.

Please note: Only x.509 creds are in scope of this project currently. Context aware or ECP credentials are not in scope of this project currently.

Next Steps: Will create a followup PR will that will utilize these helpers to implement `configure_mtls_channel` within the `AsyncAuthorizedSession` class.

### DIFF
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 """
-Helper functions for mTLS in async.
+Helper functions for mTLS in async for discovery of certs.
 """
 
+import asyncio
 import logging
 from os import getenv, path
 
@@ -42,6 +43,20 @@ def _check_config_path(config_path):
     return config_path
 
 
+async def _run_in_executor(func, *args):
+    """Run a blocking function in an executor to avoid blocking the event loop.
+
+    This implements the non-blocking execution strategy for disk I/O operations.
+    """
+    try:
+        # For python versions 3.9 and newer versions
+        return await asyncio.to_thread(func, *args)
+    except AttributeError:
+        # Fallback for older Python versions
+        loop = asyncio.get_running_loop()
+        return await loop.run_in_executor(None, func, *args)
+
+
 def has_default_client_cert_source():
     """Check if default client SSL credentials exists on the device.
 
@@ -56,7 +71,7 @@ def has_default_client_cert_source():
     return False
 
 
-def get_client_ssl_credentials(
+async def get_client_ssl_credentials(
     generate_encrypted_key=False,
     certificate_config_path=None,
 ):
@@ -84,16 +99,18 @@ def get_client_ssl_credentials(
     """
 
     # Attempt to retrieve X.509 Workload cert and key.
-    cert, key = google.auth.transport._mtls_helper._get_workload_cert_and_key(
-        certificate_config_path
+    cert, key = await _run_in_executor(
+        google.auth.transport._mtls_helper._get_workload_cert_and_key,
+        certificate_config_path,
     )
+
     if cert and key:
         return True, cert, key, None
 
     return False, None, None, None
 
 
-def get_client_cert_and_key(client_cert_callback=None):
+async def get_client_cert_and_key(client_cert_callback=None):
     """Returns the client side certificate and private key. The function first
     tries to get certificate and key from client_cert_callback; if the callback
     is None or doesn't provide certificate and key, the function tries application
@@ -117,5 +134,7 @@ def get_client_cert_and_key(client_cert_callback=None):
         cert, key = client_cert_callback()
         return True, cert, key
 
-    has_cert, cert, key, _ = get_client_ssl_credentials(generate_encrypted_key=False)
+    has_cert, cert, key, _ = await get_client_ssl_credentials(
+        generate_encrypted_key=False
+    )
     return has_cert, cert, key
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -18,30 +18,12 @@ Helper functions for mTLS in async for discovery of certs.
 
 import asyncio
 import logging
-from os import getenv, path
 
 from google.auth import exceptions
 import google.auth.transport._mtls_helper
+import google.auth.transport.mtls
 
-CERTIFICATE_CONFIGURATION_DEFAULT_PATH = "~/.config/gcloud/certificate_config.json"
 _LOGGER = logging.getLogger(__name__)
-
-
-def _check_config_path(config_path):
-    """Checks for config file path. If it exists, returns the absolute path with user expansion;
-    otherwise returns None.
-
-    Args:
-        config_path (str): The config file path for certificate_config.json for example
-
-    Returns:
-        str: absolute path if exists and None otherwise.
-    """
-    config_path = path.expanduser(config_path)
-    if not path.exists(config_path):
-        _LOGGER.debug("%s is not found.", config_path)
-        return None
-    return config_path
 
 
 async def _run_in_executor(func, *args):
@@ -58,20 +40,6 @@ async def _run_in_executor(func, *args):
         return await loop.run_in_executor(None, func, *args)
 
 
-def has_default_client_cert_source():
-    """Check if default client SSL credentials exists on the device.
-
-    Returns:
-        bool: indicating if the default client cert source exists.
-    """
-    if _check_config_path(CERTIFICATE_CONFIGURATION_DEFAULT_PATH) is not None:
-        return True
-    cert_config_path = getenv("GOOGLE_API_CERTIFICATE_CONFIG")
-    if cert_config_path and _check_config_path(cert_config_path) is not None:
-        return True
-    return False
-
-
 def default_client_cert_source():
     """Get a callback which returns the default client SSL credentials.
 
@@ -83,7 +51,9 @@ def default_client_cert_source():
         google.auth.exceptions.DefaultClientCertSourceError: If the default
             client SSL credentials don't exist or are malformed.
     """
-    if not has_default_client_cert_source():
+    if not google.auth.transport.mtls.has_default_client_cert_source(
+        include_context_aware=False
+    ):
         raise exceptions.MutualTLSChannelError(
             "Default client cert source doesn't exist"
         )
@@ -126,6 +96,7 @@ async def get_client_ssl_credentials(
     cert, key = await _run_in_executor(
         google.auth.transport._mtls_helper._get_workload_cert_and_key,
         certificate_config_path,
+        False,
     )
 
     if cert and key:
@@ -155,13 +126,11 @@ async def get_client_cert_and_key(client_cert_callback=None):
             the cert and key.
     """
     if client_cert_callback:
+        result = client_cert_callback()
         try:
-            # If it's awaitable, this works.
-            cert, key = await client_cert_callback()
+            cert, key = await result
         except TypeError:
-            # If it's not awaitable (e.g., a tuple), result is already the data.
-            cert, key = client_cert_callback()
-
+            cert, key = result
         return True, cert, key
 
     has_cert, cert, key, _ = await get_client_ssl_credentials()
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -13,22 +13,17 @@
 # limitations under the License.
 
 """
-Helper functions for mTLS in asyncio.
+Helper functions for mTLS in async.
 """
 
-import asyncio
-import contextlib
 import logging
-import os
-from os import environ, getenv, path
-import ssl
-import tempfile
-from typing import Optional
+from os import getenv, path
 
-from google.auth import exceptions
+import google.auth.transport._mtls_helper
 
 CERTIFICATE_CONFIGURATION_DEFAULT_PATH = "~/.config/gcloud/certificate_config.json"
 _LOGGER = logging.getLogger(__name__)
+
 
 def _check_config_path(config_path):
     """Checks for config file path. If it exists, returns the absolute path with user expansion;
@@ -53,16 +48,10 @@ def has_default_client_cert_source():
     Returns:
         bool: indicating if the default client cert source exists.
     """
-    if (
-	_check_config_path(CERTIFICATE_CONFIGURATION_DEFAULT_PATH)
-        is not None
-    ):
+    if _check_config_path(CERTIFICATE_CONFIGURATION_DEFAULT_PATH) is not None:
         return True
     cert_config_path = getenv("GOOGLE_API_CERTIFICATE_CONFIG")
-    if (
-        cert_config_path
-        and _check_config_path(cert_config_path) is not None
-    ):
+    if cert_config_path and _check_config_path(cert_config_path) is not None:
         return True
     return False
 
@@ -95,7 +84,9 @@ def get_client_ssl_credentials(
     """
 
     # Attempt to retrieve X.509 Workload cert and key.
-    cert, key =  google.auth.transport._mtls_helper._get_workload_cert_and_key(certificate_config_path)
+    cert, key = google.auth.transport._mtls_helper._get_workload_cert_and_key(
+        certificate_config_path
+    )
     if cert and key:
         return True, cert, key, None
 
@@ -128,4 +119,3 @@ def get_client_cert_and_key(client_cert_callback=None):
 
     has_cert, cert, key, _ = get_client_ssl_credentials(generate_encrypted_key=False)
     return has_cert, cert, key
-
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -17,9 +17,11 @@ Helper functions for mTLS in async for discovery of certs.
 """
 
 import asyncio
+import inspect
 import logging
 from os import getenv, path
 
+from google.auth import exceptions
 import google.auth.transport._mtls_helper
 
 CERTIFICATE_CONFIGURATION_DEFAULT_PATH = "~/.config/gcloud/certificate_config.json"
@@ -71,8 +73,35 @@ def has_default_client_cert_source():
     return False
 
 
+async def default_client_cert_source():
+    """Get a callback which returns the default client SSL credentials.
+
+    Returns:
+        Callable[[], [bytes, bytes]]: A callback which returns the default
+            client certificate bytes and private key bytes, both in PEM format.
+
+    Raises:
+        google.auth.exceptions.DefaultClientCertSourceError: If the default
+            client SSL credentials don't exist or are malformed.
+    """
+    if not has_default_client_cert_source():
+        raise exceptions.MutualTLSChannelError(
+            "Default client cert source doesn't exist"
+        )
+
+    async def callback():
+        try:
+            _, cert_bytes, key_bytes = await get_client_cert_and_key()
+        except (OSError, RuntimeError, ValueError) as caught_exc:
+            new_exc = exceptions.MutualTLSChannelError(caught_exc)
+            raise new_exc from caught_exc
+
+        return cert_bytes, key_bytes
+
+    return callback
+
+
 async def get_client_ssl_credentials(
-    generate_encrypted_key=False,
     certificate_config_path=None,
 ):
     """Returns the client side certificate, private key and passphrase.
@@ -82,10 +111,6 @@ async def get_client_ssl_credentials(
                Currently, only X.509 workload certificates are supported.
 
     Args:
-        generate_encrypted_key (bool): If set to True, encrypted private key
-            and passphrase will be generated; otherwise, unencrypted private key
-            will be generated and passphrase will be None. This option only
-            affects keys obtained via context_aware_metadata.json.
         certificate_config_path (str): The certificate_config.json file path.
 
     Returns:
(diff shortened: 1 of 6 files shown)


google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Returns:
+        bool: indicating if the default client cert source exists.
+    """
+    if _check_config_path(CERTIFICATE_CONFIGURATION_DEFAULT_PATH) is not None:
+        return True
+    cert_config_path = getenv("GOOGLE_API_CERTIFICATE_CONFIG")
+    if cert_config_path and _check_config_path(cert_config_path) is not None:
+        return True
+    return False
```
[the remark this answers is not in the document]
agrawalradhika-cell: context-aware metadata credentials are not in scope of this project currently

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,140 @@
...
+    Raises:
+        google.auth.exceptions.ClientCertError: if problems occurs when getting
+            the cert, key and passphrase.
+    """
+
+    # Attempt to retrieve X.509 Workload cert and key.
+    cert, key = await _run_in_executor(
+        google.auth.transport._mtls_helper._get_workload_cert_and_key,
+        certificate_config_path,
+    )
```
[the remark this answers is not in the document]
agrawalradhika-cell: Have been using the same format as [get_client_ssl_credentials](https://github.com/googleapis/google-auth-library-python/blob/9d5c0d81e8e69719a2b4cd034bf1ed5d128fdf0a/google/auth/transport/_mtls_helper.py#L351) for synchronous flow. Can remove the `generate_encrypted_key` as that is not in use for workload cert discovery.

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,140 @@
...
+            A boolean indicating if cert and key are obtained, the cert bytes
+            and key bytes both in PEM format.
+
+    Raises:
+        google.auth.exceptions.ClientCertError: if problems occurs when getting
+            the cert and key.
+    """
+    if client_cert_callback:
+        cert, key = client_cert_callback()
+        return True, cert, key
```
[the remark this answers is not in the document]
agrawalradhika-cell: can adapt for async callback, not sure if that will be ever the case. still don't mind making the check more robust.

google/auth/aio/transport/mtls.py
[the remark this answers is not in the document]
agrawalradhika-cell: Done

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+
+def has_default_client_cert_source():
+    """Check if default client SSL credentials exists on the device.
+
+    Returns:
+        bool: indicating if the default client cert source exists.
+    """
+    if _check_config_path(CERTIFICATE_CONFIGURATION_DEFAULT_PATH) is not None:
+        return True
+    cert_config_path = getenv("GOOGLE_API_CERTIFICATE_CONFIG")
```
sai-sunder-s: shouldn't we use the value from env var if set and fallback to default?

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
+# Copyright 2024 Google LLC
```
daniel-sanche: 2026 :)

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+    try:
+        # For python versions 3.9 and newer versions
+        return await asyncio.to_thread(func, *args)
+    except AttributeError:
+        # Fallback for older Python versions
+        loop = asyncio.get_running_loop()
+        return await loop.run_in_executor(None, func, *args)
+
+
+def has_default_client_cert_source():
```
sai-sunder-s: can we reuse this method from sync implementation?

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Based on [AIP-4114 ](https://google.aip.dev/auth/4114)- Users should enable Device Certificate Authentication through ADC instead of manual configuration via client options.
Thus, `~/.config/gcloud/certificate_config.json` checks if the certs are present at standard location, and
if the certs are not present at default location, then certs are checked for specific custom config. 

Same logic in synchronous operations - [see url](https://github.com/googleapis/google-auth-library-python/blob/9d5c0d81e8e69719a2b4cd034bf1ed5d128fdf0a/google/auth/transport/mtls.py#L41)

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Actually, `has_default_client_cert_source` [from synchronous flow](https://github.com/googleapis/google-auth-library-python/blob/9d5c0d81e8e69719a2b4cd034bf1ed5d128fdf0a/google/auth/transport/mtls.py#L23) checks for CONTEXT_AWARE_METADATA path too, which is not implemented for async flow yet. Sync method flow would return True if the certificate was present in that location, which would be misleading in a way that Async flow wouldn't implement mTLS configuration in that case

sai-sunder-s: got it. thanks.

sai-sunder-s (approved):

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+
+import asyncio
+import inspect
+import logging
+from os import getenv, path
+
+from google.auth import exceptions
+import google.auth.transport._mtls_helper
+
+CERTIFICATE_CONFIGURATION_DEFAULT_PATH = "~/.config/gcloud/certificate_config.json"
```
daniel-sanche: This seems weird to me... What environment are we expecting to run? Should we be setting this up to work with windows? Or is this for Cloud Shell or some other Google-managed environment?

google/auth/aio/transport/mtls.py
daniel-sanche: Are you sure that's right? That section uses the word *should*, not *must*.

If we put the default check above the env var check, that completely removes the ability to manually override, which doesn't seem like the intended approach to me, even if automatic detection is the best practice

But I just skimmed the doc, so let me know if I'm missing something

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+    if cert_config_path and _check_config_path(cert_config_path) is not None:
+        return True
+    return False
+
+
+async def default_client_cert_source():
+    """Get a callback which returns the default client SSL credentials.
+
+    Returns:
+        Callable[[], [bytes, bytes]]: A callback which returns the default
```
daniel-sanche: This is actually returning a Coroutine, not a Callable. This could be:

`Awaitable[Callable[[], [bytes, bytes]]]`

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    if _check_config_path(CERTIFICATE_CONFIGURATION_DEFAULT_PATH) is not None:
+        return True
+    cert_config_path = getenv("GOOGLE_API_CERTIFICATE_CONFIG")
+    if cert_config_path and _check_config_path(cert_config_path) is not None:
+        return True
+    return False
+
+
+async def default_client_cert_source():
```
daniel-sanche: I don't think you want this async here. There are no awaits inside the function

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+    Raises:
+        google.auth.exceptions.ClientCertError: if problems occurs when getting
+            the cert, key and passphrase.
+    """
+
+    # Attempt to retrieve X.509 Workload cert and key.
+    cert, key = await _run_in_executor(
+        google.auth.transport._mtls_helper._get_workload_cert_and_key,
+        certificate_config_path,
+    )
```
daniel-sanche, line 100: Is this justified? This is a function that runs once per program, reading a couple small files into memory, right?

Introducing async background threads wrapping sync methods can be a minefield, are you sure it's worth it just for this?

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+            A boolean indicating if cert and key are obtained, the cert bytes
+            and key bytes both in PEM format.
+
+    Raises:
+        google.auth.exceptions.ClientCertError: if problems occurs when getting
+            the cert and key.
+    """
+    if client_cert_callback:
+        result = client_cert_callback()
+        if inspect.isawaitable(result):
```
daniel-sanche: nit: I've learned from experience that `inspec.isawaitable` is very slow. It can be better to assume it is, and handle the exception:

```
if client_cert_callback:
    try:
        # If it's awaitable, this works.
        cert, key = await client_cert_callback()
    except TypeError:
        # If it's not awaitable (e.g., a tuple), result is already the data.
        cert, key = client_cert_callback()
 
    return True, cert, key
```

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Helper functions for mTLS in async for discovery of certs.
+"""
```
daniel-sanche, line 17: Can this file be given an underscore name, so the functions aren't used by end users? Right now, these functions would be part of the public API surface, but it seems like they're meant to be internal helpers

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+
+    if cert and key:
+        return True, cert, key, None
+
+    return False, None, None, None
+
+
+async def get_client_cert_and_key(client_cert_callback=None):
+    """Returns the client side certificate and private key. The function first
```
daniel-sanche, line 109: Do you think you can give the structure of this code a bit more thought? It seems like we have 3 methods that do essentially the same thing here:
- `get_client_ssl_credentials`: Returns the client side certificate, private key and passphrase.
- `get_client_cert_and_key`: Returns the client side certificate and private key.
- `default_client_cert_source`: Get a callback which returns the default client SSL credentials. (i.e certificate and key)

I'm not sure if `has_default_client_cert_source` needs to be its own thing either, it seems more like a generalization of the functions above

Can you try to condense these so it's easier to maintain in the future?

daniel-sanche (changes requested):

google/auth/aio/transport/mtls.py
agrawalradhika-cell: No, actually the `has_default_client_source` will be used on the client side to fetch the certs. Thus, they are combination of internal and external helpers. Plus, it will be consistent with the sync flow - see [url](https://github.com/googleapis/google-auth-library-python/blob/9d5c0d81e8e69719a2b4cd034bf1ed5d128fdf0a/google/auth/transport/mtls.py#L23C5-L23C35)

google/auth/aio/transport/mtls.py
agrawalradhika-cell: It is the well-known location, which is for all OS.
Please note there is also `GOOGLE_API_CERTIFICATE_CONFIG` taken into consideration as well to cater to custom cert provided by user.

google/auth/aio/transport/mtls.py
agrawalradhika-cell: The goal of this PR is to provide a fully asynchronous helper library for mTLS certificate discovery. While these files are indeed small, performing any synchronous I/O on the event loop is generally discouraged in asyncio to avoid blocking other tasks. By using _run_in_executor, we ensure that even if the filesystem is slow, the event loop remains responsive.

google/auth/aio/transport/mtls.py
agrawalradhika-cell: The redundancy is a valid point. This structure was designed to maintain parity with the existing synchronous mTLS helpers in google.auth.transport.mtls which many users are already familiar.

has_default_client_cert_source: Provides a check for configuration existence, allowing clients to avoid any overhead if mTLS is not even possible.
default_client_cert_source: This is the primary public entry point that returns the callback object expected by transport sessions.
get_client_ssl_credentials and get_client_cert_and_key are for discovering client cert which can be used in future to broaden to mTLS via ECP and SecureConnect.

google/auth/aio/transport/mtls.py
daniel-sanche: The ~ threw me off, but it looks like it's being passed into [path.expanduser](https://github.com/googleapis/google-auth-library-python/blob/8110a6fa764a5afc99a1e3d4b3fd94d60e08a298/google/auth/aio/transport/mtls.py#L40C19-L40C34). 

Assuming that path is right:

I'd probably suggest using path.expanduser here, so it's already converted (it can always be turned back into a string for _check_config_path). Or at the very least, add a comment like `# default gcloud config path, to be used with path.expanduser`? 

These docs seem to say that windows uses `%APPDATA%\gcloud` though, which would not be found by this. Are you sure this is right?

google/auth/aio/transport/mtls.py
daniel-sanche: Ok, I'm still a bit nervous adding a background thread for such a small, one-time read, but I think this should be ok

google/auth/aio/transport/mtls.py
daniel-sanche: Can we add a flag to the sync helper called `check_context_aware_metadata`, and turn it off when calling from async? It seems like that could cut down on logic duplication

google/auth/aio/transport/mtls.py
daniel-sanche: I realize now that this is coming [from _mtls_helper](https://github.com/googleapis/google-auth-library-python/blob/8110a6fa764a5afc99a1e3d4b3fd94d60e08a298/google/auth/transport/_mtls_helper.py#L28). Maybe windows isn't in scope?'

It would be good if we can import this variable from the helper instead of duplciating it. But even better if we can [share the sync logic](https://github.com/googleapis/google-auth-library-python/pull/1956/changes#r2789698616) so this isn't needed

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,168 @@
...
+            client SSL credentials don't exist or are malformed.
+    """
+    if not has_default_client_cert_source():
+        raise exceptions.MutualTLSChannelError(
+            "Default client cert source doesn't exist"
+        )
+
+    async def callback():
+        try:
+            _, cert_bytes, key_bytes = await get_client_cert_and_key()
```
daniel-sanche, line 63: It seems a little risky to me that it uses `has_default_client_cert_source` to check the location of the keys, but then uses a completely different code path to actually read them. I worry these could get out of sync in the future. Especially since we have logic split between the sync and async implementations

It looks like has_default_client_source reads:
- CERTIFICATE_CONFIGURATION_DEFAULT_PATH
- GOOGLE_API_CERTIFICATE_CONFIG

and get_client_cert_and_key uses [_get_client_cert_config_path(None)](https://github.com/googleapis/google-auth-library-python/blob/8110a6fa764a5afc99a1e3d4b3fd94d60e08a298/google/auth/transport/_mtls_helper.py#L132), which reads
- GOOGLE_API_CERTIFICATE_CONFIG
- CLOUDSDK_CONTEXT_AWARE_CERTIFICATE_CONFIG_FILE_PATH
- CERTIFICATE_CONFIGURATION_DEFAULT_PATH

Didn't you mention that CLOUDSDK_CONTEXT_AWARE_CERTIFICATE_CONFIG_FILE_PATH shouldn't be used by the async code?

daniel-sanche: It would be great if we could refactor the logic that describes where to find the cert into one helper, like this:

```
def _found_cert_path(include_context_aware=True) -> str | None:
  # a single source of truth where we search for the cert locations, and return the place it was found, or None

def has_default_client_cert_source():
  return _found_cert_path() is not None

async def has_default_client_cert_source():
  return _found_cert_path(include_context_aware=False) is not None

def get_client_cert_and_key():
  cert_path = _found_cert_path
  # read certs from path
```

That way, checking for existence vs reading; and async vs sync, it would all flow through the same logic

google/auth/aio/transport/mtls.py
daniel-sanche: And wait, don't [has_default_client_cert_source](https://github.com/googleapis/google-auth-library-python/blob/9d5c0d81e8e69719a2b4cd034bf1ed5d128fdf0a/google/auth/transport/mtls.py#L23) and [_get_cert_config_path](https://github.com/googleapis/google-auth-library-python/blob/8110a6fa764a5afc99a1e3d4b3fd94d60e08a298/google/auth/transport/_mtls_helper.py#L132) also already do essentially the same thing? Can we condense this some more?

I know some of this logic was already in place, but I think we need to be careful about duplication when adding async, otherwise it can get confusing fast

daniel-sanche (commented):
Hopefully my comments aren't too confusing, I was kind of learning the codebase as I went

In general, my feed back is that it would be great if we can re-use logic already in the sync implemtation as much as possible. I see a lot of implementations that are 90% the same as what we implemented elsewhere, with minor critical differences. And that can be very hard to build upon safely in the future

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,168 @@
...
+        google.auth.exceptions.ClientCertError: if problems occurs when getting
+            the cert and key.
+    """
+    if client_cert_callback:
+        try:
+            # If it's awaitable, this works.
+            cert, key = await client_cert_callback()
+        except TypeError:
+            # If it's not awaitable (e.g., a tuple), result is already the data.
+            cert, key = client_cert_callback()
```
daniel-sanche: Sorry, actually looking at this again, I think we do need to use a result variable, like you had before. Otherwise if it's sync, it'll be called twice:

```
    if client_cert_callback:
        result = client_cert_callback()
        try:
            # handle coroutine callback
            cert, key = await result
        except TypeError:
            # unpack non-async callback result
            cert, key = result
```

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Added the comment that you suggested as a disclaimer on the sync side, plus using logic from sync and omitting usage here.

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Added flag to the sync helper and turn off when calling from async. However, condensing has_default_client_cert_source and _get_cert_config_path can be considered out of scope for this project as that would require updating the sync helpers etc. Creating a TODO for tracking this task

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Using the same helper for sync. 

I do acknowledge concerns regarding environment variable precedence, the synchronous implementation intentionally uses a 'defaults-first' approach per AIP-4114 to encourage managed device identities via Application Default Credentials (ADC).
Keeping async logic consistent with the synchronous helper avoids subtle behavioral discrepancies.

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Following the suggestion, I've updated the logic to reuse the sync helpers with a new flag to skip context_aware_metadata checks in the async path. To keep this PR focused on async support, I've added a TODO to consolidate the overlapping discovery methods (has_default_client_cert_source and _found_cert_path) in a subsequent cleanup

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Done

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,137 @@
...
+    Raises:
+        google.auth.exceptions.ClientCertError: if problems occurs when getting
+            the cert, key and passphrase.
+    """
+
+    # Attempt to retrieve X.509 Workload cert and key.
+    cert, key = await _run_in_executor(
+        google.auth.transport._mtls_helper._get_workload_cert_and_key,
+        certificate_config_path,
+        False,
```
daniel-sanche, line 99: nit: this could be cleaner if you use a keyword argument: `include_context_aware=False`

google/auth/aio/transport/mtls.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+async def get_client_ssl_credentials(
+    certificate_config_path=None,
+):
+    """Returns the client side certificate, private key and passphrase.
+
+    We look for certificates and keys with the following order of priority:
+        1. Certificate and key specified by certificate_config.json.
+               Currently, only X.509 workload certificates are supported.
```
daniel-sanche, line 80: should there be more in this list?

google/auth/aio/transport/mtls.py
daniel-sanche: Thanks, the recent change looks great!

Is the subsequent clean up something you have planned? Or is it just a TODO for sometime in the future?

daniel-sanche (approved):
Change LGTM, but I left a few small comments

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Will track the TODO on buganizer task as a followup

google/auth/aio/transport/mtls.py
agrawalradhika-cell: No, this is as is. 

Because, we are passing False to `include_context_aware` flag. Thus, only workload certs are looked

PR #1959, opened 2026-02-17
feat: mTLS configuration via x.509 for asynchronous session in google-auth

agrawalradhika-cell:
This pull request introduces support for Mutual TLS (mTLS) in the asynchronous transport layer of the google-auth library. It enables AsyncAuthorizedSession to automatically discover and utilize client certificates for secure communication with Google Cloud APIs. See [go/caa:x509-async-support](http://goto.google.com/caa:x509-async-support) for details.

Please note: Only x.509 creds are in scope of this project currently. Context aware or ECP credentials are not in scope of this project currently.

This PR is second part of https://github.com/googleapis/google-auth-library-python/pull/1956

### DIFF
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -70,20 +70,19 @@ def make_client_cert_ssl_context(
     Raises:
         google.auth.exceptions.TransportError: If there is an error loading the certificate.
     """
-    try:
-        context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
-
-        # Write cert and key to temp files because ssl.load_cert_chain requires paths
-        with _create_temp_file(cert_bytes) as cert_path:
-            with _create_temp_file(key_bytes) as key_path:
-                context.load_cert_chain(
-                    certfile=cert_path, keyfile=key_path, password=passphrase
-                )
-        return context
-    except (ssl.SSLError, OSError) as exc:
-        raise exceptions.TransportError(
-            "Failed to load client certificate and key for mTLS."
-        ) from exc
+    with _create_temp_file(cert_bytes) as cert_path, _create_temp_file(
+        key_bytes
+    ) as key_path:
+        try:
+            context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+            context.load_cert_chain(
+                certfile=cert_path, keyfile=key_path, password=passphrase
+            )
+            return context
+        except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError) as exc:
+            raise exceptions.TransportError(
+                "Failed to load client certificate and key for mTLS."
+            ) from exc
 
 
 async def _run_in_executor(func, *args):
@@ -104,7 +103,7 @@ def default_client_cert_source():
     """Get a callback which returns the default client SSL credentials.
 
     Returns:
-        Awaitable[Callable[[], [bytes, bytes]]]: A callback which returns the default
+        Awaitable[Callable[[], Tuple[bytes, bytes]]]: A callback which returns the default
             client certificate bytes and private key bytes, both in PEM format.
 
     Raises:
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -20,7 +20,6 @@ import asyncio
 import contextlib
 import logging
 import os
-from os import getenv, path
 import ssl
 import tempfile
 from typing import Optional
@@ -29,6 +28,7 @@ from google.auth import exceptions
 import google.auth.transport._mtls_helper
 
 _LOGGER = logging.getLogger(__name__)
+
 
 @contextlib.contextmanager
 def _create_temp_file(content: bytes):
@@ -83,6 +83,7 @@ def make_client_cert_ssl_context(
         raise exceptions.TransportError(
             "Failed to load client certificate and key for mTLS."
         ) from exc
+
 
 async def _run_in_executor(func, *args):
     """Run a blocking function in an executor to avoid blocking the event loop.
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 """
-Helper functions for mTLS in async.
+Helper functions for mTLS in async for discovery of certs.
 """
 
+import asyncio
 import logging
 from os import getenv, path
 
@@ -42,6 +43,20 @@ def _check_config_path(config_path):
     return config_path
 
 
+async def _run_in_executor(func, *args):
+    """Run a blocking function in an executor to avoid blocking the event loop.
+
+    This implements the non-blocking execution strategy for disk I/O operations.
+    """
+    try:
+        # For python versions 3.9 and newer versions
+        return await asyncio.to_thread(func, *args)
+    except AttributeError:
+        # Fallback for older Python versions
+        loop = asyncio.get_running_loop()
+        return await loop.run_in_executor(None, func, *args)
+
+
 def has_default_client_cert_source():
     """Check if default client SSL credentials exists on the device.
 
@@ -56,7 +71,7 @@ def has_default_client_cert_source():
     return False
 
 
-def get_client_ssl_credentials(
+async def get_client_ssl_credentials(
     generate_encrypted_key=False,
     certificate_config_path=None,
 ):
@@ -84,16 +99,18 @@ def get_client_ssl_credentials(
     """
 
     # Attempt to retrieve X.509 Workload cert and key.
-    cert, key = google.auth.transport._mtls_helper._get_workload_cert_and_key(
-        certificate_config_path
+    cert, key = await _run_in_executor(
+        google.auth.transport._mtls_helper._get_workload_cert_and_key,
+        certificate_config_path,
     )
+
     if cert and key:
         return True, cert, key, None
 
     return False, None, None, None
 
 
-def get_client_cert_and_key(client_cert_callback=None):
+async def get_client_cert_and_key(client_cert_callback=None):
     """Returns the client side certificate and private key. The function first
     tries to get certificate and key from client_cert_callback; if the callback
     is None or doesn't provide certificate and key, the function tries application
@@ -117,5 +134,7 @@ def get_client_cert_and_key(client_cert_callback=None):
         cert, key = client_cert_callback()
         return True, cert, key
 
-    has_cert, cert, key, _ = get_client_ssl_credentials(generate_encrypted_key=False)
+    has_cert, cert, key, _ = await get_client_ssl_credentials(
+        generate_encrypted_key=False
+    )
     return has_cert, cert, key
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -17,14 +17,72 @@ Helper functions for mTLS in async for discovery of certs.
 """
 
 import asyncio
+import contextlib
 import logging
+import os
+from os import getenv, path
+import ssl
+import tempfile
+from typing import Optional
 
 from google.auth import exceptions
 import google.auth.transport._mtls_helper
-import google.auth.transport.mtls
 
 _LOGGER = logging.getLogger(__name__)
 
+@contextlib.contextmanager
+def _create_temp_file(content: bytes):
+    """Creates a temporary file with the given content.
+
+    Args:
+        content (bytes): The content to write to the file.
+
+    Yields:
+        str: The path to the temporary file.
+    """
+    # Create a temporary file that is readable only by the owner.
+    fd, file_path = tempfile.mkstemp()
+    try:
+        with os.fdopen(fd, "wb") as f:
+            f.write(content)
+        yield file_path
+    finally:
+        # Securely delete the file after use.
+        if os.path.exists(file_path):
+            os.remove(file_path)
+
+
+def make_client_cert_ssl_context(
+    cert_bytes: bytes, key_bytes: bytes, passphrase: Optional[bytes] = None
+) -> ssl.SSLContext:
+    """Creates an SSLContext with the given client certificate and key.
+    This function writes the certificate and key to temporary files so that
+    ssl.create_default_context can load them, as the ssl module requires
+    file paths for client certificates.
+    Args:
+        cert_bytes (bytes): The client certificate content in PEM format.
+        key_bytes (bytes): The client private key content in PEM format.
+        passphrase (Optional[bytes]): The passphrase for the private key, if any.
+    Returns:
+        ssl.SSLContext: The configured SSL context with client certificate.
+
+    Raises:
+        google.auth.exceptions.TransportError: If there is an error loading the certificate.
+    """
+    try:
+        context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+
+        # Write cert and key to temp files because ssl.load_cert_chain requires paths
+        with _create_temp_file(cert_bytes) as cert_path:
+            with _create_temp_file(key_bytes) as key_path:
+                context.load_cert_chain(
+                    certfile=cert_path, keyfile=key_path, password=passphrase
+                )
+        return context
+    except (ssl.SSLError, OSError) as exc:
+        raise exceptions.TransportError(
+            "Failed to load client certificate and key for mTLS."
+        ) from exc
 
 async def _run_in_executor(func, *args):
     """Run a blocking function in an executor to avoid blocking the event loop.
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -26,6 +26,7 @@ from typing import Optional
 
 from google.auth import exceptions
 import google.auth.transport._mtls_helper
+import google.auth.transport.mtls
 
 _LOGGER = logging.getLogger(__name__)
 
--- a/google/auth/aio/transport/mtls.py
+++ b/google/auth/aio/transport/mtls.py
@@ -18,30 +18,12 @@ Helper functions for mTLS in async for discovery of certs.
 
 import asyncio
 import logging
-from os import getenv, path
 
 from google.auth import exceptions
 import google.auth.transport._mtls_helper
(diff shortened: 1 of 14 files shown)


google/auth/aio/transport/mtls.py
```diff
@@ -26,6 +31,61 @@
...
+
+    Raises:
+        google.auth.exceptions.TransportError: If there is an error loading the certificate.
+    """
+    try:
+        context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+
+        # Write cert and key to temp files because ssl.load_cert_chain requires paths
+        with _create_temp_file(cert_bytes) as cert_path:
+            with _create_temp_file(key_bytes) as key_path:
```
daniel-sanche: nit: I think you can double up the with statement, to reduce the nesting

```
with _create_temp_file(cert_bytes) as cert_path, _create_temp_file(key_bytes) as key_path:
```

google/auth/aio/transport/mtls.py
```diff
@@ -26,6 +31,61 @@
...
+                context.load_cert_chain(
+                    certfile=cert_path, keyfile=key_path, password=passphrase
+                )
+        return context
+    except (ssl.SSLError, OSError) as exc:
```
daniel-sanche: I think IOError could also be raised. And maybe others

google/auth/aio/transport/sessions.py
```diff
@@ -124,12 +127,70 @@ def __init__(
...
         self._auth_request = _auth_request
 
+    async def configure_mtls_channel(self, client_cert_callback=None):
+        """Configure the client certificate and key for SSL connection.
+
+        The function does nothing unless `GOOGLE_API_USE_CLIENT_CERTIFICATE` is
+        explicitly set to `true`. In this case if client certificate and key are
+        successfully obtained (from the given client_cert_callback or from application
+        default SSL credentials), the underlying transport will be reconfigured
+        to use mTLS.
```
daniel-sanche, line 157: The docstring should also mention that this function does nothing if aiohttp isn't installed (should it print a log or something?)

google/auth/aio/transport/sessions.py
```diff
@@ -124,12 +127,70 @@ def __init__(
...
+                ssl_context = await mtls._run_in_executor(
+                    mtls.make_client_cert_ssl_context, cert, key
+                )
+
+                # Re-create the auth request with the new SSL context
+                if AIOHTTP_INSTALLED and isinstance(self._auth_request, AiohttpRequest):
+                    connector = aiohttp.TCPConnector(ssl=ssl_context)
+                    new_session = aiohttp.ClientSession(connector=connector)
+                    await self._auth_request.close()
+                    self._auth_request = AiohttpRequest(session=new_session)
```
daniel-sanche: Wouldn't there be concurrency issues to worry about here? It seems like there is a window where it yields to the event loop with `self._auth_request` in a closed state

I'm not familiar with the session class, but wouldn't that result in failed requests?

daniel-sanche: Also, does calling close() immediately cancel any ongoing requests? Or does it let ongoing ones finish first?

daniel-sanche (changes requested):

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Added

google/auth/aio/transport/mtls.py
agrawalradhika-cell: done

google/auth/aio/transport/sessions.py
agrawalradhika-cell: Added comment on docstring

google/auth/aio/transport/sessions.py
agrawalradhika-cell: you are right, fixed by atomic swap

yaoliugg (approved):

google/auth/aio/transport/sessions.py
```diff
@@ -133,12 +137,73 @@ def __init__(
...
+                    mtls.make_client_cert_ssl_context, cert, key
+                )
+
+                # Re-create the auth request with the new SSL context
+                if AIOHTTP_INSTALLED and isinstance(self._auth_request, AiohttpRequest):
+                    connector = aiohttp.TCPConnector(ssl=ssl_context)
+                    new_session = aiohttp.ClientSession(connector=connector)
+                    old_auth_request = self._auth_request
+                    self._auth_request = AiohttpRequest(session=new_session)
+                    await old_auth_request.close()
```
daniel-sanche: This question still stands: `does calling close() immediately cancel any ongoing requests? Or does it let ongoing ones finish first?`

Do you know if calling `await old_auth_request.close()` would instantly kill existing requests? Would that be a problem?

google/auth/aio/transport/sessions.py
```diff
@@ -33,6 +36,7 @@
         ClientTimeout = None
 
 try:
+    import aiohttp
```
daniel-sanche: it seems a bit weird to me that ClientTimeout is imported in a separate block, so we could get into a situation where ClientTimeout is present, but AIOHTTP_INSTALLED is False

google/auth/aio/transport/sessions.py
```diff
@@ -186,18 +251,29 @@ async def request(
...
             await with_timeout(
                 # Note: before_request will attempt to refresh credentials if expired.
                 self._credentials.before_request(
                     self._auth_request, method, url, headers
                 )
             )
+            if AIOHTTP_INSTALLED and hasattr(timeout, "total"):
+                # If it's a ClientTimeout, extract the total float
+                actual_timeout = typing.cast(float, timeout.total)
+            elif timeout is None:
```
daniel-sanche: Is None an option? Type types say it shouldn't be

google/auth/aio/transport/sessions.py
```diff
@@ -186,18 +251,29 @@ async def request(
...
+                actual_timeout = 0.0
+            else:
+                actual_timeout = typing.cast(float, timeout)
```
daniel-sanche: This logic seems off to me. If AIOHTTP_INSTALLED is False, but timeout is a ClientTimeout (which seems like [it should be possible](https://github.com/googleapis/google-auth-library-python/pull/1959/changes#r2843330512)), we would fall through the else clause and cast the ClientTimeout into a float. 

In general, typing.cast should be avoided as much as possible, since it can hide real typing issues. Would something like this work?

```
actual_timeout: float = 0
if ClientTimeout is not None and isinstance(timeout, ClientTimeout):
  actual_timeout = timeout.total
elif isinstance(timeout, (int, float)):
  actual_timeout = timeout
```

google/auth/aio/transport/mtls.py
```diff
@@ -26,6 +31,60 @@
...
+            os.remove(file_path)
+
+
+def make_client_cert_ssl_context(
+    cert_bytes: bytes, key_bytes: bytes, passphrase: Optional[bytes] = None
+) -> ssl.SSLContext:
+    """Creates an SSLContext with the given client certificate and key.
+    This function writes the certificate and key to temporary files so that
+    ssl.create_default_context can load them, as the ssl module requires
+    file paths for client certificates.
```
daniel-sanche: To be clear, the temporary files are created and destroyed within the scope of this function, right? Maybe that should be mentioned in the docstring?

google/auth/aio/transport/sessions.py
daniel-sanche: It looks like ClientTimeout.total can be None too. How do we want to handle that case? `actual_timeout = 0`?

google/auth/aio/transport/mtls.py
agrawalradhika-cell: Added in dosctring

google/auth/aio/transport/sessions.py
agrawalradhika-cell: Fixed

google/auth/aio/transport/sessions.py
agrawalradhika-cell: Yes, calling close() does not wait for ongoing requests to finish, and any in-flight requests would be abruptly terminated with a connection error.
As for whether this is a problem: it shouldn't be, provided that configure_mtls_channel() is only called during the initial phase of the session, before any actual API requests are made. Assuming this is strictly an initialization step, it should be safe, Added a note in the docstring mentioning it.

google/auth/aio/transport/sessions.py
agrawalradhika-cell: You are right, Updated

google/auth/aio/transport/sessions.py
agrawalradhika-cell: Good catch. ClientTimeout.total can indeed be None. I've updated the logic to handle that case by defaulting to 0.0, which maintains the 'no timeout' behavior while satisfying the float requirement of the transport layer.

I've also switched to using isinstance checks for int and float to narrow the type more effectively for mypy.

google/auth/aio/transport/sessions.py
```diff
@@ -154,7 +154,7 @@ async def configure_mtls_channel(self, client_cert_callback=None):
         default SSL credentials), the underlying transport will be reconfigured
         to use mTLS.
         Note: This function does nothing if the `aiohttp` library is not
-        installed.
+        installed. Plus, will close any ongoing API requests.
```
daniel-sanche: Putting this at the end of this note feels misleading. This isn't related to aiohttp, right? I'd suggest breaking this out into its own section

google/auth/aio/transport/sessions.py
daniel-sanche, line 45: can `from google.auth.aio.transport.aiohttp import Request as AiohttpRequest` be grouped in with the other  try block? Or at least move AIOHTTP_INSTALLED to follow the other aiohttp imports?

This structure still seems like it could lead to confusing bugs

google/auth/aio/transport/sessions.py
daniel-sanche: But this is implemented as a coroutine, so we don't have the same sequencing guarantees. Doesn't that create an issue? When it calls await, it defers to the event loop, letting other tasks run. One of those tasks may start an API request, that would then be cancelled when this task resumes

IIUC, if you want to make sure this is run before any API requests are made, it should be a blocking synchronous call (and ideally part of \_\_init__)

google/auth/aio/transport/sessions.py
```diff
@@ -260,13 +260,13 @@ async def request(
...
-            elif timeout is None:
-                actual_timeout = 0.0
+            actual_timeout: float = 0.0
+            if AIOHTTP_INSTALLED and isinstance(timeout, aiohttp.ClientTimeout):
+                actual_timeout = timeout.total if timeout.total is not None else 0.0
+            elif isinstance(timeout, (int, float)):
+                actual_timeout = float(timeout)
             else:
-                actual_timeout = typing.cast(float, timeout)
+                actual_timeout = 0.0
```
daniel-sanche: nit: the else block is redundant since actual_timeout was initialized to 0

daniel-sanche (changes requested):

google/auth/aio/transport/sessions.py
agrawalradhika-cell: Done

google/auth/aio/transport/sessions.py
agrawalradhika-cell: I appreciate the suggestion for consolidation, but I’d prefer to keep these blocks separate as it follows existing logic in the codebase where we track these dependencies independently.

The first block specifically manages the external aiohttp package and types like ClientTimeout, while the second block tracks our internal google.auth.aio.transport.aiohttp. The AIOHTTP_INSTALLED flag is intentionally scoped to the second block because its specific role is to signal that the AiohttpRequest class is available to be used as a default in the constructor.

Keeping them decoupled prevents a situation where a failure in importing our internal adapter unnecessarily impacts the presence of the external types used for type narrowing elsewhere. For these reasons, I suggest we keep the current structure.

Added comment mentioning this to help avoid confusion

google/auth/aio/transport/sessions.py
agrawalradhika-cell: I see the issue—the coroutine yield allows 'interrupter' tasks to start requests that then get killed by the close() call when the setup resumes.

To solve this while keeping the configuration async, I’ve implemented a gatekeeper in request(). Now, any call to request() will automatically await the completion of the mTLS configuration task if one exists. This guarantees that the mTLS setup always completes its 'close and swap' logic before any API requests are allowed to leave the client, effectively restoring the sequencing guarantee.

google/auth/aio/transport/sessions.py
agrawalradhika-cell: fixed

google/auth/aio/transport/sessions.py
```diff
@@ -247,7 +263,8 @@ async def request(
                 the configured `max_allowed_time` or the request exceeds the configured
                 `timeout`.
         """
-
+        if self._mtls_init_task:
+            await self._mtls_init_task
```
daniel-sanche: if _mtls_init_task raises any exceptions, they would be raised here too. Maybe you should put this in a try/catch?

google/auth/aio/transport/sessions.py
daniel-sanche, line 150: I think this new structure should work..l honestly, I think this should just be a blocking call. Is the complexity of managing a background task really worth the benefit of readying the keys from disk asynchronously? It seems like everything could be much cleaner if it was done in synchronously in __init__

google/auth/aio/transport/sessions.py
```diff
@@ -167,42 +172,53 @@ async def configure_mtls_channel(self, client_cert_callback=None):
...
+
+                except (
+                    exceptions.ClientCertError,
+                    ImportError,
+                    OSError,
+                ) as caught_exc:
+                    new_exc = exceptions.MutualTLSChannelError(caught_exc)
+                    raise new_exc from caught_exc
+
+            self._mtls_init_task = asyncio.create_task(_do_configure())
```
daniel-sanche, line 219: Something to consider: `self._mtls_init_task = asyncio.create_task(self.configure_mtls_channel())` could be put in \_\_init__ if you wanted the background task to start automatically on creation. We did [something similar in buigtable](https://github.com/googleapis/python-bigtable/blob/9f7ec2e5f651715407cda5cb4c46a4f5ca17f507/google/cloud/bigtable/data/_async/client.py#L1046). You'd just need to check for RuntimeErrors if the class is created outside an event loop context

google/auth/aio/transport/sessions.py
```diff
@@ -261,12 +278,10 @@ async def request(
                 )
             )
             actual_timeout: float = 0.0
-            if AIOHTTP_INSTALLED and isinstance(timeout, aiohttp.ClientTimeout):
+            if isinstance(timeout, aiohttp.ClientTimeout):
```
daniel-sanche: wouldn't this cause an error if aiohttp was missing? Or is that ok? I'm a bit confused by the whole AIOHTTP_INSTALLED situation

agrawalradhika-cell: If you see I have mentioned in prior [comment](https://github.com/googleapis/google-auth-library-python/pull/1959/changes#r2845210828) that  AIOHTTP_INSTALLED checks the internal aiohttp import and ClientTimeout uses the external aiohttp. I'll add a check `ClientTimeout is not None` to help this won't cause in any error

google/auth/aio/transport/sessions.py
agrawalradhika-cell: We don't want the background task to start start automatically on creation, `configure_mtls_channel` is called by client specifically.

google/auth/aio/transport/sessions.py
agrawalradhika-cell: appreciate the suggestion to simplify the flow. However, the GenAI client specifically asked for this to be asynchronous. To ensure we meet their immediate needs, I'd like to proceed with the current design and treat the transition to a synchronous call as a separate item to discuss with them later.

discussed offline with Daniel on this

google/auth/aio/transport/sessions.py
agrawalradhika-cell: added try catch block

google/auth/aio/transport/sessions.py
```diff
@@ -182,22 +263,39 @@ async def request(
...
-
+        if self._mtls_init_task:
+            try:
+                await self._mtls_init_task
+            except exceptions.MutualTLSChannelError:
+                # Re-raise the already-wrapped mTLS error
+                raise
+            except Exception as caught_exc:
+                # Wrap any other unexpected exceptions from the task
+                raise exceptions.MutualTLSChannelError(caught_exc) from caught_exc
```
daniel-sanche: IIUC with this code, if _mtls_init_task fails, every subsequent request will raise an exception. That doesn't seem like the behavior you'd want

My suggestion was actually to just supress all exceptions from `self._mtls_init_task` here. The only reason we're awaiting it here is to make sure it completes before we start new requests, but I don't think the caller of this function cares too much about the status of that task, do they?

agrawalradhika-cell: supressed all exceptions from self._mtls_init_task here.

google/auth/aio/transport/sessions.py
```diff
@@ -182,22 +263,37 @@ async def request(
...
+            headers = {}
         async with timeout_guard(max_allowed_time) as with_timeout:
             await with_timeout(
                 # Note: before_request will attempt to refresh credentials if expired.
                 self._credentials.before_request(
                     self._auth_request, method, url, headers
                 )
             )
+            actual_timeout: float = 0.0
+            if ClientTimeout is not None and isinstance(timeout, aiohttp.ClientTimeout):
```
daniel-sanche: Can you replace `aiohttp.ClientTimeout` with `ClientTimeout` here? It seems weird to reference it in two different ways in the same line

agrawalradhika-cell: Done

daniel-sanche (approved):
LGTM!

